PR #63, opened 2021-02-08
fix: sar LSTM decoder symbol error

charlesmindee:
This is a fix for the SAR decoder module, which was missing an injection at timestep t of the t-1 decoded symbol:

"From step2, the output of the previous step is fed into LSTM until the “END” token is received. All the LSTM inputs are represented by one-hot vectors, followed by a linear transformation Ψ()." [https://arxiv.org/pdf/1811.00751.pdf](url)

logits at time t (outputs of lstm cell) are decoded (argmax) to find the corresponding symbol, which is then one-hot encoded and embedded (linear transformation) to be fed at time t+1


doctr/models/recognition/sar.py
```diff
@@ -134,7 +137,10 @@ def call(
             logits = tf.concat([logits, glimpse], axis=-1)
             # shape (N, rnn_units + 1) -> (N, num_classes + 1)
             logits = self.output_dense(logits, **kwargs)
+            # update symbol with predicted logits for t+1 step
+            symbol = tf.argmax(logits, axis=-1)
```
fg-mindee, line 141: Could you elaborate a bit more in the PR description what was not done correctly and how this fixes it please? If there is some TF documentation reference, please add it as well :pray:

fg-mindee (commented):
Thanks for the PR! Just some clarification needed but LGTM

fg-mindee (dismissed):
Looks good to me, thanks!

doctr/models/recognition/sar.py
```diff
@@ -127,6 +127,7 @@ def call(
         _, states = self.lstm_decoder(holistic, states, **kwargs)
         sos_symbol = self.vocab_size + 1
         symbol = sos_symbol * tf.ones(shape=(batch_size,), dtype=tf.int32)
```
fg-mindee, line 129: Sorry, last request: could you change this to:
```
symbol = tf.fill(batch_size, sos_symbol)
```
to avoid the unnecessary multiplication

fg-mindee (approved):
Thanks again!

PR #74, opened 2021-02-09
Fix: <eos> symbol removed from vocab str

charlesmindee:
<eos> symbol has been removed from vocab (but kept in _embedded), len(vocab) = size of the alphabet

### DIFF
--- a/doctr/models/recognition/core.py
+++ b/doctr/models/recognition/core.py
@@ -103,8 +103,8 @@ class RecognitionPostProcessor:
         ignore_accents: bool = False
     ) -> None:
 
-        self.vocab = vocab + '<eos>'
-        self._embedding = tf.constant([char for char in self.vocab], dtype=tf.string, shape=[len(self.vocab)])
+        self.vocab = vocab
+        self._embedding = tf.constant([char for char in self.vocab] + ['<eos>'], dtype=tf.string)
         self.ignore_case = ignore_case
         self.ignore_accents = ignore_accents
 
--- a/doctr/models/recognition/postprocessor.py
+++ b/doctr/models/recognition/postprocessor.py
@@ -47,7 +47,7 @@ class CTCPostProcessor(RecognitionPostProcessor):
 
         # masking -1 of CTC with num_classes (embedding <eos>)
         pred_shape = tf.shape(_predictions)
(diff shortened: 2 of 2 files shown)


doctr/models/recognition/sar.py
```diff
@@ -125,13 +125,13 @@ def call(
         # run first step of lstm
         # holistic: shape (N, rnn_units)
         _, states = self.lstm_decoder(holistic, states, **kwargs)
-        sos_symbol = self.vocab_size + 1
+        sos_symbol = self.vocab_size
```
fg-mindee, line 128: The vocab_size here is unchanged compared to earlier. If I understand well, this points to the index of the last symbol ("<eos>") of the vocab. If that is so, I believe there is a typo "sos_symbol"
Additionally, for clarity, this is an index, so I suggest naming it `eos_idx`

doctr/models/recognition/sar.py
```diff
@@ -125,13 +125,13 @@ def call(
...
         symbol = sos_symbol * tf.ones(shape=(batch_size,), dtype=tf.int32)
```
fg-mindee, line 129: Let's avoid the multiplication here:
```
symbol = tf.fill(batch_size, eos_idx)
```

fg-mindee (changes requested):
I'm still confused about why we don't have to change the unittest :man_shrugging: I added extra suggestions to improve the PR

doctr/models/recognition/sar.py
```diff
@@ -125,8 +125,8 @@ def call(
         # run first step of lstm
         # holistic: shape (N, rnn_units)
         _, states = self.lstm_decoder(holistic, states, **kwargs)
-        sos_symbol = self.vocab_size
-        symbol = sos_symbol * tf.ones(shape=(batch_size,), dtype=tf.int32)
+        # fill with START symbol (index = vocab_size + 1)
```
fg-mindee, line 128: Perhaps something more explicit:
```
# Initialize with the index of virtual START symbol (placed after <eos>)
```

fg-mindee (commented):
Only a comment suggestion but the rest looks fine to me!

fg-mindee (dismissed):
Thanks for this critical fix!

PR #82, opened 2021-02-11
refactor: Refactored visualization

### LINKED ISSUE #57: [utils] Move doctr dependences out of doctr.utils
The `doctr.utils.visualization` introduces a dependency of other modules, which might be troublesome later. Several options are at our disposal:
- sterilize these imports, and implement the specific version in the modules of the former dependency
- use exported dictionary versions of element to plot (typing would then not require the imports of doctr.documents)

Any other suggestion is welcome!

fg-mindee:
- removes the dependency of doctr.documents in doctr.utils by using the export dict of page to plot them
- updates unittest

Successfully merging this PR will close #57 

Any feedback is welcome!

### DIFF
--- a/doctr/utils/visualization.py
+++ b/doctr/utils/visualization.py
@@ -7,78 +7,36 @@ import matplotlib.pyplot as plt
 import matplotlib.patches as patches
 import mplcursors
 import numpy as np
-from typing import Union, Tuple, List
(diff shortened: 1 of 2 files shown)


charlesmindee (dismissed):
Nice refacto, maybe we should add a step for drawing artefacts too ?

charlesmindee (approved):

PR #84, opened 2021-02-12
Refactor: changed default cfg for both DBnet & CRNN

charlesmindee:
Changed default cfg for both dbnet and crnn according to weights uploaded on the v0.1 tag.


fg-mindee (approved):
Thanks for the PR!

PR #106, opened 2021-03-01
docs: Added model zoo table in documentation

### LINKED ISSUE #94: [models] Add a descriptive table for pretrained models
The documentation does provide the list of implementation architectures, but some information is lacking without reading the code:
- input image size
- size of the model (# params)
- performance

fg-mindee:
- adds a table for text detection models in the documentation (archi, input shape, num of params, recall, precision, fps)
- adds a table for text recognition models (archi, input shape, num of params, accuracy, fps)
- adds a reference to those tables in the README

Successfully merging this PR will close #94

### DIFF
--- a/docs/source/models.rst
+++ b/docs/source/models.rst
@@ -17,6 +17,23 @@ Text Detection
 --------------
 Localizing text elements in images
 
+.. list-table:: Text detection model zoo
+   :widths: 20 20 15 10 10 10
+   :header-rows: 1
+
+   * - Architecture
+     - Input shape
+     - # params
+     - Recall
+     - Precision
+     - FPS
+   * - db_resnet50
+     - (1024, 1024, 3)
+     -
+     -
+     -
+     -
+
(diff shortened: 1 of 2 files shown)


docs/source/models.rst
```diff
@@ -17,6 +17,23 @@ Text Detection
...
 Localizing text elements in images
 
+.. list-table:: Text detection model zoo
+   :widths: 20 20 15 10 10 10
+   :header-rows: 1
+
+   * - Architecture
+     - Input shape
+     - # params
+     - Recall
```
charlesmindee, line 27: I think we also need to reference the test dataset on which we evaluate the models

charlesmindee (approved):

docs/source/models.rst
fg-mindee: Indeed, I left the value blank but I assumed we would mention the test set used above the table since it will have to apply to all entries for easier comparison :+1:

PR #107, opened 2021-03-02
feat: add sar teacher forcing

charlesmindee:
Successfully merging this PR closes #71 , any feedback is welcome!

PR #108, opened 2021-03-02
fix: pad teacher forcing

charlesmindee:
This is a fix for the PR #107: dense labels for teacher forcing were not padded to the length (max_lentgh + 1), and it was crashing the model. SAR is now operational for training. Any feedback is welcome !


fg-mindee (approved):
Thanks for the PR!

PR #110, opened 2021-03-04
feat: Add recognition metric (exact match)

charlesmindee:
This PR implements ExactMatch metric (word-level accuracy) for recognition task
Any feedback is welcome !


doctr/utils/metrics.py
```diff
@@ -0,0 +1,75 @@
...
+                pred_word = remove_accent(pred_word)
+
+            if self.ignore_case:
+                gt_word = gt_word.lower()
+                pred_word = pred_word.lower()
+
+            if pred_word == gt_word:
+                self.matches += 1
+
+            self.total += 1
```
fg-mindee, line 67: You can indent this out of the loop and simply:
```
self.total += len(gt)
```

doctr/utils/metrics.py
```diff
@@ -0,0 +1,75 @@
...
+        pred: List[str],
+    ) -> None:
+        """Update the state of the metric with new predictions
+
+        Args:
+            gt: list of groung-truth character sequences
+            pred: list of predicted character sequences"""
+
+        if len(gt) != len(pred):
+            raise AttributeError("prediction size does not match with ground-truth labels size")
```
fg-mindee, line 53: I think it might be more appropriate to raise an `AssertionError`

doctr/utils/metrics.py
```diff
@@ -0,0 +1,75 @@
...
+    """Implements exact match metric (word-level accuracy) for recognition task
+
+    Args:
+        ignore_case: if true, ignore letter case when computing metric
+        ignore_accents: if true, ignore accents errors when computing metrics"""
+
+    def __init__(
+        self,
+        ignore_case,
+        ignore_accents
```
fg-mindee, line 21: Let's set default values (`False`) for those arguments

test/test_utils_metrics.py
```diff
@@ -0,0 +1,12 @@
+from doctr.utils import metrics
```
fg-mindee, line 1: Add a `import pytest` before this if you want to use the refactor I suggested below :ok_hand:

test/test_utils_metrics.py
```diff
@@ -0,0 +1,12 @@
...
+
+def test_exact_match():
+    mock_gt = ['grass', '56', 'True', 'STOP']
+    mock_pred = ['grass', '56', 'true', 'stop']
+    metric_a = metrics.ExactMatch(ignore_case=True, ignore_accents=False)
+    metric_b = metrics.ExactMatch(ignore_case=False, ignore_accents=False)
+    metric_a.update_state(mock_gt, mock_pred)
+    metric_b.update_state(mock_gt, mock_pred)
+    assert metric_a.result() == 1.0
+    assert metric_b.result() == 0.5
```
fg-mindee, line 12: Thanks to pytest, we could refactor the test to check many cases like so:
```
@pytest.mark.parametrize(
    "gt, pred, ignore_case, ignore_accents, result",
    [
        [['grass', '56', 'True', 'STOP'], ['grass', '56', 'true', 'stop'], True, False, 1.0],
        [['grass', '56', 'True', 'STOP'], ['grass', '56', 'true', 'stop'], False, False, .5],
    ],
)
def test_exact_match(gt, pred, ignore_case, ignore_accents, result):
    metric = metrics.ExactMatch(ignore_case, ignore_accents)
    metric.update_state(gt, pred)
    assert metric.result() == result
```

fg-mindee (commented):
Thanks for the PR! I added a few comments. Besides, especially if we don't ignore accents or case, we could actually do this with TF and compare the output before decoding?

charlesmindee:
> Thanks for the PR! I added a few comments. Besides, especially if we don't ignore accents or case, we could actually do this with TF and compare the output before decoding?

I think so, but I would require tensor manipulations to compare row-wise both tensors. What I was doing previously was translating with the dictionary to a string tensor and I concatenated the columns to have a vector of words, but I think we can compare the number sequences at a row-level without translating the outputs to save time.

fg-mindee (approved):
We might want later to convert this to a TF-only computation, but it will do for now! Thanks for the edits :pray:

PR #111, opened 2021-03-04
feat: Added pretrained weights for SAR

charlesmindee:
This PR repaces the url of the SAR pretrained model in the default cfgs


fg-mindee (approved):
Thanks for the training :pray:

PR #112, opened 2021-03-05
add: default url crnn

charlesmindee:
This PR adds new default url in crnn cfgs


fg-mindee (approved):
Thanks for training this one!

Issue #115, opened 2021-03-08
[encoding] Unify string encoding for datasets and models


fg-mindee:
As of the latest commit, models are trained with a string encoder that is not particularly ordered (TF records). To clean all of this, I suggest the following:
- [x] Define pre-established vocab that we will use as encoders for your future models (#116)
- [ ] Address the topic of vocab mapping (converting an accented character to its raw version for instance)
- [x] Standardize dataset label encoding using a given vocab (#116)

On the topic of vocab definition, there are two aspects to consider: the character selection, their order in the vocab.

cc @charlesmindee

fg-mindee:
So, here are a few suggestions:
```
import string
latin = string.digits + string.ascii_letters + string.punctuation + '°'
french = latin + 'àâéèêëîïôùûçÀÂÉÈËÎÏÔÙÛÇ'
```
For reference, the 'latin' vocab would yield:
```
In [3]: latin
Out[3]: '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ!"#$%&\'()*+,-./:;<=>?@[\\]^_`{|}~°'
```

For later, I think it would be useful to have:
```
currency = '£€¥¢฿'
```

charlesmindee:
To address the 2nd point of the issue, I looked at Huggin Face Normalizer class:
[huggin face module-tokenizers.normalizers](https://huggingface.co/docs/tokenizers/python/latest/api/reference.html#module-tokenizers.normalizers). 
The normalizers perform conversion to lower case, accent stripping, chinese characters removal, spaces removal and contains basic unicode normalizers: NFC/D, NFKC/D. The detection on chinese characters [here](https://github.com/huggingface/tokenizers/blob/master/tokenizers/src/normalizers/bert.rs) seems hardcoded, but t he rest of the techniques use classical high-level functions (equivalent of unicodedata.normalize('NFC/D'), .lowercase(), ...)

fg-mindee:
Let's go with the option of having something specifying which alphabet is supported by which model in the document.
As of #238, we can now close this issue!

PR #116, opened 2021-03-08
feat: add string translating/encoding utilities

charlesmindee:
This PR is linked to issue #115, it implements:

- a translate function to translate a str sequence to another str sequence of another vocabulary
- an encoding function to encode an str sequence to an array to be fed in the model
- a decoding function to decode an array of int into a str sequence

A dictionnary of vocabularies is also provided

Any feeback is welcome!

### DIFF
--- a/doctr/datasets/utils.py
+++ b/doctr/datasets/utils.py
@@ -1,0 +1,97 @@
+# Copyright (C) 2021, Mindee.
+
+# This program is licensed under the Apache License version 2.
+# See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0.txt> for full license details.
+
+import string
+import unicodedata
+import numpy as np
+from typing import Dict
+
+__all__ = ['translate', 'encode_sequence', 'decode_sequence']
+
+
+vocabs: Dict[str, str] = {
+    'digits': string.digits,
+    'ascii_letters': string.ascii_letters,
+    'punctuation': string.punctuation,
+    'currency': '£€¥¢฿',
+    'latin': string.digits + string.ascii_letters + string.punctuation + '°',
+    'french': string.digits + string.ascii_letters + string.punctuation + '°' + 'àâéèêëîïôùûçÀÂÉÈËÎÏÔÙÛÇ' + '£€¥¢฿',
+    'tfrecords_training': '3K}7eé;5àÎYho]QwV6qU~W"XnbBvcADfËmy.9ÔpÛ*{CôïE%M4#ÈR:g@T$x?0î'
+                          '£|za1ù8,OG€P-kçHëÀÂ2É/ûIJ\'j(LNÙFut[)èZs+&°Sd=Ï!<â_Ç>rêi`l'
+}
+
+
+def translate(
+    input_string: str,
+    vocab: str,
+) -> str:
+    """Translate a string input in a given vocabulary
+
+    Args:
+        input_string: input string to translate
+        vocab: vocabulary to use (french, latin, ...)
+
+    Returns:
+        A string translated in a given vocab"""
+
+    if vocab not in vocabs.keys():
+        raise AttributeError("output vocabulary must be in vocabs dictionnary")
+
+    translated = ''
+    for char in input_string:
+        if char not in vocabs[vocab]:
+            # we need to translate char into a vocab char
+            if char in string.whitespace:
+                # remove whitespaces
+                continue
+            # normalize character if it is not in vocab
+            char = unicodedata.normalize('NFD', char).encode('ascii', 'ignore').decode('ascii')
+            if char == '' or char not in vocabs[vocab]:
+                # if normalization fails or char still not in vocab, return a black square (unknown symbol)
+                char = '■'
+        translated += char
+    return translated
+
+
+def encode_sequence(
+    input_string: str,
+    mapping: str,
+) -> np.array:
+    """Given a predefined mapping, encode the string to a sequence of numbers
+
+    Args:
+        input_string: string to encode
+        mapping: vocabulary (string), the encoding is given by the indexing of the character sequence
+
+    Returns:
+        A tensor encoding the input_string"""
+
+    encoded = []
+    for char in input_string:
+        if char not in mapping:
+            raise AttributeError("Input string vocabulary does not match encoding vocabulary")
+        encoded.append(mapping.index(char))
+    return np.array(encoded)
+
+
+def decode_sequence(
+    input_array: np.array,
+    mapping: str,
+) -> str:
+    """Given a predefined mapping, decode the sequence of numbers to a string
+
+    Args:
+        input_array: array to decode
+        mapping: vocabulary (string), the encoding is given by the indexing of the character sequence
+
+    Returns:
+        A string, decoded from input_array"""
+
+    if not input_array.dtype == np.int_ or max(input_array) >= len(mapping):
+        raise AttributeError("Input must be an array of int, with max less than mapping size")
+    decoded = ''
+    for index in list(input_array):
+        decoded += mapping[index]
+    return decoded
--- a/doctr/datasets/__init__.py
+++ b/doctr/datasets/__init__.py
@@ -1,0 +1,1 @@
+from .utils import *
--- a/test/test_datasets_utils.py
+++ b/test/test_datasets_utils.py
@@ -1,0 +1,36 @@
+import pytest
+from doctr import datasets
+
+
+@pytest.mark.parametrize(
+    "input_str, vocab, output_str",
+    [
+        ['f orêt', 'latin', 'foret'],
+        ['f or êt', 'french', 'forêt'],
+        ['¢¾©téØßřůž', 'french', '¢■■té■■ruz'],
+        ['Ûæëð', 'french', 'Û■ë■'],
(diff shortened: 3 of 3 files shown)


doctr/datasets/utils.py
```diff
@@ -0,0 +1,97 @@
...
+
+vocabs: Dict[str, str] = {
+    'digits': string.digits,
+    'ascii_letters': string.ascii_letters,
+    'punctuation': string.punctuation,
+    'currency': '£€¥¢฿',
+    'latin': string.digits + string.ascii_letters + string.punctuation + '°',
+    'french': string.digits + string.ascii_letters + string.punctuation + '°' + 'àâéèêëîïôùûçÀÂÉÈËÎÏÔÙÛÇ' + '£€¥¢฿',
+    'tfrecords_training': '3K}7eé;5àÎYho]QwV6qU~W"XnbBvcADfËmy.9ÔpÛ*{CôïE%M4#ÈR:g@T$x?0î'
+                          '£|za1ù8,OG€P-kçHëÀÂ2É/ûIJ\'j(LNÙFut[)èZs+&°Sd=Ï!<â_Ç>rêi`l'
```
fg-mindee, line 22: I suggest removing this, and keeping only "standard" vocab. This will still be specified in the default config of the model anyway

doctr/datasets/utils.py
```diff
@@ -0,0 +1,97 @@
...
+
+    Args:
+        input_string: input string to translate
+        vocab: vocabulary to use (french, latin, ...)
+
+    Returns:
+        A string translated in a given vocab"""
+
+    if vocab not in vocabs.keys():
+        raise AttributeError("output vocabulary must be in vocabs dictionnary")
```
fg-mindee, line 40: KeyError actually ;)

doctr/datasets/utils.py
```diff
@@ -0,0 +1,97 @@
...
+    """Translate a string input in a given vocabulary
+
+    Args:
+        input_string: input string to translate
+        vocab: vocabulary to use (french, latin, ...)
+
+    Returns:
+        A string translated in a given vocab"""
+
+    if vocab not in vocabs.keys():
```
fg-mindee, line 39:
```
if VOCABS.get(vocab) is None:
```
would be more optimal

doctr/datasets/utils.py
```diff
@@ -0,0 +1,97 @@
...
+
+import string
+import unicodedata
+import numpy as np
+from typing import Dict
+
+__all__ = ['translate', 'encode_sequence', 'decode_sequence']
+
+
+vocabs: Dict[str, str] = {
```
fg-mindee, line 14: if this is a constant, let's use upper case `VOCABS`

doctr/datasets/utils.py
```diff
@@ -0,0 +1,97 @@
...
+        if char not in vocabs[vocab]:
+            # we need to translate char into a vocab char
+            if char in string.whitespace:
+                # remove whitespaces
+                continue
+            # normalize character if it is not in vocab
+            char = unicodedata.normalize('NFD', char).encode('ascii', 'ignore').decode('ascii')
+            if char == '' or char not in vocabs[vocab]:
+                # if normalization fails or char still not in vocab, return a black square (unknown symbol)
+                char = '■'
```
fg-mindee, line 53: Mmmh, don't you think it would be safer for us to hardcode the conversion between vocabs?

doctr/datasets/utils.py
```diff
@@ -0,0 +1,97 @@
...
+        translated += char
+    return translated
+
+
+def encode_sequence(
+    input_string: str,
+    mapping: str,
+) -> np.array:
```
fg-mindee, line 61: Considering we don't actually use numpy for the computation, I'd suggest returning a `List[str]`

doctr/datasets/utils.py
```diff
@@ -0,0 +1,97 @@
...
+
+    Returns:
+        A tensor encoding the input_string"""
+
+    encoded = []
+    for char in input_string:
+        if char not in mapping:
+            raise AttributeError("Input string vocabulary does not match encoding vocabulary")
+        encoded.append(mapping.index(char))
+    return np.array(encoded)
```
fg-mindee, line 76: `return encoded`

doctr/datasets/utils.py
```diff
@@ -0,0 +1,97 @@
...
+        mapping: vocabulary (string), the encoding is given by the indexing of the character sequence
+
+    Returns:
+        A tensor encoding the input_string"""
+
+    encoded = []
+    for char in input_string:
+        if char not in mapping:
+            raise AttributeError("Input string vocabulary does not match encoding vocabulary")
+        encoded.append(mapping.index(char))
```
fg-mindee, line 75:
```
encoded = list(map(lambda x: mapping.index(x), input_string))
```
The `.index` method will raise a ValueError by itself if this isn't in the mapping

doctr/datasets/utils.py
```diff
@@ -0,0 +1,97 @@
...
+    """Given a predefined mapping, decode the sequence of numbers to a string
+
+    Args:
+        input_array: array to decode
+        mapping: vocabulary (string), the encoding is given by the indexing of the character sequence
+
+    Returns:
+        A string, decoded from input_array"""
+
+    if not input_array.dtype == np.int_ or max(input_array) >= len(mapping):
```
fg-mindee, line 92: input_array.max()

doctr/datasets/utils.py
```diff
@@ -0,0 +1,97 @@
...
+        raise AttributeError("Input must be an array of int, with max less than mapping size")
```
fg-mindee, line 93: AssertionError

AttributeError is for when you try to access the attribute of an object with the wrong name

doctr/datasets/utils.py
```diff
@@ -0,0 +1,97 @@
...
+    decoded = ''
+    for index in list(input_array):
+        decoded += mapping[index]
```
fg-mindee, line 96:
```
decoded = ''.join(mapping[idx] for idx in input_array)
```

doctr/datasets/utils.py
```diff
@@ -0,0 +1,97 @@
...
+
+# This program is licensed under the Apache License version 2.
+# See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0.txt> for full license details.
+
+import string
+import unicodedata
+import numpy as np
+from typing import Dict
+
+__all__ = ['translate', 'encode_sequence', 'decode_sequence']
```
fg-mindee, line 11: you might want to add VOCABS

doctr/datasets/utils.py
```diff
@@ -0,0 +1,97 @@
...
+    'latin': string.digits + string.ascii_letters + string.punctuation + '°',
+    'french': string.digits + string.ascii_letters + string.punctuation + '°' + 'àâéèêëîïôùûçÀÂÉÈËÎÏÔÙÛÇ' + '£€¥¢฿',
+    'tfrecords_training': '3K}7eé;5àÎYho]QwV6qU~W"XnbBvcADfËmy.9ÔpÛ*{CôïE%M4#ÈR:g@T$x?0î'
+                          '£|za1ù8,OG€P-kçHëÀÂ2É/ûIJ\'j(LNÙFut[)èZs+&°Sd=Ï!<â_Ç>rêi`l'
+}
+
+
+def translate(
+    input_string: str,
+    vocab: str,
```
fg-mindee, line 28: vocab_name ?

fg-mindee (commented):
Thanks for the PR! The general design is quite good, I added a few optimization suggestions!

doctr/datasets/utils.py
```diff
@@ -0,0 +1,89 @@
...
+        if char not in VOCABS[vocab_name]:
+            # we need to translate char into a vocab char
+            if char in string.whitespace:
+                # remove whitespaces
+                continue
+            # normalize character if it is not in vocab
+            char = unicodedata.normalize('NFD', char).encode('ascii', 'ignore').decode('ascii')
+            if char == '' or char not in VOCABS[vocab_name]:
+                # if normalization fails or char still not in vocab, return a black square (unknown symbol)
+                char = '■'
```
fg-mindee, line 51: Could you make the default unknown character a function's argument (default to this value)?

doctr/datasets/utils.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+VOCABS: Dict[str, str] = {
+    'digits': string.digits,
+    'ascii_letters': string.ascii_letters,
+    'punctuation': string.punctuation,
+    'currency': '£€¥¢฿',
+    'latin': string.digits + string.ascii_letters + string.punctuation + '°',
+    'french': string.digits + string.ascii_letters + string.punctuation + '°' + 'àâéèêëîïôùûçÀÂÉÈËÎÏÔÙÛÇ' + '£€¥¢฿',
+}
```
fg-mindee, line 21: Perhaps we should move this to `datasets/vocabs.py`?

This will most likely be extended often in the future (adding language specific characters, math symbols, etc.), perhaps more often then utils.py

fg-mindee (commented):
I suggested a small refactoring, let me know what you think!

fg-mindee (approved):
Thank for the edits!

PR #117, opened 2021-03-08
feat: Added text localization metric

fg-mindee:
- updated ExactMatch method name to something more explicit (result --> summary, update_state --> update)
- added implementation of LocalizationConfusion yielding recall, precision and mean IoU
- added numerous unittests
- updated requirements
- added metrics to documentation

Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR! Interesting way to optimize box matching by the way

Issue #118, opened 2021-03-08
ValueError on model call 


mohamedmindee:
## 🐛 Bug

When calling the model on a document, I get a ValueError. 

## To Reproduce

```
from doctr.documents import read_pdf 
from doctr.models import ocr_db_crnn 
model = ocr_db_crnn(pretrained=True)                                                                                                                                                                
doc = read_pdf("path/to/pdf") # write your path to the pdf 
result = model([doc]) 
```

```
~/tensorflow_2/lib/python3.6/site-packages/doctr/models/core.py in __call__(self, documents, **kwargs)
     53         # Reorganize
     54         num_pages = [len(doc) for doc in documents]
---> 55         results = self.doc_builder(boxes, char_sequences, num_pages, [page.shape[:2] for page in pages])
     56 
     57         return results

~/tensorflow_2/lib/python3.6/site-packages/doctr/models/core.py in __call__(self, boxes, char_sequences, num_pages, page_shapes)
    204                         self._build_blocks(
    205                             page_boxes[:num_crops[page_idx]],
--> 206                             char_sequences[crop_idx: crop_idx + num_crops[page_idx]]
    207                         ),
    208                         page_idx,

~/tensorflow_2/lib/python3.6/site-packages/doctr/models/core.py in _build_blocks(self, boxes, char_sequences)
    164                         ((boxes[idx, 0], boxes[idx, 1]), (boxes[idx, 2], boxes[idx, 3]))
    165                     ) for idx in line]
--> 166                 ) for line in lines]
    167             )
    168         ]

~/tensorflow_2/lib/python3.6/site-packages/doctr/models/core.py in <listcomp>(.0)
    164                         ((boxes[idx, 0], boxes[idx, 1]), (boxes[idx, 2], boxes[idx, 3]))
    165                     ) for idx in line]
--> 166                 ) for line in lines]
    167             )
    168         ]

~/tensorflow_2/lib/python3.6/site-packages/doctr/documents/elements.py in __init__(self, words, geometry)
    108         # Resolve the geometry using the smallest enclosing bounding box
    109         if geometry is None:
--> 110             geometry = resolve_enclosing_bbox([w.geometry for w in words])
    111 
    112         super().__init__(words=words)

~/tensorflow_2/lib/python3.6/site-packages/doctr/utils/geometry.py in resolve_enclosing_bbox(bboxes)
     20 
     21 def resolve_enclosing_bbox(bboxes: List[BoundingBox]) -> BoundingBox:
[...]

PR #119, opened 2021-03-08
fix: Fixed edge case of result export when there is no detection

fg-mindee:
Following up on #118, this PR introduces the following modifications:
- fixed block building when there is no detection in the page
- added a corresponding unittest

Any feedback is welcome!


charlesmindee (approved):
Thanks for the fix!

PR #126, opened 2021-03-10
fix: Fixed docstring examples and added predictor dimension checks

fg-mindee:
- fixed docstring examples for predictors
- added a dimension check in predictor's call method
- added unittest for this check

Any feedback is welcome!

### DIFF
--- a/doctr/models/core.py
+++ b/doctr/models/core.py
@@ -41,6 +41,10 @@ class OCRPredictor(NestedObject):
         **kwargs: Any,
     ) -> List[Document]:
 
+        # Dimension check
+        if any(page.ndim != 3 for doc in documents for page in doc):
+            raise ValueError("incorrect input shape: all documents are expected to be list of multi-channel 2D images.")
+
         pages = [page for doc in documents for page in doc]
 
         # Localize text elements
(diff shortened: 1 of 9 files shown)


doctr/models/core.py
```diff
@@ -41,6 +41,10 @@ def __call__(
         **kwargs: Any,
     ) -> List[Document]:
 
+        # Dimension check
+        if any(page.ndim != 3 for doc in documents for page in doc):
+            raise ValueError("incorrect input shape: all documents are expected to be list of multi-channel 2D images.")
```
charlesmindee, line 46: We should handle the 1 channel case

charlesmindee: at least a duplication of the channel thrice to simulate a RBG image

charlesmindee (approved):
Thanks for the fixs! As I mentionned in the feedback I think we should handle the 1 channel case, at least duplicate the channels thrice to simulate a RGB image

doctr/models/core.py
fg-mindee: Not sure I follow? I'm checking the number of dimensions / axes here, not the number of channels on the channel axis. Or are you suggesting to add a channel check? (if so, it depends on the architecture, we can easily imagine a model using grayscale)

charlesmindee: ok sorry I thought you were checking the number of channel on the channel axis

charlesmindee (approved):
Thanks!

Issue #127, opened 2021-03-10
[metrics] Benchmark python options for text-distance computing


charlesmindee:
**Options for computing distance between 2 character sequences in python:**

- textdistance: full python lib
- jellyfish: full python lib
- strsimpy: full python lib
- python-Levenshtein: C lib
- edlib: C++ lib (python binding)
- RapidFuzz: C++/python lib
- polyleven: C/python lib

All these libs provides algorithms such as Levenshtein distance, Hamming distance, Jaro-Winkler distance, ... To compute distance between 2 strings (character sequences).

This graph taken from the [RapidFuzz documentation](https://maxbachmann.github.io/RapidFuzz/string_metric.html) highlights the strong dependence of runtime on string length:

![Capture d’écran de 2021-03-10 16-00-17](https://user-images.githubusercontent.com/70526046/110651405-b32efa80-81bb-11eb-8414-7b4f468e1110.png)

We can see for example that python-Levenshtein is faster on short sequences than edlib, but much slower on long sequences. **In our cases, we are typically using character sequences from 0 to 30 characters, very often between 0 and 15 characters and almost never more than 30 characters**. According to these plots, it seems that polyleven, python-Levenshtein and rapidfuzzy are the fastest solutions on these typical lengths. 

I conducted a short study: I called levenshtein distance 1000 times for each of these package, on 2 strings between 15 and 20 characters. Here are the results for 1000 iterations:

- RapidFuzz / python-Levenshtein: 0.0002 s
- Polyleven: 0.0003 s
- Jellyfish: 0.0009 s
- edlib: 0.002 s
- textdistance: 0.0064 s
- strsimpy: 0.09 s

**Conclusion:**
Jellyfish is fast for a full-python lib. However, as expected C libs are faster, especially **RapidFuzz and  python-Levenshtein on these sequence lengths**. In terms of dependencies, python-Levenshtein is really light but rapidfuzz is heavier (cpp files)

@fg-mindee

fg-mindee:
Thanks a lot for the investigation!
Among the two best candidates ([rapidfuzz](https://github.com/maxbachmann/RapidFuzz) & [python levenshtein](https://github.com/miohtama/python-Levenshtein#maintainer-needed)) , I checked a few additional things:
- ease of installation: pip available for both, but rapidfuzz also has a conda build
- dependency complexity: C or C++ compilation for both
- maintenance: apparently python levenshtein is looking for maintainers while last commit on rapidfuzz was 3 hours ago

[...]

PR #128, opened 2021-03-10
refactor: changed Levenshtein distance package (textdistance --> RapidFuzz)

charlesmindee:
This PR switches packages (textdistance --> RapidFuzz) for computing Levenshtein distance between character sequences for end to end metric. Benchmark #127 has led to this change.


fg-mindee (approved):
Thanks a lot for this dependency switch!

PR #130, opened 2021-03-11
fix: dbnet default postprocessor params

charlesmindee:
This PR adjusts default postprocessing parameters (bin_threshold) for dbnet. It is a temporary fix before training with properly normalized dataset.

Previous post-processing:
old
New postprocessing:
new


fg-mindee:
Would you mind posting the difference in the outcome of the detection predictor for the same image please?

fg-mindee (approved):
Thanks!

PR #132, opened 2021-03-11
feat: add resnet-31 backbone for SAR

charlesmindee:
This PR plugs resnet31 backbone on SAR model.
Any feedback is welcome!


fg-mindee (approved):
Thanks a lot!

PR #133, opened 2021-03-12
feat: add mean & std for detection preprocessor

charlesmindee:
This PR adds mean & std of our dataset distribution in detection preprocessor to normalize dataset. These values are computed channel-wise on the whole dataset. Any feedback is welcome !


test/test_models_detection.py
```diff
@@ -32,7 +32,7 @@ def test_detpreprocessor(mock_pdf):  # noqa: F811
     assert batched_docs[-1].shape[0] == (8 * num_docs) % batch_size
 
     # Repr
-    assert repr(processor) == 'DetectionPreProcessor(output_size=(512, 512), mean=[0.5 0.5 0.5], std=[1. 1. 1.])'
+    assert repr(processor) == 'DetectionPreProcessor(output_size=(512, 512), mean=[0.798, 0.785, 0.772], std=[0.264, 0.274, 0.287])'
```
fg-mindee: The repr does not include the commas between figures ;)

fg-mindee: And it's too long for flake8

fg-mindee (changes requested):
Thanks for the PR! A small edit is required to avoid unnecessary breaking change (cf. comments)

test/test_models_detection.py
charlesmindee: yes I changed it

doctr/models/detection/core.py
```diff
@@ -35,8 +35,8 @@ def __init__(
         self,
         output_size: Tuple[int, int],
         batch_size: int = 1,
-        mean: Tuple[float, float, float] = (.5, .5, .5),
-        std: Tuple[float, float, float] = (1., 1., 1.),
+        mean: Tuple[float, float, float] = (0.798, 0.785, 0.772),
+        std: Tuple[float, float, float] = (0.264, 0.274, 0.287),
```
fg-mindee, line 39: We need to specify these in the model's default dict. If we change the default values here, the previously trained models won't work anymore

doctr/models/detection/core.py
```diff
@@ -35,7 +35,7 @@ def __init__(
         self,
         output_size: Tuple[int, int],
         batch_size: int = 1,
-        mean: Tuple[float, float, float] = (.5, .5, .5),
+        mean: Tuple[float, float, float] = (0.5, 0.5, 0.5),
```
fg-mindee: No need for the zeros before the '.' (so that the PR doesn't display any diff on this file)

test/test_models_detection.py
```diff
@@ -32,7 +32,8 @@ def test_detpreprocessor(mock_pdf):  # noqa: F811
     assert batched_docs[-1].shape[0] == (8 * num_docs) % batch_size
 
     # Repr
-    assert repr(processor) == 'DetectionPreProcessor(output_size=(512, 512), mean=[0.5 0.5 0.5], std=[1. 1. 1.])'
+    assert repr(processor) == 'DetectionPreProcessor(output_size=(512, 512), ' \
+        'mean=[0.5 0.5 0.5], std=[1. 1. 1.])'
```
fg-mindee: you can revert all changes

doctr/models/detection/zoo.py
```diff
@@ -8,10 +8,19 @@
...
+    'db_resnet50_predictor_normed': {
+        'model': 'db_resnet50',
+        'post_processor': 'DBPostProcessor',
+        'mean': (0.798, 0.785, 0.772),
+        'std': (0.264, 0.274, 0.287),
+    },
+    'db_resnet50_predictor': {
+        'model': 'db_resnet50',
+        'post_processor': 'DBPostProcessor',
+    },
```
fg-mindee, line 23: Actually, for now we have no "db_resnet50_predictor_normed" so I suggest:
- we just specify the old mean & std for "db_resnet50" in `differentiable_binarization.py`
- once we have a new (supposedly better) pretrained db_resnet50 with proper normalization, we upload the checkpoint, edit the mean, std, and url. (long term we have no need for lesser versions of the same model)

fg-mindee: so you don't need to modify anything in detection zoo apart from: passing
`mean=_model.cfg['mean'], std=_model.cfg['std']` to the `DetectionPreProcessor` in `_predictor`

test/test_models_detection.py
```diff
@@ -112,6 +113,7 @@ def test_detectionpredictor(mock_pdf):  # noqa: F811
     "arch_name",
     [
         "db_resnet50_predictor",
+        "db_resnet50_predictor_normed",
```
fg-mindee, line 115: can be removed

fg-mindee (changes requested):
Sorry I wasn't clear earlier on: since we don't yet have a pretrained models for this properly normalized version, we don't change anything about this. However, we change prepare the change by specifying in differentiable_binarization.py default_config the mean & std, then correctly passing them to the preprocessor in the `_predictor` of the detection zoo

doctr/models/detection/zoo.py
```diff
@@ -46,3 +55,23 @@ def db_resnet50_predictor(pretrained: bool = False, **kwargs: Any) -> DetectionP
...
+        >>> out = model([input_page])
+
+    Args:
+        pretrained (bool): If True, returns a model pre-trained on ImageNet
+
+    Returns:
+        Detection predictor
+    """
+
+    return _predictor('db_resnet50_predictor_normed', pretrained, **kwargs)
```
fg-mindee, line 77: Can be removed

fg-mindee (approved):
Thanks a lot!

Issue #124, opened 2021-03-09
[docs] Add performance benchmark for all pretrained models


fg-mindee:
There are a few tables in the documentation that still need filling:
- [x] Text detection (#143)
- [x] Text recognition (#143)
- [x] End-to-End OCR (#143)
- [ ] Comparison with similar solutions (#149)

We need to select a public dataset for each task and run the evaluation on their respective test sets, then report the results back in these tables.

PR #136, opened 2021-03-12
feat: Added FUNSD dataset

fg-mindee:
- added an abstract VisionDataset class
- added the FUNSD dataset inherited from VisionDataset
- added it to the documentation
- refactored verified file downloading functions to use it both for model and dataset downloading

Any feedback is welcome!

### DIFF
--- a/doctr/datasets/core.py
+++ b/doctr/datasets/core.py
@@ -34,7 +34,7 @@ class VisionDataset:
 
         archive_path = download_from_url(url, file_name, file_hash, cache_subdir='datasets')
 
-        # Extract to funsd
+        # Extract the archive
         if extract_archive:
             archive_path = Path(archive_path)
             dataset_path = archive_path.parent.joinpath(archive_path.stem)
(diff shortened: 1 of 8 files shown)


doctr/datasets/core.py
```diff
@@ -0,0 +1,56 @@
...
+
+        # Download the file if not present
+        archive_path = os.path.join(dataset_cache, file_name)
+
+        if not os.path.exists(archive_path) and not download:
+            raise ValueError("the dataset needs to be downloaded first with download=True")
+
+        archive_path = download_from_url(url, file_name, file_hash, cache_subdir='datasets')
+
+        # Extract to funsd
```
charlesmindee, line 37: why do you mention funsd on this meta-class ?

charlesmindee (approved):
Thanks for the PR !

doctr/datasets/core.py
fg-mindee: Good catch, it was a refactoring, I just fixed it!

charlesmindee (approved):
Thanks!

PR #137, opened 2021-03-13
feat: Simplified repr of NestedObject when they have an empty list of children

fg-mindee:
- simplified empty child list repr of NestedObject
- adding a unittest to check this

Before this PR, the following snippet:
```
from doctr.documents.elements import Line
# Pass empty list of words
print(Line([], ((0, 0), (1, 1))))
```

would yield
```
Line(
  (words): [
  ]
)

```
while now it yields
```
Line(
  (words): []
)
```

Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR !

PR #138, opened 2021-03-13
docs: Fixed multi-versions doc build

fg-mindee:
This PR fixes multi-version documentation build from CI. 

Currently, the CI has a shallow view of git history, so a `git checkout <HASH>` fails as the history is shallow. 
You can notice the `error: pathspec '571af3dc' did not match any file(s) known to git` error message in the doc deployment [action](https://github.com/mindee/doctr/runs/2096643999?check_suite_focus=true). The consequence of that, is that all versions of the doc that are deployed are the same: cf. https://mindee.github.io/doctr/

I added a `git fetch` to fix this.

Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR!

PR #140, opened 2021-03-15
fix: edge case intersection iou metric

charlesmindee:
This PR fixes iou intersection computation for edge cases (clipping by (0, inf) the right-left and bot-top arrays to ensure that intersection will be set to 0 when boxes are not overlapping)

### DIFF
--- a/test/test_utils_metrics.py
+++ b/test/test_utils_metrics.py
@@ -56,6 +56,7 @@ def test_box_iou(box1, box2, iou, abs_tol):
         [[[0, 0, .5, .5]], [[0, 0, .5, .5]], 0.5, 1, 1, 1],  # Perfect match
         [[[0, 0, 1, 1]], [[0, 0, .5, .5], [.6, .6, .7, .7]], 0.2, 1, 0.5, 0.125],  # Bad match
         [[[0, 0, 1, 1]], [[0, 0, .5, .5], [.6, .6, .7, .7]], 0.5, 0, 0, 0.125],  # Bad match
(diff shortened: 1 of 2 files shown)


fg-mindee (changes requested):
I suggest adding a unittest in test_box_iou for this:
```
[[0, 0, .1, .1], [.9, .9, 1, 1], 0, 0],  # Boxes far from each other
```

test/test_utils_metrics.py
```diff
@@ -56,6 +56,7 @@ def test_box_iou(box1, box2, iou, abs_tol):
         [[[0, 0, .5, .5]], [[0, 0, .5, .5]], 0.5, 1, 1, 1],  # Perfect match
         [[[0, 0, 1, 1]], [[0, 0, .5, .5], [.6, .6, .7, .7]], 0.2, 1, 0.5, 0.125],  # Bad match
         [[[0, 0, 1, 1]], [[0, 0, .5, .5], [.6, .6, .7, .7]], 0.5, 0, 0, 0.125],  # Bad match
+        [[0, 0, .1, .1], [.9, .9, 1, 1], 0, 0],  # Boxes far from each other
```
fg-mindee, line 59: This isn't the right unittest :sweat_smile:

fg-mindee: You need to add this to test_box_iou

fg-mindee (changes requested):
Small issue on the unittest where you added a case: it's not the right one :sweat_smile:

fg-mindee (approved):
Thanks!

PR #141, opened 2021-03-15
feat: Added evaluation script

fg-mindee:
- added image reading in the dataset indexing method
- updated FUNSD unittest
- renamed metric args
- enabled GPU on analyze script
- added evaluation script

Any feedback is welcome

### DIFF
--- a/doctr/datasets/funsd.py
+++ b/doctr/datasets/funsd.py
@@ -5,6 +5,7 @@
 
 import os
 import json
+import numpy as np
 from pathlib import Path
 from typing import List, Dict, Any, Tuple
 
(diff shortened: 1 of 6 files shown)


charlesmindee (approved):
Thanks for the PR ! The script is only for end to end evaluation, are we going to upload detection-only and recognition-only scripts also ?

fg-mindee:
> Thanks for the PR ! The script is only for end to end evaluation, are we going to upload detection-only and recognition-only scripts also ?

I think we should leverage the end-to-end predictor and do the evaluation at the same time! But in another PR

PR #142, opened 2021-03-15
chore: Added a CI job to check that analysis script works

fg-mindee:
- added a non blocking viz option in analysis script (to avoid blocking the CI)
- added a CI job to run the analysis script on a simple 1-page pdf

Any feedback is welcome!


charlesmindee (approved):
thanks for the PR !

PR #145, opened 2021-03-15
feat: Added detection and recognition metric to the evaluation script

fg-mindee:
- extended support of crop extraction to absolute coordinates
- updated unittests
- added detection and recognition metrics to the evaluation script

Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR !

PR #146, opened 2021-03-15
feat: Added minimal demo app for text detection

fg-mindee:
- added demo app for text detection
- added app running instructions in the README

Here is how the app renders after performing text detection on a sample pdf:
doctr_demo

Any feedback is welcome!


fg-mindee:
FYI: I initially added a CI job to make sure the app runs but since this is a blocking process, I would have to run it in a docker for the job to avoid running indefinitely.

charlesmindee (approved):
Thanks for the PR !

PR #150, opened 2021-03-16
feat: update model cfg with mean & std

charlesmindee:
- add new DB weights & mean/std values in cfg
- add sar_resnet31 weights
- add mean & std in cfg for recognition task, and set these values for each model cfg
- changed DB postprocessor parameters accordingly to new normalization
- add perf in documentation

Any feedback is welcome!


fg-mindee (commented):
Thanks very much! Would you mind running the evaluation with this new version and update the entries in the doc accordingly? :pray:

doctr/models/zoo.py
```diff
@@ -9,10 +9,11 @@
 from .recognition import zoo as reco_zoo
 
 
-__all__ = ["ocr_db_sar", "ocr_db_crnn"]
+__all__ = ["ocr_db_sar", "ocr_db_crnn", "ocr_db_sar_resnet"]
 
 default_cfgs: Dict[str, Dict[str, Any]] = {
     'ocr_db_sar': {'detection': 'db_resnet50_predictor', 'recognition': 'sar_vgg16_bn_predictor'},
```
fg-mindee, line 15: Could you rename ocr_db_sar into ocr_db_sar_vgg for clarity?
Also you need to update the documentation and add the new entries in the unittests ;)

fg-mindee (changes requested):
Almost there! One renaming, few things to update in the documentation and unittest

fg-mindee (changes requested):
My bad , you need to update the "example" in ocr zoo, the README instructions, and the default model for the analysis script :pray:

fg-mindee (approved):
Thanks!

PR #151, opened 2021-03-16
fix: Fixed evaluation script

fg-mindee:
This PR fixes the evaluation script, by looping on both subsets. The previous patching was not taking into account the different roots of both subsets.

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #152, opened 2021-03-17
feat: Added proper kernel sizing to silence loading warnings

fg-mindee:
- updated DBNet, CRNN & SAR to correctly size the kernel of the entire model
- used this to save checkpoints correctly that won't trigger annoying warnings like previously

(I checked the analyze script result and they are unchanged)
Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR !

PR #155, opened 2021-03-17
fix: postprocessor parameter

charlesmindee:
This PR fixes the postprocessing of detection model (adjusting the small boxes filter parameter) and reports new perfomances in documentation. 
Any feedback is welcome !


fg-mindee (approved):
Thanks for the fix!

PR #156, opened 2021-03-17
fix: Fixed pretrained models

fg-mindee:
- Reverted to old checkpoints
- Fixed CRNN architecture
- Partially fixed SAR (inference going smoothly as before, but from the pretrained model, saving the weights and reloading it make inference go nuts :man_shrugging: )

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #157, opened 2021-03-18
fix: cv2.findContours param

charlesmindee:
This PR improves DB postprocessor using the RETR_EXTERNAL flag in cv2 function to detect only external contours.
Any feedback is welcome !

With RETR_LIST:
avant

With RETR_EXTRENAL:
Capture du 2021-03-18 12-04-08


fg-mindee (approved):
Thanks for the PR! For reference, if you have some, would you mind adding screenshots of the difference in terms of localization?

PR #158, opened 2021-03-18
docs: Added documentation about model evaluation

fg-mindee:
- specified on which dataset the evaluation runs were conducted
- added disclaimer about the low dataset size
- added detailed explanation about metric computation for each task

Here is how the documentation renders for `OCRMetric`:
image

Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR !

PR #160, opened 2021-03-23
feat: add crnn_resnet31

charlesmindee:
This PR adds ocr_db_crnn_resnet in the model zoo and reports the evaluation of the model on the test set; same results than ocr_db_sar_resnet. It also corrects a typo in the SAR init.
Any feedback is welcome!


fg-mindee (approved):
Thanks for the new addition to the model zoo!

PR #161, opened 2021-03-23
feat: add core detection data loader

charlesmindee:
This PR implements core detection data loader. Any feedback is welcome!


doctr/datasets/loaders/detection/core.py
```diff
@@ -0,0 +1,147 @@
...
+        labels = json.load(f)
+
+    polys = [[[int(x), int(y)] for [x, y] in polygon] for polygon in labels["boxes_1"] + labels["boxes_2"]]
+    polys_mask = [[[int(x), int(y)] for [x, y] in polygon] for polygon in labels["boxes_3"]]
+
+    return polys, polys_mask
+
+
+def resize(image, input_size):
+    return cv2.resize(image, input_size, interpolation=cv2.INTER_AREA)
```
fg-mindee, line 40: It might be best to stick to TF ops when possible: https://github.com/mindee/doctr/blob/main/doctr/models/detection/core.py#L58

doctr/datasets/loaders/detection/core.py
```diff
@@ -0,0 +1,147 @@
...
+        self.batch_size = batch_size
+        self.images_path = images_path
+        self.labels_path = labels_path
+        self.shuffle = shuffle
+        self.on_epoch_end()
+        self.min_size_box = 3
+
+    def __len__(self):
+        # Denotes the number of batches per epoch
+        return int(np.floor(len(os.listdir(self.images_path)) / self.batch_size))
```
fg-mindee, line 72: Are we positive that when it's not a multiple of the batch size, it does not create one last unfull batch? (if so `math.ceil` might be useful)

doctr/datasets/loaders/detection/core.py
```diff
@@ -0,0 +1,147 @@
...
+        self.indexes = np.arange(len(os.listdir(self.images_path)))
+        if self.shuffle is True:
+            np.random.shuffle(self.indexes)
+
+    def __getitem__(
+        self,
+        index: int
+    ) -> Tuple[tf.Tensor, np.array, np.array]:
+        # Get one batch of data
+        indexes = self.indexes[index * self.batch_size:(index + 1) * self.batch_size]
```
fg-mindee, line 85: Same here, we need some safe guard for the last batch

doctr/datasets/loaders/detection/core.py
```diff
@@ -0,0 +1,147 @@
...
+        # Find list of paths
+        list_paths = [os.listdir(self.images_path)[k] for k in indexes]
+        # Generate data
+        return self.__data_generation(list_paths)
+
+    def __data_generation(
+        self,
+        list_paths: List[str],
+    ) -> Tuple[tf.Tensor, np.array, np.array]:
```
fg-mindee, line 94: I suggest going full `tf.Tensor`

doctr/datasets/loaders/detection/core.py
```diff
@@ -0,0 +1,147 @@
...
+            mask = resize(mask, self.input_size)
+
+            batch_images[index, ] = image
+            batch_gts[index, ] = gt
+            batch_masks[index, ] = mask
+
+        # Normalize
+        batch_images = tf.stack(batch_images, axis=0)
+        std = tf.cast((0.287, 0.274, 0.264), tf.float32)
+        mean = tf.cast((0.772, 0.785, 0.798), tf.float32)
```
fg-mindee, line 144: This has to be stored as a class attribute

doctr/datasets/loaders/detection/core.py
```diff
@@ -0,0 +1,147 @@
...
+    """Loads detection annotations (boxes) for an image, from a folder containing
+    annotations in json files for each image
+
+    Agrs:
+        labels_path: path to the folder containing all json annotations
+        img_name: name of the image to find the corresponding json annotation
+
+    Returns:
+        A tuple of 2 lists: a list of polygons to draw on the segmentation gt map,
+        and a list of ambiguous polygons to mask during the training
```
fg-mindee, line 27: I think it might be better, to keep all boxes together, but having a boolean vector saying whether the box is ambiguous or not (it will be easier to batch, and thus easier to avoid taking it into account in the loss later on)

fg-mindee (commented):
Thanks for the PR! I added a few comments to the design suggestion

doctr/datasets/loaders/detection/core.py
```diff
@@ -23,21 +23,15 @@ def load_annotation(
...
     with open(os.path.join(labels_path, img_name + '.json')) as f:
         labels = json.load(f)
 
-    polys = [[[int(x), int(y)] for [x, y] in polygon] for polygon in labels["boxes_1"] + labels["boxes_2"]]
-    polys_mask = [[[int(x), int(y)] for [x, y] in polygon] for polygon in labels["boxes_3"]]
-
-    return polys, polys_mask
-
+    polys = [[[int(x), int(y)] for [x, y] in polygon] for polygon in labels["boxes_1"] + labels["boxes_2"] + labels["boxes_3"]]
+    to_masks = [False for polygon in labels["boxes_1"] + labels["boxes_2"]] + [True for polygon in labels["boxes_3"]]
```
fg-mindee:
```
to_masks = [False] * (len(labels["boxes_1"]) + len(labels["boxes_2"])) + [True] * len(labels["boxes_3"])
```

doctr/datasets/loaders/detection/core.py
```diff
@@ -66,10 +60,12 @@ def __init__(
         self.shuffle = shuffle
         self.on_epoch_end()
         self.min_size_box = 3
+        self.std_rgb = (0.264, 0.274, 0.287)
+        self.mean_rgb = (0.798, 0.785, 0.772)
```
fg-mindee, line 66: They need to be passed as constructor arguments as well

doctr/datasets/loaders/detection/core.py
```diff
@@ -106,42 +102,46 @@ def __data_generation(
...
-        # Normalize
+        # Stack batches
         batch_images = tf.stack(batch_images, axis=0)
-        std = tf.cast((0.287, 0.274, 0.264), tf.float32)
-        mean = tf.cast((0.772, 0.785, 0.798), tf.float32)
+        batch_gts = tf.stack(batch_gts, axis=0)
+        batch_masks = tf.stack(batch_masks, axis=0)
+        # Normalize
+        std = tf.cast(self.std_rgb, tf.float32)
+        mean = tf.cast(self.mean_rgb, tf.float32)
```
fg-mindee, line 146: You can cast them into TF tensors in the constructors since this doesn't change afterwards

fg-mindee (commented):
Good edits! A few extra suggestions

doctr/datasets/loaders/detection/core.py
```diff
@@ -54,6 +54,8 @@ def __init__(
         labels_path: str,
         batch_size: int = 1,
         shuffle: bool = True,
+        std_rgb: Tuple[float, float, float] = (0.264, 0.274, 0.287),
+        std_mean: Tuple[float, float, float] = (0.798, 0.785, 0.772),
```
fg-mindee, line 62: Typo here 
Let's name them mean & std adn specify in the docstring that they are in RGB order

test/conftest.py
```diff
@@ -26,3 +27,35 @@ def mock_pdf(mock_pdf_stream, tmpdir_factory):
...
+    label = {
+        "boxes_1": [[[1, 2], [1, 3], [2, 1], [2, 3]], [[10, 20], [10, 30], [20, 10], [20, 30]]],
+        "boxes_2": [[[3, 2], [3, 3], [4, 1], [4, 3]], [[30, 20], [30, 30], [40, 10], [40, 30]]],
+        "boxes_3": [[[1, 5], [1, 8], [2, 5], [2, 8]]],
+    }
+    for i in range(5):
+        fn = folder.join("mock_image_file_" + str(i) + ".jpeg.json")
+        with open(fn, 'w') as f:
+            json.dump(label, f)
+    return str(folder)
```
fg-mindee, line 61: This part should be moved in `test_datasets_loaders.py` for clarity (it won't be used by others, while the image folder could be useful elsewhere)

test/test_datasets_loaders.py
```diff
@@ -0,0 +1,18 @@
...
+    core_loader = loaders.detection.core.DataGenerator(
+        input_size=(1024, 1024),
+        images_path=mock_image_folder,
+        labels_path=mock_detection_label,
+        batch_size=1,
+    )
+    assert core_loader.__len__() == 5
+    for batch_step, batch in enumerate(core_loader):
+        image, gt, mask = batch
+        assert isinstance(image, tf.Tensor)
```
fg-mindee, line 16: I suggest also checking the image size

test/conftest.py
```diff
@@ -26,3 +27,35 @@ def mock_pdf(mock_pdf_stream, tmpdir_factory):
...
+
+
+@pytest.fixture(scope="session")
+def mock_detection_label(tmpdir_factory):
+    folder = tmpdir_factory.mktemp("labels")
+    label = {
+        "boxes_1": [[[1, 2], [1, 3], [2, 1], [2, 3]], [[10, 20], [10, 30], [20, 10], [20, 30]]],
+        "boxes_2": [[[3, 2], [3, 3], [4, 1], [4, 3]], [[30, 20], [30, 30], [40, 10], [40, 30]]],
+        "boxes_3": [[[1, 5], [1, 8], [2, 5], [2, 8]]],
```
fg-mindee, line 55: I don't understand was those "boxes_1", "boxes_2" are ?
Why do we need this as a key? Should'nt it be the name of the image?

fg-mindee (commented):
This is starting to be rather good, a few comments left :ok_hand:

test/test_datasets_loaders.py
```diff
@@ -1,6 +1,23 @@
 from doctr.datasets import loaders
 from typing import Tuple
 import tensorflow as tf
+import pytest
+import json
+
+
+@pytest.fixture(scope="session")
```
fg-mindee, line 7: `scope="function"`

doctr/datasets/loaders/detection/core.py
```diff
@@ -84,7 +84,7 @@ def on_epoch_end(self):
     def __getitem__(
         self,
         index: int
-    ) -> Tuple[tf.Tensor, tf.Tensor, tf.Tensor]:
+    ) -> Tuple[Any]:
```
fg-mindee: Mmmh, perhaps when we use the inheritance, this will have to change, but in this PR, I think that 
`) -> Tuple[tf.Tensor, tf.Tensor, tf.Tensor]:` is the only correct possibility

fg-mindee (commented):
Nitpicking but most of it is fine by me!

fg-mindee (approved):
Looks good to me, thanks!

PR #165, opened 2021-03-24
feat: Added SROIE dataset from ICDAR2019

fg-mindee:
- added SROIE class for a cleaned version of the original dataset (removed duplicates, images without annotation, etc.)
- added corresponding entry in the documentation
- added corresponding unittest
- added default file name downloading resolution
- fixed cache folder location in download_url

Any feedback is welcome!

### DIFF
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -8,6 +8,7 @@ from doctr import datasets
     "dataset_name, size",
     [
         ['FUNSD', 149],
+        ['SROIE', 626],
     ],
 )
 def test_dataset(dataset_name, size):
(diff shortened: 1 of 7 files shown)


test/test_datasets.py
```diff
@@ -8,6 +7,7 @@
     "dataset_name, size",
     [
         ['FUNSD', 149],
```
charlesmindee, line 9: are we testing only on train sets ?

fg-mindee: agreed, I'll update this!

charlesmindee (approved):
Thanks!

PR #170, opened 2021-03-25
refactor: Changed predictor call args to list of pages

### LINKED ISSUE #168: [models] Models should accept list of pages as inputs and not list of documents
The current behaviour overcomplicates things since a document is already a list of pages.
The only advantage of the current behaviour is to save potentially 1 forward. Assuming we have 1 document with A pages and a second document with B pages, and our batch size N:
- currently, we do `math.ceil((A + B) / N)` forwards
- the proposition would have `math.ceil(A / N) + math.ceil(B / N)` forwards

The real advantage is when A and B are less than N.

The interface would be much cleaner though.

fg-mindee:
- switched OCRPredictor input to a list of pages and output to a `Document`
- updated unittests, usage instructions, documentation and scripts accordingly

Resolves #168 

Any feedback is welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,12 +59,12 @@ from doctr.models import ocr_model
 model = ocr_model(pretrained=True)
 # PDF
 doc = read_pdf("path/to/your/doc.pdf")
-result = model([doc])
+result = model(doc)
 # Image
 page = read_img("path/to/your/img.jpg")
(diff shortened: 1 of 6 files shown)


charlesmindee (approved):
Thanks for the PR

PR #176, opened 2021-03-29
fix: Fixed IoU computation for zero-sized inputs

fg-mindee:
- only performs the computation of IoU when both inputs are not zero-sized
- added unittests accordingly

Any feedback is welcome!


charlesmindee (approved):
Thanks for the fix!

PR #177, opened 2021-03-29
fix: typo evaluation script + edge case metrics + resize

charlesmindee:
This PR fixes a typo in evaluate.py and 2 edge cases:
- resize a tf tensor with a 0-dimension
- set iou to 0 when iou matrix is empty

Any feedback is welcome!


scripts/evaluate.py
```diff
@@ -10,7 +10,7 @@
 os.environ["TF_CPP_MIN_LOG_LEVEL"] = "2"
 
 from doctr.utils.metrics import LocalizationConfusion, ExactMatch, OCRMetric
-from doctr.datasets import FUNSD
+from doctr.datasets import FUNSD, SROIE
```
fg-mindee, line 13: let's avoid importing it since we don't use it for now

fg-mindee (commented):
Only a small import to remove and we're good to merge!

scripts/evaluate.py
charlesmindee: ok

fg-mindee (approved):
Thanks for the fix!

PR #178, opened 2021-03-29
feat: add detection training script

charlesmindee:
This PR implements a detection training script for doctr models, and a readme to properly use the script.


references/detection/train.py
```diff
@@ -35,7 +36,7 @@ def main(args):
     optimizer = tf.keras.optimizers.Adam(learning_rate=args.learning_rate, clipnorm=5)
 
     # Load doctr model
-    model = detection.__dict__[args.model](pretrained=False, input_shape=args.input_size)
+    model = detection.__dict__[args.model](pretrained=False, input_shape=(*args.input_size, 3))
```
fg-mindee, line 39: I didn't realize before but I think it's not possible to pass a "tuple" to a bash script as an argument. Either we pass only one side (and then use it twice in the input_shape), or we need to find another way

fg-mindee (commented):
I added a question about the bash script argument but the rest looks good!

fg-mindee (approved):
Thanks!

PR #180, opened 2021-03-30
fix: preprocessing (normalize) in training scripts

charlesmindee:
This PR adds normalization in both training scripts with a call to the preprocessor.
Any feedback is welcome!


fg-mindee (approved):
Thanks for the PR!

PR #183, opened 2021-03-30
feat: Added image support and model selection to demo app

fg-mindee:
- updated document reading
- added support of image as uploaded file
- added detection model selector

Here is how it renders:
image

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #187, opened 2021-03-31
refactor: Refactored recognition & backbone models

fg-mindee:
- moved vgg & resnet in a "backbones" submodule
- separated backbone test file
- removed vocab_size from recognition constructors since we pass the vocab itself
- moved vocab instantiation to core RecognitionModel constructor since it's common to all recognition models
- adapted unittests

Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR!

PR #188, opened 2021-03-31
feat: Added both localization & end-to-end visualization in demo app

fg-mindee:
- adds the possibility to turn page visualization to a static version (text plotted above boxes)
- adds task visualization in the demo app
- updated demo app screenshot in readme

Here is how it renders now:
doctr_demo_app

Any feedback is welcome!


charlesmindee (approved):
Thanks for this PR !

PR #189, opened 2021-03-31
feat: add reco loss in training script

charlesmindee:
This PR adds recognition loss in the training script.
The script is now operational to train crnn & sar model.
Any feedback is welcome!


fg-mindee (approved):
Thanks for the PR! Can you confirm that this is running without errors before merging?

charlesmindee:
Yes it is running!

PR #191, opened 2021-04-01
feat: add Linknet model

charlesmindee:
This PR adds Linknet model and corresponding unitests.
Any feedback is welcome!


doctr/models/utils.py
```diff
@@ -140,7 +142,9 @@ def conv_sequence(
     kwargs['use_bias'] = kwargs.get('use_bias', not(bn))
     # Add activation directly to the conv if there is no BN
     kwargs['activation'] = activation if not bn else None
-    conv_seq = [layers.Conv2D(out_channels, padding=padding, kernel_initializer=kernel_initializer, **kwargs)]
+    conv_seq = [
+        layers.Conv2D(out_channels, strides=strides, padding=padding, kernel_initializer=kernel_initializer, **kwargs)
+    ]
```
fg-mindee, line 147: Since the kwargs are passed to the Conv2D, we don't need to add it (simply passing `strides=...` when calling `conv_sequence`)

fg-mindee (changes requested):
Thanks for the PR! I left a few comments

fg-mindee:
Also, don't forget to add it to the documentation and add the paper reference in the README like the other models!

doctr/models/utils.py
```diff
@@ -115,6 +115,7 @@ def conv_sequence(
     out_channels: int,
     activation: str = None,
     bn: bool = False,
+    strides: int = 1,
```
fg-mindee, line 118: You can remove this

doctr/models/utils.py
```diff
@@ -128,6 +129,7 @@ def conv_sequence(
 
     Args:
         out_channels: number of output channels
+        strides: square strides to use in the Conv2D
```
fg-mindee, line 132: same here

fg-mindee (changes requested):
Minor things to correct and we're good to go!

fg-mindee (approved):
Looks good to me, thanks!

PR #192, opened 2021-04-01
fix: Fixed unnecessary os.listdir for each sample

fg-mindee:
- fixes a typo in variable naming (indexes -> indices)
- fixes `os.listdir` being computed for each sample unnecessarily

Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR !

PR #194, opened 2021-04-02
fix: typo script train recognition

charlesmindee:
Very small PR to fix a scaling issue in recognition training script.


references/recognition/train.py
```diff
@@ -52,7 +52,7 @@ def main(args):
     MEAN_RGB = (0.694, 0.695, 0.693)
     STD_RGB = (0.299, 0.296, 0.301)
     preprocessor = RecognitionPreProcessor(
-        output_size=(args.input_size, args.input_size),
+        output_size=(args.input_size, 4 * args.input_size),
```
fg-mindee, line 55: shouldn't this also be changed on the `DataGenerator` earlier in this script?

fg-mindee (commented):
Thanks for the fix! But I think there is another modification to perform

fg-mindee (approved):
Thanks for the edits!

PR #195, opened 2021-04-02
fix: ctc loss + args train scripts

charlesmindee:
This PR fixes ctc loss (removing softmax) and img_folder arg in both training scripts.
Any feedback is welcome!


fg-mindee (approved):
Thanks for the fix!

PR #196, opened 2021-04-06
feat: add a more robust version of the detection model

charlesmindee:
- The detection model is now pretrained on more than 15 epochs of our dataset, and it is far more robust
- The postprocessor performs opening (erosion + dilatation) on the binary map, which improves performances

Any feedback is welcome!


fg-mindee (approved):
Thanks for the PR!

Issue #190, opened 2021-04-01
[datasets] Optimize dataloaders for faster iterations


fg-mindee:
The current training scripts in `references` are way too slow.
Let's discuss here how we can improve them

fg-mindee:
Let's benchmark our current pipeline
```
train_set = FUNSD(train=True, download=True)

    start_ts = time()

    for _ in range(args.epochs):
        for page, target in tqdm(train_set):
            page = tf.image.resize(tf.cast(page, dtype=tf.float32), (1024, 1024), method='bilinear') / 255.

    print(time() - start_ts)
```
yields 25.6 sec

Using `ImageDataGenerator`

```
image_generator = ImageDataGenerator(rescale=1./255)
data_generator = image_generator.flow_from_directory(
        batch_size=4,
        directory=FOLDER,
        shuffle=True,
        target_size=(1024, 1024),
        interpolation='bilinear',
        class_mode='categorical')

tart_ts = time()

    for _ in range(args.epochs):
        ds_iter = iter(data_generator)
        for _ in tqdm(range(len(data_generator))):
            page, _ = next(ds_iter)


    print(time() - start_ts)
```
yields 112sec

Using `tf.data.Dataset`:
```
list_ds = tf.data.Dataset.list_files(folder)
list_ds = list_ds.shuffle(len(list_ds), reshuffle_each_iteration=False)

def process_img(file_path):
        # label = get_label(file_path)
        # load the raw data from the file as a string
        img = tf.io.read_file(file_path)
        img = tf.image.decode_jpeg(img, channels=3)
        return tf.image.resize(img, [1024, 1024], method='bilinear')


AUTOTUNE = tf.data.AUTOTUNE
labeled_ds = list_ds.map(process_img, num_parallel_calls=AUTOTUNE)
start_ts = time()

for _ in range(args.epochs):
        for img in labeled_ds.take(1):
            pass

print(time() - start_ts)
```
yields 1.36s

Using `tf.data.Dataset` while batching:
```
ds = tf.data.Dataset.list_files(folder)
# ds = ds.shuffle(len(ds), reshuffle_each_iteration=False)

def process_img(file_path):
    # label = get_label(file_path)
    # load the raw data from the file as a string
    img = tf.io.read_file(file_path)
    img = tf.image.decode_jpeg(img, channels=3)
    return tf.image.resize(img, [1024, 1024], method='bilinear')


AUTOTUNE = tf.data.AUTOTUNE
[...]

PR #198, opened 2021-04-07
feat: Added DataLoader object on top of datasets

fg-mindee:
Following up on #190, this PR introduces the following modifications:
- refactored detection & recognition loaders into dataset objects (only getitem and how they are expected to be batched)
- implemented an abstract dataset wrapper performing the dataloading (parallel fetching + batching)
- added unittests
- added `DataLoader` to the documentation
- updated training reference scripts

Any feedback is welcome!


charlesmindee (approved):
Thanks for this very clean PR, did you check that this way of loading data is faster than the fastest option you benchmarked last week ?

charlesmindee (approved):
Thanks!

PR #199, opened 2021-04-08
feat: add GVision & AWS textract perf. in doc

charlesmindee:
This PR reports in the doc the results of Gvision (text detection & document text detection) and AWS textract on CORD and FUNSD datasets.
Any feedback is welcome!


fg-mindee (approved):
Thanks a lot!

PR #203, opened 2021-04-09
Refacto: Postprocessors

charlesmindee:
This PR is a refactoring of postprocessors:
- all postprocessors are instantiated inside the models to avoid mismatchings
- CTC postprocessor has been moved to CRNN file
- both detection postprocessors share  boxe_score and call methods, which are now implemented in core class.

Any feedback is welcome!


fg-mindee (approved):
Thanks for the refactoring!

PR #209, opened 2021-04-20
feat: add FPS in docs

charlesmindee:
This PR adds in the docs:
- FPS for each model, averaged on 1000 random tensors
- FPS for each end-to-end predictor, averaged on each dataset
Default batch size is now 128 for the recognition task in the predictor.
Any feedback is welcome!


fg-mindee (approved):
Thanks!

fg-mindee (approved):
Thanks for the details!

PR #210, opened 2021-04-21
refactor: Refactored the recognition training script

fg-mindee:
- added a conditional checkpoint saving policy
- added fastprogress console bar
- updated detection model empty GT case

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #211, opened 2021-04-21
fix: Fixed image scaling and added transformations to training

fg-mindee:
- added back the division by 255 after reading the image
- added transformations to detection & recognition training

Any feedback is welcome!

### DIFF
--- a/doctr/transforms/modules.py
+++ b/doctr/transforms/modules.py
@@ -12,7 +12,7 @@ from . import functional as F
 
 
 __all__ = ['Compose', 'Resize', 'Normalize', 'LambdaTransformation', 'ToGray', 'InvertColorize',
-           'RandomBrightness', 'RandomContrast', 'RandomSaturation', 'RandomHue', 'Gamma', 'RandomJpegQuality',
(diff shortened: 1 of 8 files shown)


doctr/transforms/modules.py
```diff
@@ -8,10 +8,12 @@
 from typing import List, Any, Tuple, Callable
 
 from doctr.utils.repr import NestedObject
+from . import functional as F
 
 
 __all__ = ['Compose', 'Resize', 'Normalize', 'LambdaTransformation', 'ToGray', 'InvertColorize',
-           'Brightness', 'Contrast', 'Saturation', 'Hue', 'Gamma', 'JpegQuality', 'OneOf', 'RandomApply']
+           'RandomBrightness', 'RandomContrast', 'RandomSaturation', 'RandomHue', 'Gamma', 'RandomJpegQuality',
```
charlesmindee, line 15: Why do we leave Gamma instead of RandomGamma here ?

charlesmindee (commented):
Thanks, only a small point to check on the naming of Gamma

charlesmindee (approved):
Thanks!

PR #212, opened 2021-04-22
fix: Fixed inplace operation in DetectionDataset

fg-mindee:
- adds a short name alternative for the batch size arg in the reference scripts
- fixed an inplace box operation in `DetectionDataset`

Any feedback is welcome!


charlesmindee (approved):
Thanks for the fix

PR #214, opened 2021-04-26
fix: train recognition script, /255 validation dataset

charlesmindee:
This PR adds a /255 normalization (which was missing) when loading the validation recognition dataset in the training script.


fg-mindee (approved):
Good catch! Thanks

PR #222, opened 2021-04-28
feat: Added annotation reading for PDF documents

### LINKED ISSUE #144: [documents] Check for input PDF with PyMuPDF when it's source content
So far, we have used the content reading feature of PyMuPDF. When a source PDF is read, the library actually extracts all the localization and text information from the document. Two options are then available:
- skipping the model inference and use this
- combining it with the model inference to improve our predictions

fg-mindee:
- changed `read_pdf` to return a `fitz.Document`
- added a document template class for PDF to retrieve either the images or the words/artefacts
- updated documentation
- updated instructions in README
- updated unittests

Resolves #144

Any feedback is welcome!


charlesmindee (approved):
Thanks !

charlesmindee (approved):

PR #224, opened 2021-04-28
feat: add blocks detection

charlesmindee:
This PR implements resolve_blocks inside DocumentBuilder:
Each document can be built using resolve_block=True (only works if resolve_line=True and both are set to False by default to lighten postprocessing), which will determine the blocks of the pages with a hierarchical clustering on the lines.
Any feedback is welcome!

Computation time: highly insignificant, less than 0.01% of the prediction time for a light page and 0.1% for a dense page on GPU

Examples of detected blocks:
invoice
receipt
visit

### DIFF
--- a/doctr/models/core.py
+++ b/doctr/models/core.py
@@ -5,12 +5,14 @@
 
 
 import numpy as np
+from scipy.cluster.hierarchy import fclusterdata
 from typing import List, Any, Tuple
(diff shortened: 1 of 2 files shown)


fg-mindee (approved):
Quite a nice feature, thanks! Would you mind specifying in the description if this adds a significant amount of computation duration please?

PR #226, opened 2021-04-29
fix: dtype target["boxes"] for CORD and FUNSD

charlesmindee:
This PR fixes dtype (float --> int) for target["boxes"] of CORD and FUNSD, to avoid confusion between relative and absolute boxes.


fg-mindee (approved):
Thanks!

PR #227, opened 2021-04-30
fix: Fixed localization metric update

fg-mindee:
- fixes edge case of metric update when there are no predictions (previously it would reset IoU total to 0)
- updates unittests

Any feedback is welcome!


charlesmindee (approved):
Thanks

PR #228, opened 2021-04-30
feat: Improved training script

fg-mindee:
- adds console print of dataset information
- added option to run only the evaluation loop
- refactored training and evaluation loops

Any feedback is welcome!


charlesmindee (approved):
thanks!

PR #229, opened 2021-04-30
feat: Improved error when read_pdf gets an incorrect path

fg-mindee:
- added appropriate error when read_pdf gets a path to inexisting file
- updated unittests

Any feedback is welcome!


charlesmindee (approved):
thanks!

PR #230, opened 2021-04-30
fix: adjust detection postprocessing parameters to img size

charlesmindee:
This PR adjusts detection postprocessing to img size: If input is smaller than 1000 x 1000, don't perform opening (erosion + dilatation of boxes) because boxes can be too small. The others parameters don't depend on the size of the input (we could reduce min_size_box but this doesn't make sense because a box smaller than 2 pixels would be impossible to read, no matter the size of the input image).
Any feedback is welcome!

### DIFF
--- a/doctr/models/detection/core.py
+++ b/doctr/models/detection/core.py
@@ -85,12 +85,10 @@ class DetectionPostProcessor(NestedObject):
 
     def __init__(
         self,
-        min_size_box: int = 5,
         box_thresh: float = 0.5,
         bin_thresh: float = 0.5
     ) -> None:
 
-        self.min_size_box = min_size_box
         self.box_thresh = box_thresh
         self.bin_thresh = bin_thresh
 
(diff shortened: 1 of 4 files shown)


doctr/models/detection/core.py
```diff
@@ -145,13 +145,15 @@ def __call__(
...
+        # kernel for opening
+        kernel = np.ones((3, 3), np.uint8)
 
         for p_, bitmap_ in zip(p, bitmap):
             p_ = p_.numpy()
             bitmap_ = bitmap_.numpy()
-            # perform opening (erosion + dilatation)
-            kernel = np.ones((3, 3), np.uint8)
-            bitmap_ = cv2.morphologyEx(bitmap_, cv2.MORPH_OPEN, kernel)
+            if p_.shape[0] > 1000:
```
fg-mindee, line 154: We need to tune this hyperparameter, at 900, we don't need this step?

fg-mindee (commented):
Thanks for the PR! I added some questions

charlesmindee:
After a thorough investigation, this PR has evolved:
- The weak values for the recall in 512 x 512 are explained by a too low bin_threshold (0.15), the small number of pixels between words at that resolution made it impossible for the model to properly separate words. The bin_threshold is now set to 0.3, the baseline value as described in the DB paper.
- An empirical law has been established to find the "critical pixel length" at each scale, which is used for both kernel size for opening and min_size_box: ksize = min_size_box = 1 + int(height/512), which gives us a critical lentgh of 2 for 512, 3 for 1024, 5 for 2048. This yields to better results, as illustrated below.

Binary map for a 512 x 512 image:
512bin3

Opening performed with a kernel of 3x3: to large 
512opeining3
Opening performed with the right critical length: a kernel of 2x2
512opening2

For 2048:
Binary map for the same image
2048bin5
Opening 5x5 on this map:

doctr/models/detection/core.py
```diff
@@ -145,12 +143,14 @@ def __call__(
         bitmap = tf.unstack(bitmap, axis=0)
 
         boxes_batch = []
+        # Kernel for opening, empirical law for ksize
+        k_size = 1 + int(p[0].shape[0] / 512)
```
fg-mindee, line 147: I think we only need to investigate whether this depends on the input size the model was trained on

doctr/models/detection/linknet.py
```diff
@@ -70,6 +69,7 @@ def bitmap_to_boxes(
         """
         label_num, labelimage = cv2.connectedComponents(bitmap.astype(np.uint8), connectivity=4)
         height, width = bitmap.shape[:2]
+        min_size_box = 1 + int(height / 512)
```
fg-mindee, line 72: same here

fg-mindee (commented):
Thanks a lot for the PR! Just a little investigation to conduct and we're good for merge!

fg-mindee (approved):
Thanks for the investigation!

PR #231, opened 2021-05-03
feat: add freezing backbone arg in training script

charlesmindee:
This PR adds a freeze_bckb arg in the detection training script to freeze resnet50 when training db net.
Any feedback is welcome!

### DIFF
--- a/doctr/models/detection/differentiable_binarization.py
+++ b/doctr/models/detection/differentiable_binarization.py
@@ -60,7 +60,7 @@ class DBPostProcessor(DetectionPostProcessor):
         super().__init__(
             min_size_box,
             box_thresh,
-            bin_thresh,
+            bin_thresh
         )
         self.unclip_ratio = unclip_ratio
         self.max_candidates = max_candidates
@@ -114,7 +114,6 @@ class DBPostProcessor(DetectionPostProcessor):
                 containing x, y, w, h, score for the box
         """
         height, width = bitmap.shape[:2]
-        min_size_box = 1 + int(height / 512)
         boxes = []
         # get contours from connected components on the bitmap
         contours, _ = cv2.findContours(bitmap.astype(np.uint8), cv2.RETR_EXTERNAL, cv2.CHAIN_APPROX_SIMPLE)
--- a/doctr/models/detection/differentiable_binarization.py
+++ b/doctr/models/detection/differentiable_binarization.py
@@ -527,7 +527,6 @@ class DBNet(DetectionModel, NestedObject):
 def _db_resnet(
     arch: str,
     pretrained: bool,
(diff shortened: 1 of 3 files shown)


doctr/models/detection/differentiable_binarization.py
```diff
@@ -524,7 +524,13 @@ def call(
         return dict(proba_map=proba_map)
```
fg-mindee, line 524: no need to make modifications to differentiable_binarization!

references/detection/train.py
```diff
@@ -126,7 +126,12 @@ def main(args):
     optimizer = tf.keras.optimizers.Adam(learning_rate=args.lr, clipnorm=5)
 
     # Load doctr model
-    model = detection.__dict__[args.model](pretrained=False, input_shape=(args.input_size, args.input_size, 3))
+    model = detection.__dict__[args.model](
+        pretrained=False,
+        freeze_bckb=args.freeze_bckb,
+        input_shape=(args.input_size, args.input_size, 3)
+    )
+
```
fg-mindee, line 134: you can roll that back, we can modify this afterwards

references/detection/train.py
```diff
@@ -193,6 +198,7 @@ def parse_args():
     parser.add_argument('-j', '--workers', type=int, default=4, help='number of workers used for dataloading')
     parser.add_argument('--resume', type=str, default=None, help='Path to your checkpoint')
     parser.add_argument("--test-only", dest='test_only', action='store_true', help="Run the validation loop")
+    parser.add_argument('--freeze_bckb', type=bool, default=False, help='freeze model backbone for fine-tuning')
```
fg-mindee, line 201:
```
parser.add_argument('--freeze-backbone', dest='freeze_backbone', action='store_true', help='freeze model backbone for fine-tuning')
```

references/detection/train.py
fg-mindee: Right before the training loop & loss queue, add this:
```
if args.freeze_backbone:
    for layer in model.feat_extractor.layers:
        layer.trainable = False
```

fg-mindee (changes requested):
Thanks for the PR! I suggested a lighter change that would yield the same result without changing the DB implementation :ok_hand:

test/test_models_detection.py
```diff
@@ -116,6 +116,7 @@ def test_detection_models(arch_name, input_shape, output_size, out_prob):
         assert np.all(np.logical_and(out >= 0, out <= 1))
 
 
+
```
fg-mindee: flake is going to say there are too many blank lines :sweat_smile:

fg-mindee (approved):
Thanks!

PR #237, opened 2021-05-04
feat: set default image resolution to 1024 in read_pdf

### LINKED ISSUE #232: [document] Check integrity of PDF --> img conversion with default DPI
The current PDF reading implies a conversion to an image. As we are using the default args for this conversion, the DPI value might be low. We need to check that the default parameter is not bringing about performance issues due to low resolution rendering.

charlesmindee:
Before this PR, the image resolution was computed by default this way in `read_pdf`:
Find pdf dimension, scale by 72 DPI to obtain image dimension.

This raises 2 issues:
- Different document sizes yield to different resolutions (constant DPI)
- For A4 sizes (a huge amount of docs), this method was re-scaling to 840 x 595 pixels which is way to low for very dense docs.

Now we have a constant resolution of 1024 x 1024 by default.
Any feedback is welcome!

closes #232

### DIFF
--- a/demo/app.py
+++ b/demo/app.py
@@ -48,7 +48,7 @@ def main():
     uploaded_file = st.sidebar.file_uploader("Upload files", type=['pdf', 'png', 'jpeg', 'jpg'])
     if uploaded_file is not None:
         if uploaded_file.name.endswith('.pdf'):
-            doc = DocumentFile.from_pdf(uploaded_file.read()).as_images()
+            doc = DocumentFile.from_pdf(uploaded_file.read()).as_images(output_size=(1024, 1024))
         else:
             doc = DocumentFile.from_images(uploaded_file.read())
         cols[0].image(doc[0], "First page", use_column_width=True)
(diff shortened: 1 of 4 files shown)


fg-mindee (changes requested):
Thanks for the PR! While the diagnosis seems correct to me, I disagree with how it's handled: making 1024x1024 the default output of read_pdf is only helpful if all subsequent models need this input size (which may change in the future). So my suggestion would be: modify the eval / train / demo script & usage instructions to change read_pdf output_size when needed. For the rest, the inputs are already tensors so this edge case won't be relevant

Besides, resizing with read_pdf, is only beneficial for detection. For recognition, when extracting the crops, I guess we prefer to have non-distorted origin image

charlesmindee:
I agree with your first comment, but not the second one: we need to have good quality images for recognition task, and default sizing for A4 pdf is 840 x 590 which is low to distinguish letters on small crops. If we want to preserve aspect ratio we can compute another sizing resolution but I think we can't leave this to 840 x 590.

charlesmindee:
Yes, a zoom x2 would be nice to apply for all A4 docs

fg-mindee:
Alright, let's reopen the issue until we make a PR for that then :ok_hand:

PR #238, opened 2021-05-04
docs: add vocabs in doc

### LINKED ISSUE #235: [docs] Better specify which vocab is supported by which model
Without reading the code, it is not possible currently to know the specific alphabet used, and which model is using which. We need to tackle this

charlesmindee:
This PR adds:
- tab of supported vocabs in doc (dataset)
- vocab to use for each model (models)

Any feedback is welcome!

closes #235

### DIFF
--- a/docs/source/models.rst
+++ b/docs/source/models.rst
@@ -98,6 +98,8 @@ Identifying strings in images
 All text recognition models above have been evaluated using both the training and evaluation sets of FUNSD and CORD (cf. :ref:`datasets`).
 Explanations about the metrics being used are available in :ref:`metrics`.
 
+All these recognition models are trained with our french vocab (cf. :ref:`vocabs`).
+
 *Disclaimer: both FUNSD subsets combine have 30595 word-level crops which might not be representative enough of the model capabilities*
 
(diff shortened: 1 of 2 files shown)


docs/source/models.rst
```diff
@@ -98,6 +98,8 @@ Identifying strings in images
 All text recognition models above have been evaluated using both the training and evaluation sets of FUNSD and CORD (cf. :ref:`datasets`).
 Explanations about the metrics being used are available in :ref:`metrics`.
 
+All these recognition models are trained with our french vocab (cf. :ref:`vocabs`).
```
fg-mindee, line 101: If I'm not mistaking, for now, we're using a custom alphabet (not in the same order as the french's one?)

fg-mindee (approved):
Thanks for the PR! I added a small comment as I believe our pretrained recognition models are still using the former alphabet for now

docs/source/models.rst
charlesmindee: yes, do I need to change the order in the doc ?

charlesmindee: yes, do I need to change the order in the docs ?

fg-mindee: No that's alright, we're gonna change the model soon anyway!

PR #240, opened 2021-05-05
feat: Increased default DPI for page rendering of PDFs

### LINKED ISSUE #232: [document] Check integrity of PDF --> img conversion with default DPI

fg-mindee:
Following up on #237, this PR introduces the following modifications:
- doubled the default DPI for page rendering
- fixed import of PDF class
- added specific unittest

Resolves #232

Any feedback is welcome!

### DIFF
--- a/doctr/documents/reader.py
+++ b/doctr/documents/reader.py
@@ -105,12 +105,13 @@ def convert_page_to_numpy(
         the rendered image in numpy format
     """
 
-    transform_matrix = None
-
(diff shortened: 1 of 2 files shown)


charlesmindee (approved):
Thanks!

PR #241, opened 2021-05-05
fix: Fixed tensor scaling in recognition script

fg-mindee:
This PR fixes a double tensor scaling happening in the recognition script


charlesmindee (approved):
Thanks for the fix

Issue #233, opened 2021-05-04
[api] Add a minimal API setup with DocTR


fg-mindee:
Since the library can be used to build an API, we could add a minimal codebase for users to deploy a light API for OCR on documents.

Here is a more detailed suggestion:
- put everything in a `api` folder
- implement it with FastAPI
- implement a POST method "/analyze" that returns the OCR results

spread into 3 PRs:
- [x] Text recognition route (#242)
- [x] Text detection route (#245)
- [x] OCR route (#247)

PR #242, opened 2021-05-05
feat: Added minimal API template for text recognition

fg-mindee:
Following up on #233, this PR introduces the following modifications:
- adds a minimal FastAPI-based implementation of a text recognition route
- added Docker orchestration
- added API unittests
- added CI job to ensure it runs correctly & that corresponding unittests are run correctly

Here is how the API documentation renders:
doctr_api_doc

I made a test on my own with the following image:
invite
Once the API is running locally on port 8050, with this piece of code:
```
import requests
import io
with open('/home/fg/Downloads/invite.jpeg', 'rb') as f:
    data = f.read()
response = requests.post("http://localhost:8050/recognition", files={'file': io.BytesIO(data)})
```
I get the following result
```
In [10]: response.json()
Out[10]: {'value': 'invite'}
```

Any feedback is welcome!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -177,3 +177,17 @@ jobs:
         run: |
           wget https://github.com/mindee/doctr/releases/download/v0.1.0/sample.pdf
           python scripts/analyze.py sample.pdf --noblock
+
+  api-ready:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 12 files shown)


charlesmindee (approved):
Thanks for this feature!

PR #244, opened 2021-05-06
feat: add testset in datasets

charlesmindee:
This PR adds the following:
-TESTSET in datasets to load and evaluate confidential test sets
- evolution of the evaluate script to make it more flexible
- removed unused imports
- removed SROIE deprecated files

Any feedback is welcome!


scripts/evaluate.py
```diff
@@ -78,6 +86,8 @@ def parse_args():
     parser.add_argument('detection', type=str, help='Text detection model to use for analysis')
     parser.add_argument('recognition', type=str, help='Text recognition model to use for analysis')
     parser.add_argument('--iou', type=float, default=0.5, help='IoU threshold to match a pair of boxes')
+    parser.add_argument('-dataset', type=str, default='FUNSD', help='choose a dataset: FUNSD, CORD')
+    parser.add_argument('-path', type=str, default=None, help='Only for local sets, path to your OCRDataset')
```
fg-mindee, line 90: If those are optional args, we need two hyphens (one hyphen only if that's the shortcut arg with a single character)

fg-mindee (commented):
Thanks for the PR! Do you think we could add some unittest on a tiny subset?

doctr/datasets/ocrdataset.py
```diff
@@ -0,0 +1,71 @@
+# Copyright (C) 2021, Mindee.
```
fg-mindee, line 1: let's call the module `ocr.py` for consistency

doctr/datasets/ocrdataset.py
```diff
@@ -0,0 +1,71 @@
...
+        **kwargs: Any,
+    ) -> None:
+
+        self.sample_transforms = (lambda x: x) if sample_transforms is None else sample_transforms
+        self.root = path
+
+        # List images
+        self.data: List[Tuple[str, Dict[str, Any]]] = []
+        with open(os.path.join(self.root, 'labels/typed_word/labels.json'), 'rb') as f:
+            data = json.load(f)
```
fg-mindee, line 42: I would argue that the img_folder and the labels path could be completely different, perhaps both should be different args in the constructor?

scripts/evaluate.py
```diff
@@ -61,7 +68,7 @@ def main(args):
             e2e_metric.update(gt_boxes, np.asarray(pred_boxes), gt_labels, pred_labels)
 
     # Unpack aggregated metrics
-    print(f"Model Evaluation (model= {args.detection} + {args.recognition}, dataset='FUNSD')")
+    print(f"Model Evaluation (model= {args.detection} + {args.recognition}, dataset={args.dataset})")
```
fg-mindee, line 71: `{'OCRDataset' if args.path else args.dataset}`

fg-mindee (commented):
Thanks for the PR! I think for debug, we should have two args: 1 for img_folder, 1 for label_file

test/test_datasets_ocrdataset.py
```diff
@@ -0,0 +1,53 @@
...
+            "coordinates": [[[0.1, 0.2], [0.2, 0.2], [0.2, 0.3], [0.1, 0.3]]],
+            "string": ['!']
+        }
+    ]
+    with open(label_file, 'w') as f:
+        json.dump(label, f)
+
+    file = BytesIO(mock_image_stream)
+    for i in range(5):
+        fn = root.join("images", f"mock_image_file_{i}.jpg")
```
fg-mindee, line 44: You need to create the `images` subfolder or you won't be able to write the file

scripts/evaluate.py
```diff
@@ -78,6 +86,8 @@ def parse_args():
     parser.add_argument('detection', type=str, help='Text detection model to use for analysis')
     parser.add_argument('recognition', type=str, help='Text recognition model to use for analysis')
     parser.add_argument('--iou', type=float, default=0.5, help='IoU threshold to match a pair of boxes')
+    parser.add_argument('--dataset', type=str, default='FUNSD', help='choose a dataset: FUNSD, CORD')
+    parser.add_argument('--paths', type=str, default=None, help='Only for local sets, (img_folder, label_file)')
```
fg-mindee, line 90: Either we split this into 2 args or we keep one but append what we need to each constructor args

scripts/evaluate.py
```diff
@@ -16,30 +16,34 @@
...
 
 
 def main(args):
 
     model = ocr_predictor(args.detection, args.recognition, pretrained=True)
 
-    train_set = FUNSD(train=True, download=True)
-    test_set = FUNSD(train=False, download=True)
+    if args.path:
+        testset = datasets.OCRDataset(*args.paths)
```
fg-mindee, line 28: if the two args are independent, they need to be inputted as such I think

fg-mindee (commented):
Sorry, I still have a few comments

fg-mindee (approved):
Thanks for the edits!

PR #245, opened 2021-05-06
feat: Added text detection route to API template

fg-mindee:
As per #233, this PR introduces the following modifications:
- added text detection route to the API template
- added corresponding API unittests

For the following image:
detection_sample
with this piece of code:
```
import requests
import io
with open('/home/fg/Downloads/detection_sample.jpg', 'rb') as f:
    data = f.read()
response = requests.post("http://localhost:8050/detection", files={'file': io.BytesIO(data)})
```
which yields
```
In [4]: response.json()
Out[4]: 
[{'box': [0.826171875, 0.185546875, 0.90234375, 0.201171875]},
 {'box': [0.75390625, 0.185546875, 0.8173828125, 0.201171875]}]
```
(IoU > 85% with the GT)

Any feedback is welcome!

### DIFF
--- a/api/app/main.py
+++ b/api/app/main.py
@@ -8,7 +8,7 @@ from fastapi import FastAPI, Request
 from fastapi.openapi.utils import get_openapi
 
 from app import config as cfg
-from app.routes import recognition
+from app.routes import recognition, detection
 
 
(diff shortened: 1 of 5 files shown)


charlesmindee (approved):
Thanks for this feature!

PR #248, opened 2021-05-07
refacto: postprocessor instantiation

charlesmindee:
This PR fixes a double instantiation of postprocessors in predictors.

Any feedback is welcome!

### DIFF
--- a/doctr/models/detection/zoo.py
+++ b/doctr/models/detection/zoo.py
@@ -10,19 +10,16 @@ from .. import detection
 
 __all__ = ["detection_predictor"]
 
-default_cfgs: Dict[str, Dict[str, Any]] = {
-    'db_resnet50': {'model': 'db_resnet50'},
-    'linknet': {'model': 'linknet'},
-}
+archs = ['db_resnet50', 'linknet']
 
 
 def _predictor(arch: str, pretrained: bool, **kwargs: Any) -> DetectionPredictor:
 
-    if default_cfgs.get(arch) is None:
+    if arch not in archs:
         raise ValueError(f"unknown architecture '{arch}'")
 
     # Detection
-    _model = detection.__dict__[default_cfgs[arch]['model']](pretrained=pretrained)
+    _model = detection.__dict__[arch](pretrained=pretrained)
     kwargs['mean'] = kwargs.get('mean', _model.cfg['mean'])
     kwargs['std'] = kwargs.get('std', _model.cfg['std'])
     predictor = DetectionPredictor(
--- a/doctr/models/detection/zoo.py
+++ b/doctr/models/detection/zoo.py
@@ -10,12 +10,12 @@ from .. import detection
 
 __all__ = ["detection_predictor"]
 
-archs = ['db_resnet50', 'linknet']
+ARCHS = ['db_resnet50', 'linknet']
 
 
 def _predictor(arch: str, pretrained: bool, **kwargs: Any) -> DetectionPredictor:
 
-    if arch not in archs:
+    if arch not in ARCHS:
         raise ValueError(f"unknown architecture '{arch}'")
 
     # Detection
@@ -24,8 +24,7 @@ def _predictor(arch: str, pretrained: bool, **kwargs: Any) -> DetectionPredictor
     kwargs['std'] = kwargs.get('std', _model.cfg['std'])
     predictor = DetectionPredictor(
         DetectionPreProcessor(output_size=_model.cfg['input_shape'][:2], **kwargs),
-        _model,
-        _model.postprocessor
+        _model
     )
     return predictor
 
--- a/doctr/models/detection/zoo.py
+++ b/doctr/models/detection/zoo.py
@@ -11,8 +11,8 @@ from .. import detection
 __all__ = ["detection_predictor"]
 
 default_cfgs: Dict[str, Dict[str, Any]] = {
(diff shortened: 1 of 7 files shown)


doctr/models/detection/zoo.py
```diff
@@ -11,8 +11,8 @@
 __all__ = ["detection_predictor"]
 
 default_cfgs: Dict[str, Dict[str, Any]] = {
-    'db_resnet50': {'model': 'db_resnet50', 'post_processor': 'DBPostProcessor'},
-    'linknet': {'model': 'linknet', 'post_processor': 'LinkNetPostProcessor'},
+    'db_resnet50': {'model': 'db_resnet50'},
+    'linknet': {'model': 'linknet'},
 }
```
fg-mindee, line 16: We can remove this now

fg-mindee: and replace L25 by `_model = detection.__dict__[arch](pretrained=pretrained)`

doctr/models/recognition/zoo.py
```diff
@@ -11,10 +11,10 @@
...
 default_cfgs: Dict[str, Dict[str, Any]] = {
-    'crnn_vgg16_bn': {'model': 'crnn_vgg16_bn', 'post_processor': 'CTCPostProcessor'},
-    'crnn_resnet31': {'model': 'crnn_resnet31', 'post_processor': 'CTCPostProcessor'},
-    'sar_vgg16_bn': {'model': 'sar_vgg16_bn', 'post_processor': 'SARPostProcessor'},
-    'sar_resnet31': {'model': 'sar_resnet31', 'post_processor': 'SARPostProcessor'},
+    'crnn_vgg16_bn': {'model': 'crnn_vgg16_bn'},
+    'crnn_resnet31': {'model': 'crnn_resnet31'},
+    'sar_vgg16_bn': {'model': 'sar_vgg16_bn'},
+    'sar_resnet31': {'model': 'sar_resnet31'},
 }
```
fg-mindee, line 18: can be removed

fg-mindee: L26 can be updated as well

fg-mindee (changes requested):
Thanks for the PR! A few things to change and we're good to go!

doctr/models/detection/zoo.py
fg-mindee: mmmh, you mean that it could be a subset of what's available as detection architectures?

doctr/models/detection/zoo.py
```diff
@@ -10,25 +10,22 @@
 
 __all__ = ["detection_predictor"]
 
-default_cfgs: Dict[str, Dict[str, Any]] = {
-    'db_resnet50': {'model': 'db_resnet50', 'post_processor': 'DBPostProcessor'},
-    'linknet': {'model': 'linknet', 'post_processor': 'LinkNetPostProcessor'},
-}
+archs = ['db_resnet50', 'linknet']
```
fg-mindee, line 13: if we do this, `archs` is a session constant so it needs to be named in uppercase

doctr/models/recognition/zoo.py
```diff
@@ -10,26 +10,21 @@
 
 __all__ = ["recognition_predictor"]
 
-default_cfgs: Dict[str, Dict[str, Any]] = {
-    'crnn_vgg16_bn': {'model': 'crnn_vgg16_bn', 'post_processor': 'CTCPostProcessor'},
-    'crnn_resnet31': {'model': 'crnn_resnet31', 'post_processor': 'CTCPostProcessor'},
-    'sar_vgg16_bn': {'model': 'sar_vgg16_bn', 'post_processor': 'SARPostProcessor'},
-    'sar_resnet31': {'model': 'sar_resnet31', 'post_processor': 'SARPostProcessor'},
-}
+archs = ['crnn_vgg16_bn', 'crnn_resnet31', 'sar_vgg16_bn', 'sar_resnet31']
```
fg-mindee, line 13: `ARCHS = ...`

doctr/models/detection/zoo.py
```diff
@@ -10,25 +10,22 @@
...
     # Detection
-    _model = detection.__dict__[default_cfgs[arch]['model']](pretrained=pretrained)
+    _model = detection.__dict__[arch](pretrained=pretrained)
     kwargs['mean'] = kwargs.get('mean', _model.cfg['mean'])
     kwargs['std'] = kwargs.get('std', _model.cfg['std'])
     predictor = DetectionPredictor(
         DetectionPreProcessor(output_size=_model.cfg['input_shape'][:2], **kwargs),
         _model,
-        detection.__dict__[default_cfgs[arch]['post_processor']]()
+        _model.postprocessor
```
fg-mindee, line 28: even though, it won't be a double instantiation because we pass it by reference here, if you print the predictor, you will see 2 postprocessors. I think we have 2 options here:
1. not passing the postprocessor to the predictor and in the predictor's call, it will use `self.model.postprocessor`
2. in model, only the class (not an instance) is set to `self.postprocessor` so we pass it in the predictor but using the preprocessor's constructor

The easiest would probably be the first option

fg-mindee (commented):
Sorry, a few comments still pending

doctr/models/detection/zoo.py
charlesmindee: yes

fg-mindee (approved):
Thanks for the edits!

PR #249, opened 2021-05-10
feat: Added sample display option in training script

fg-mindee:
- added option to plot samples for training sanity check
- refactored transformations
- updated probability of color inversion in training scripts

Here is how it renders for text detection:
det_samples

and for text recognition:
reco_samples

Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR !

PR #250, opened 2021-05-10
refactor: Refactored model calls and fixed validation loop

fg-mindee:
- fully moved post_processor inside model definition (the predictor has no knowledge of it)
- removed final activation layers from detection models for loss computation
- model output is now a dictionary for all models : if a target is passed, the loss will be computed and returned in the dict. If it isn't, boxes are returned for detection models and preds for recognition models. Additional call options (return_model_outputs, return_preds, return_boxes) can enforce adding extra information in the returned dict. This way, the training mode is completely decorrelated to the model output.
```python
det_model = db_resnet50()
print(det_model(input_t).keys())
print(det_model(input_t, return_model_output=True).keys())
print(det_model(input_t, target).keys())
print(det_model(input_t, target, return_boxes=True).keys())
print(det_model(input_t, target, return_boxes=True, return_model_output=True).keys())
```
yields
```
dict_keys(['boxes'])
dict_keys(['boxes', 'out_map'])
dict_keys(['loss'])
dict_keys(['loss', 'boxes'])
dict_keys(['loss', 'boxes', 'out_map'])
```
_`out_map` is mainly for debug and plotting heatmaps_ 
- updated unittests accordingly
- updated training scripts accordingly

Any feedback is welcome!

### DIFF
--- a/doctr/models/detection/core.py
+++ b/doctr/models/detection/core.py
@@ -125,7 +125,7 @@ class DetectionPostProcessor(NestedObject):
 
     def __call__(
         self,
-        x: Dict[str, tf.Tensor],
+        proba_map: tf.Tensor,
     ) -> List[np.ndarray]:
         """Performs postprocessing for a list of model outputs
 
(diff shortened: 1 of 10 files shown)


charlesmindee (approved):
Thanks!

PR #259, opened 2021-05-14
feat: add QRCodeDetector in models.artefacts

charlesmindee:
This PR adds a QRCodeDetector in artefacts, based on cv2 QRCodeDetector, which detects and decode QR codes on images
Samples are displayed below.
Any feedback is welcome!

qr1
qr3
qr4
qr4
qr5
qr6


charlesmindee:
It takes 10s to run this detector 100 times on the same image, a 1024 x 1024 picture with a QR code on it. Experiments were conducted on a CPU (intel core i7 9th gen).

doctr/models/artefacts/qrcode.py
```diff
@@ -0,0 +1,41 @@
...
+class QRCodeDetector:
+
+    """ Implements a QR-code detector.
+    Based on open CV QRCodeDetector. Returns both localization and decoded text.
+    """
+
+    def __init__(
+        self,
+    ) -> None:
+        self.detector = cv2.QRCodeDetector()
```
fg-mindee, line 22: just need to inherit from cv2.QRCodeDetector, so that you don't have to implement the constructor ;)

fg-mindee (commented):
Thanks for the PR! Small improvement suggestion in the comments

fg-mindee (commented):
there seems to be a segmentation fault in the unittest

fg-mindee:
Small hypothesis here: cv2 is using a DL model under the hood or something similarly heavy on computation resources :thinking: 
(Since OpenCV 4, I think there has been a lot of those high computation features)

charlesmindee:
Since we can't fix the issue with the test, I am closing this for now

PR #266, opened 2021-05-18
fix: Fixed Resize transformation when preserving aspect ratio

fg-mindee:
- switched Resize transforms to preserving aspect ratio in recognition training script
- moved `ColorInversion` before Resize to avoid getting a padding with value != 0
- fixed padding offset when preserving aspect ratio
- extended unittests

The following snippet:
```shell
python references/recognition/train.py /home/fg/Downloads/reco_samples sar_vgg16_bn -b 16 --show-samples
```
used to yield:
resize_legacy

by switching to aspect ratio preserving Resize:
resize_before

and now yields (by fixing the resize):
resize_after

Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR!

PR #267, opened 2021-05-18
fix: Fixed OCRMetric update when there is no predictions

fg-mindee:
Following up on #227, this PR introduces the following modifications:
- fixes edge case of metric update when there are no predictions (previously it would reset IoU total to 0)
- only performs IoU computation & assignment in appropriate cases in the update method
- updates unittests

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #269, opened 2021-05-19
docs: add bench private datasets in docs

charlesmindee:
This PR adds a benchmark in the doc of our baseline model compared to Gvision and AWS Textract on 3 private datasets.
Any feedback is welcome!


fg-mindee (approved):
Thanks for the new benchmark!

PR #270, opened 2021-05-19
feat: handle blank img in ocr datasets

charlesmindee:
This PR adds the support for blank img in OCRDataset.Any feedback is welcome!

### DIFF
--- a/doctr/datasets/ocr.py
+++ b/doctr/datasets/ocr.py
@@ -47,6 +47,10 @@ class OCRDataset(AbstractDataset):
             # Get image path
             img_name = Path(os.path.basename(file_dic["raw-archive-filepath"])).stem + '.jpg'
             box_targets = []
+            # handle empty images
+            if file_dic["coordinates"] == ['N/A']:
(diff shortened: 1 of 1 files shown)


doctr/datasets/ocr.py
```diff
@@ -47,6 +47,10 @@ def __init__(
             # Get image path
             img_name = Path(os.path.basename(file_dic["raw-archive-filepath"])).stem + '.jpg'
             box_targets = []
+            # handle empty images
+            if file_dic["coordinates"] == ['N/A']:
```
fg-mindee, line 51: That will throw warnings if there are several coordinates, I suggest:
```python
if len(file_dic["coordinates"]) == 0 or (len(file_dic["coordinates"]) == 1 and file_dic["coordinates"][0] == "N/A"):
```

fg-mindee (changes requested):
Thanks for the PR! Small improvement suggestion :ok_hand:

fg-mindee (approved):
Thanks!

PR #272, opened 2021-05-20
refactor: improved DB postprocessing

charlesmindee:
This PR simplify the postprocessing pipeline: we directly fit a bounding rectangle to contours instead of approximating contours by polygons. This lead to much better recall results:
after
Any feedback is welcome!


fg-mindee (changes requested):
Thanks for the PR! Since this has direct consequences on DBNet performances, could you also update the benchmark perf (IoU>=0.5) accordingly so we can measure whether this is a positive modification?

fg-mindee (approved):
Very nice improvement! Thanks a lot for the benchmark update

PR #276, opened 2021-05-21
fix: multi-page demo & RGB for pdf reader

charlesmindee:
This PR fixes 2 bugs spotted on the demo:
- multi-pages pdf were crashing the app
- pdf reader as images was outputing BGR instead of RGB

Any feedback is welcome!


fg-mindee (approved):
Thanks for the fix!

PR #278, opened 2021-05-21
refactor: Refactored dataset methods and removed unused imports

fg-mindee:
- refactored duplicated dataset methods
- forced fp32 for box targets of `OCRDataset` (previously np.float64)
- removed unused imports
- removed unused variables
- added unittests for `OCRDataset`

Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR!

PR #280, opened 2021-05-25
docs: Added num of params in tables

fg-mindee:
This PR adds the number of parameters in each architecture using the following snippet:

```python
import numpy as np
np.sum([np.prod(v.get_shape()) for v in model.trainable_weights])
```


charlesmindee (approved):
Thanks for the PR !

PR #282, opened 2021-05-26
fix: Fixed image decoding in API template

fg-mindee:
- fixed default channels in image decoding (previously 0 = grayscale, now 3)
- enforced correct resizing by passing list of TF tensor rather than plan tensor
- updated README and CONTRIBUTING references to the API template

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #284, opened 2021-05-27
feat: Added recognition sequence probability

fg-mindee:
- adds computation of the sequence probabilities in text recognition model (CRNN: switch from [greedy decoder](https://www.tensorflow.org/api_docs/python/tf/nn/ctc_greedy_decoder) to [beam search decoder](https://www.tensorflow.org/api_docs/python/tf/nn/ctc_beam_search_decoder) / SAR: took the minimum character probability in the sequence)
- reflected changes on API, training and evaluation scripts
- updated unittests
- removed `ignore_case` and `ignore_accents` from recognition postprocessors as those operations can be conducted by the user on its own after the export.

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #285, opened 2021-05-28
docs: Improved landing page of the documentation

fg-mindee:
- added README illustration in the documentation
- reordered sections
- changed model zoo listing to display only the name of the model (rather than adding the full name of the paper)

Here is how it renders:
image

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #290, opened 2021-06-01
feat: Added IoU computation for masks

fg-mindee:
- added vectorized function to compute IoU between masks
- added function to convert boxes to masks (some extra vectorization is needed to make this optimal)

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #291, opened 2021-06-01
fix: Fixed recognition output unpacking

fg-mindee:
This PR fixes prediction unpacking from recognition models when there are none.


charlesmindee (approved):
Thanks!

PR #294, opened 2021-06-03
feat: improve loading time when test only

charlesmindee:
This PR moves the train set loader after the test-only check to avoid a 1-min wait when there is no need to load the full train set.

Any feedback is welcome!


fg-mindee (approved):
Thanks!

PR #296, opened 2021-06-03
feat: add edge mask in linknet compute_target

charlesmindee:
This PR adds an edge_mask in linknet compute_target and compute_loss to enhance spatial resolution of boxes. 
It will add more weight to the edges of boxes, and thus help the model to easily separate words.

Any feedback is welcome!

### DIFF
--- a/doctr/models/detection/linknet.py
+++ b/doctr/models/detection/linknet.py
@@ -221,12 +221,13 @@ class LinkNet(DetectionModel, NestedObject):
         self,
         target: List[Dict[str, Any]],
         output_shape: Tuple[int, int, int],
-    ) -> Tuple[tf.Tensor, tf.Tensor]:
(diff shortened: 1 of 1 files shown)


doctr/models/detection/linknet.py
```diff
@@ -262,16 +263,24 @@ def compute_target(
...
 
-        return seg_target, seg_mask
+        return seg_target, seg_mask, edge_mask
 
     def compute_loss(
         self,
         out_map: tf.Tensor,
-        target: List[Dict[str, Any]]
+        target: List[Dict[str, Any]],
+        factor: int = 2
```
fg-mindee, line 283: let's use a float

fg-mindee (changes requested):
Thanks for the PR! Only the arg type to change

fg-mindee (approved):
Looks good to me!

PR #298, opened 2021-06-08
test: Fixed ColorInversion unittest

fg-mindee:
This PR fixes the edge case on the unittest of ColorInversion


charlesmindee (approved):
Thanks!

PR #299, opened 2021-06-08
fix: Fixed CORD dataset and refactored dataset constructors

fg-mindee:
- added bounding box clipping to fix negative coordinates on CORD dataset
- fixed dataset docstrings
- refactored dataset constructors

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #304, opened 2021-06-11
feat: add focal loss in linknet model

charlesmindee:
This PR adds a compute_focal_loss method in Linknet, the old compute_loss method  is now compute_bce_loss. The focal loss is enabled by default, but the two are available with a focal_loss flag in the call.

Any feedback is welcome!

### DIFF
--- a/doctr/models/detection/linknet.py
+++ b/doctr/models/detection/linknet.py
@@ -276,25 +276,25 @@ class LinkNet(DetectionModel, NestedObject):
 
         return seg_target, seg_mask, edge_mask
 
-    def compute_loss(
+    def compute_bce_loss(
         self,
         out_map: tf.Tensor,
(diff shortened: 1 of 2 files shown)


doctr/models/detection/linknet.py
```diff
@@ -276,25 +276,25 @@ def compute_target(
 
         return seg_target, seg_mask, edge_mask
 
-    def compute_loss(
+    def compute_bce_loss(
```
fg-mindee, line 279: it would be better to keep `compute_loss`, perhaps we should add a bool arg `use_focal` that defaults to False

fg-mindee (commented):
Thanks for the focal loss! Small suggestion for the method flexibility

fg-mindee (approved):
Thanks for the edits!

Issue #305, opened 2021-06-11
[pytorch] Add PyTorch support


fg-mindee:
This is a tracker for PyTorch integration for DocTR release 0.3.0

- [x] Enable flexible framework selection for build & imports (#306)
- [ ] PyTorch mirror of doctr.datasets
- [ ] PyTorch mirror of doctr.models (#310)
- [ ] PyTorch mirror of doctr.transforms

fg-mindee:
Closing this since it's a duplicate of #261

PR #306, opened 2021-06-11
feat: Added a dual framework availability checker

fg-mindee:
Following up on #305, this PR introduces the following modifications:
- added a dynamic framework selection module (which can then be used to dynamically import the correct modules in `__init__.py`)
- updated all installation instructions
- updated installation commands
- added a unittest

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #310, opened 2021-06-13
feat: Added DBNet implementation in PyTorch

fg-mindee:
Following up on #305, this PR introduces the following modifications:
- added implementation of DBNet (without postprocessor or loss computation so far)
- added unittests for pytorch codebase
- added CI job to run pytorch unittests

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #311, opened 2021-06-14
feat: Updated focal loss parameters

charlesmindee:
This PR sets the focal loss parameters to alpha = .5 and gamma = 2. in the Linknet.


fg-mindee (approved):
Thanks!

PR #315, opened 2021-06-16
chore: Ensured either TF or PT is installed and enabled

fg-mindee:
- raises an error if neither TF or Pytorch is installed and enabled
- split CI jobs into multiple workflows


charlesmindee (dismissed):
Thanks!

charlesmindee (approved):
Thanks!

PR #316, opened 2021-06-16
feat: Added layer initialization in DBNet

fg-mindee:
This PR simply adds initialization for layers of DBNet

Any feedback is welcome!


charlesmindee (approved):
Thanks!

Issue #261, opened 2021-05-17
[pytorch] Add PyTorch backend integration


fg-mindee:
Integrating PyTorch as a possible backend is a natural evolution of DocTR. Two questions need to be answered though, I added some suggestions.

## How do we add the PyTorch counterpart of an object / class / function?
For module `my_module.py`, 
- we create `my_module.tf.py` and put the rest in `my_module.pt.py`. If there are non framework-related features, they can be kept in `my_module`. The latter will reference the tf or pytorch implementation depending on the ibrary build
- we add a folder for each framework, and put the module named the same way in each folder for each framework.

## Where do we start?
Here are my thoughts on all DocTR modules:
- `datasets`: the `loader.py` module and the call of each dataset
- `documents`: nothing framework-specific
- `models`: every module will be impacted
- `transforms`: the call methods will be impacted if we cannot used plain torchvision transforms
- `utils`: nothing framework-specific

## How do we ensure that each build is running correctly?
- edit the `setup.py` to build the specified version
- add CI jobs to run the installation for each framework (probably will have to rethink unittests for each separate framework build)

## Next steps
- [x] Enable flexible framework selection for build & imports (#306, #314, #315)
- [x] PyTorch mirror of doctr.datasets (#319)
- [x] PyTorch mirror of doctr.models
    - [x] doctr.models.preprocessor (#313)
    - [x] doctr.models.backbones (#327, #328)
    - [x] doctr.models.utils (#323)
    - [x] doctr.models.detection (#310, #313, #316, #317)
    - [x] doctr.models.recognition (#318, #329, #333, #335, #340, #342)
- [x] PyTorch mirror of doctr.transforms (#313, #322)

fg-mindee:
Closing this since ONNX support and reference script are a bit far off from this PR

PR #317, opened 2021-06-20
feat: Added support of LinkNet in PyTorch

fg-mindee:
- added PyTorch support of LinkNet architectures
- renamed `out_chan` to `num_classes` for the number of output classes in detection models
- refactored preprocessor call in tensorflow
- updated unittests

Any feedback is welcome!

### DIFF
--- a/doctr/models/detection/linknet/pytorch.py
+++ b/doctr/models/detection/linknet/pytorch.py
@@ -233,7 +233,7 @@ class LinkNet(nn.Module, _LinkNet):
 
         else:
             # Compute BCE loss with highlighted edges
-            loss = ((1 + (edge_factor - 1) * torch.from_numpy(edge_mask)) * bce).mean()
+            loss = ((1 + (edge_factor - 1) * edge_mask) * bce).mean()
 
         return loss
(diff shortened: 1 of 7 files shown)


doctr/models/detection/linknet/pytorch.py
```diff
@@ -0,0 +1,275 @@
...
+        )
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        out = self.stage1(x)
+        if self.shortcut is not None:
+            out += self.shortcut(x)
+        out = F.relu(out, inplace=True)
+        out = self.stage2(out) + out
+
+        return out
```
charlesmindee, line 64: ResNet block, maybe we should code it in resnet.py and then import it here, but it can be refactored later.

charlesmindee (approved):

doctr/models/detection/linknet/pytorch.py
fg-mindee: Yes I thought I would have to modify a few things first! I'll do this in a follow-up PR

PR #319, opened 2021-06-21
feat: Added support of Pytorch for datasets

fg-mindee:
- removed tensorflow deps of `doctr.models.utils.download_from_url`
- moved it to `doctr.models.data_utils.download_from_url`
- made a base/pytorch/tensorflow split for the core classes of `doctr.datasets` and have all supported implementations inheriting from them
- updated requirements
- added unittests for pytorch datasets

Please note that the `DataLoader` already exists in pure pytorch thus no need to reimplement it!


charlesmindee (dismissed):
Thanks!

charlesmindee (approved):
Thanks

PR #320, opened 2021-06-21
feat: add a drawing_page function in utils & demo

charlesmindee:
This PR adds a drawing function in utils/visualization.py, which is called in the demo (samples below).

Any feedback is welcome!

Example of draw_page:
Screenshot_2021-06-21 app · Streamlit(5) Screenshot_2021-06-21 app · Streamlit(4)

Demo1:

Screenshot_2021-06-21 app · Streamlit(3)

Demo2:

Screenshot_2021-06-21 app · Streamlit(2)

Demo3:

Screenshot_2021-06-21 app · Streamlit(1)


fg-mindee (commented):
Thanks for the PR!
Two high-level comments: we might need to find a better font size resolution method (on your last examples it looks too small), and regarding the demo, considering the user would have to scroll down, do you think we should display on 2 rows? or keep 1 row and find another layout option?

charlesmindee:
For the font size, it is very hard to find a global rule because for dense texts it is hard to avoid overlapping between words. I can easily augment the font size but it will be hard to separate words for dense blocks. 
image
For the display, I agree but I think displaying the four pics in the same row would be unreadable, do you have any suggestion ?

doctr/utils/visualization.py
```diff
@@ -170,6 +171,53 @@ def visualize_page(
...
+            for word in line["words"]:
+                # Resize word geometry
+                (xmin, ymin), (xmax, ymax) = word["geometry"]
+                xmin, xmax = w * xmin, w * xmax
+                ymin, ymax = h * ymin, h * ymax
+
+                # Font computation
+                smoothing.append(ymax - ymin)  # Update line smoother
+                # Convert Pix -> Pts, add line smoothing and offset to reduce boxes to the size of characters.
+                font_size = int(.75 * ((ymax - ymin + np.mean(smoothing)) / 2 - 4))
```
fg-mindee, line 212: This needs to be commented as well otherwise flake8 is unhappy :sweat_smile:

fg-mindee (changes requested):
I really don't think it's a good idea to add the font file in the library. Would you mind removing it? If it's for the default font, perhaps we should move everything to OpenCV rather than Pillow :man_shrugging:

charlesmindee:
This is the rendering after refactoring the writing process (fixed-size font):
Screenshot_2021-06-23 app · Streamlit(3)
Screenshot_2021-06-23 app · Streamlit(2)
Screenshot_2021-06-23 app · Streamlit(1)

fg-mindee (approved):
Thanks for the edits!

Issue #225, opened 2021-04-29
[models] detect page orientation


charlesmindee:
We should be able to detect the rotation of a page (angle from 0 to 359°) to redress the page before sending it to the OCR.
This would highly improve our predictor on "tricky" datasets where pages are often rotated.

Some documents are very complex and have areas of text with different orientations, but even for these documents we can define a main orientation for the page (most of the lines would be oriented this way).

This leads us to define 2 levels of orientation:
- **Page orientation**: for most of the documents it would be the orientation of all text lines, for tricky documents the main orientation of the lines (most of the lines are oriented this way).
- **Box/line/block level orientation**: Once the document is rotated (after the detection of page orientation), find the areas which are still rotated and redress only these areas.

Any suggestion is welcome! I think firstly we should implement page orientation which should be relatively easy and then focus on the other part which is far more trickier.

@fg-mindee

Rob192:
Hello ! 

First of all, thank you for your work, this library is very good.
I am having issues with the detection of skewed documents (french ID cards). I can share my thoughts on this matter, maybe we can discuss them. 

1. I found that the method `get_bitmap_angle` is not very robust in estimating the general skew angle of the image. By using a fixed `n_ct` of 20, you get much noise from small boxes that are not very relevant. These artefacts will artifically increase the std and eventually result in not changing the angle. I suggest we use a smaller n_ct or a n_ct based on a ratio (top 5% bigger boxes) or even use all the contours alltogether.
2. When passing the boxes and the associated angle resulting from the `detection_predictor` to the `recognition_predictor`, I found that the results are significantly less reliable compared with the results obtained when first rotating the document with the angle and then using the complete pipeline. 

For my use case, I will currently stay with the method I have for the french ID card which is : 
1. Detect the side of the ID card
2. Rotate the card based on a canvas with a `DecriptorMatcher` from OpenCV
3. Apply DocTr on rotated document. 
4. Enjoy

fg-mindee:
Hi there @Rob192 :wave: 

Glad you found the library useful!
About both your points, I fully agree that this is still not optimal. We're aiming at stabilizing the rotation feature for our next release. To give you more information on this topic:

[...]

PR #324, opened 2021-06-22
Orient images before processing OCR

Rob192:
Following our discussion on #225 I propose the following PR to orientate the document before the processing with the ocr_predictor. In my experience, this helps both the `detection_predictor`and the `recognition_predictor`.

### DIFF
--- a/doctr/documents/reader.py
+++ b/doctr/documents/reader.py
@@ -8,6 +8,8 @@ import cv2
 from pathlib import Path
 import fitz
 from weasyprint import HTML
+from math import floor
+from statistics import median
 from typing import List, Tuple, Optional, Any, Union, Sequence, Dict
 
 __all__ = ['read_pdf', 'read_img', 'read_html', 'DocumentFile', 'PDF']
@@ -17,11 +19,55 @@ AbstractPath = Union[str, Path]
 AbstractFile = Union[AbstractPath, bytes]
 Bbox = Tuple[float, float, float, float]
 
+def orient_image(img):
+    gray_img = cv2.cvtColor(img, cv2.COLOR_BGR2GRAY)
+    gray_img = cv2.medianBlur(gray_img, 5)
+    thresh = cv2.threshold(gray_img, thresh=0, maxval=255, type=cv2.THRESH_BINARY_INV + cv2.THRESH_OTSU)[1]
+
+    # try to merge words in lines
+    (h, w) = img.shape[:2]
+    k_x = max(1, (floor(w / 100)))
+    k_y = max(1, (floor(h / 100)))
+    kernel = cv2.getStructuringElement(cv2.MORPH_RECT, (k_x, k_y))
+    thresh = cv2.dilate(thresh, kernel, iterations=1)
+
+    # extract contours
+    contours, _ = cv2.findContours(thresh, cv2.RETR_LIST, cv2.CHAIN_APPROX_SIMPLE)
+
+    # Sort contours
+    contours = sorted(contours, key=cv2.contourArea, reverse=True)
+
+    n_ct = 50
+    ratio_threshold_for_lines = 5
+
+    angles = []
+    for contour in contours[:n_ct]:
+        _, (w, h), angle = cv2.minAreaRect(contour)
+        if w / h > ratio_threshold_for_lines:  # select only contours with ratio like lines
+            angles.append(angle)
+        elif w / h < 1 / ratio_threshold_for_lines:  # if lines are vertical, substract 90 degree
+            angle -= 90
+            angles.append(angle)
+    angle = median(angles)
+
+    if angle < -45:
+        # if image is closer to portrait mode, go for a full 90° rotation
+        img = cv2.rotate(img, cv2.cv2.ROTATE_90_CLOCKWISE)
+        angle += 90 #track the rotation by updating angle
+
+    (h, w) = img.shape[:2]
+    center = (w // 2, h // 2)
+    M = cv2.getRotationMatrix2D(center, angle, 1.0)
+    img = cv2.warpAffine(img, M, (w, h),
+                         flags=cv2.INTER_CUBIC, borderMode=cv2.BORDER_REPLICATE)
+    return img
+
 
 def read_img(
     file: AbstractFile,
     output_size: Optional[Tuple[int, int]] = None,
     rgb_output: bool = True,
+    orient_img: bool = True,
 ) -> np.ndarray:
     """Read an image file into numpy format
 
@@ -56,6 +102,8 @@ def read_img(
     # Switch the channel order
     if rgb_output:
         img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
+    if orient_img:
+        img = orient_image(img)
     return img
 
 
--- a/doctr/documents/reader.py
+++ b/doctr/documents/reader.py
@@ -8,8 +8,6 @@ import cv2
 from pathlib import Path
 import fitz
 from weasyprint import HTML
-from math import floor
-from statistics import median
 from typing import List, Tuple, Optional, Any, Union, Sequence, Dict
 
 __all__ = ['read_pdf', 'read_img', 'read_html', 'DocumentFile', 'PDF']
@@ -19,55 +17,11 @@ AbstractPath = Union[str, Path]
 AbstractFile = Union[AbstractPath, bytes]
 Bbox = Tuple[float, float, float, float]
 
-def orient_image(img):
-    gray_img = cv2.cvtColor(img, cv2.COLOR_BGR2GRAY)
-    gray_img = cv2.medianBlur(gray_img, 5)
-    thresh = cv2.threshold(gray_img, thresh=0, maxval=255, type=cv2.THRESH_BINARY_INV + cv2.THRESH_OTSU)[1]
-
-    # try to merge words in lines
-    (h, w) = img.shape[:2]
-    k_x = max(1, (floor(w / 100)))
-    k_y = max(1, (floor(h / 100)))
-    kernel = cv2.getStructuringElement(cv2.MORPH_RECT, (k_x, k_y))
-    thresh = cv2.dilate(thresh, kernel, iterations=1)
-
-    # extract contours
-    contours, _ = cv2.findContours(thresh, cv2.RETR_LIST, cv2.CHAIN_APPROX_SIMPLE)
-
-    # Sort contours
-    contours = sorted(contours, key=cv2.contourArea, reverse=True)
-
-    n_ct = 50
-    ratio_threshold_for_lines = 5
-
-    angles = []
-    for contour in contours[:n_ct]:
-        _, (w, h), angle = cv2.minAreaRect(contour)
-        if w / h > ratio_threshold_for_lines:  # select only contours with ratio like lines
-            angles.append(angle)
-        elif w / h < 1 / ratio_threshold_for_lines:  # if lines are vertical, substract 90 degree
-            angle -= 90
-            angles.append(angle)
-    angle = median(angles)
-
-    if angle < -45:
-        # if image is closer to portrait mode, go for a full 90° rotation
-        img = cv2.rotate(img, cv2.cv2.ROTATE_90_CLOCKWISE)
-        angle += 90 #track the rotation by updating angle
-
-    (h, w) = img.shape[:2]
-    center = (w // 2, h // 2)
-    M = cv2.getRotationMatrix2D(center, angle, 1.0)
-    img = cv2.warpAffine(img, M, (w, h),
-                         flags=cv2.INTER_CUBIC, borderMode=cv2.BORDER_REPLICATE)
-    return img
-
 
 def read_img(
     file: AbstractFile,
     output_size: Optional[Tuple[int, int]] = None,
     rgb_output: bool = True,
-    orient_img: bool = True,
 ) -> np.ndarray:
     """Read an image file into numpy format
 
@@ -102,8 +56,6 @@ def read_img(
     # Switch the channel order
     if rgb_output:
         img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
-    if orient_img:
-        img = orient_image(img)
     return img
 
 
--- a/doctr/models/_utils.py
+++ b/doctr/models/_utils.py
@@ -89,9 +89,9 @@ def extract_rcrops(img: np.ndarray, boxes: np.ndarray) -> List[np.ndarray]:
 
 
 def rotate_page(
-        image: np.ndarray,
-        angle: float = 0.,
-        min_angle: float = 1.
+    image: np.ndarray,
+    angle: float = 0.,
+    min_angle: float = 1.
 ) -> np.ndarray:
     """Rotate an image counterclockwise by an ange alpha (negative angle to go clockwise).
 
@@ -120,7 +120,6 @@ def rotate_page(
 def estimate_orientation(img: np.ndarray, n_ct: int = 50, ratio_threshold_for_lines: int = 5) -> int:
     """Estimate the angle of the general document orientation based on the
      lines of the document and the assumption that they should be horizontal.
-
 
         Args:
             img: the img to analyze
@@ -196,9 +195,9 @@ def get_bitmap_angle(bitmap: np.ndarray, n_ct: int = 20, std_max: float = 3.) ->
 
 
 def rotate_boxes(
-        boxes: np.ndarray,
-        angle: float = 0.,
-        min_angle: float = 1.
+    boxes: np.ndarray,
+    angle: float = 0.,
+    min_angle: float = 1.
 ) -> np.ndarray:
     """Rotate a batch of straight bounding boxes (xmin, ymin, xmax, ymax) of an angle,
     if angle > min_angle, around the center of the page.
--- a/doctr/models/_utils.py
+++ b/doctr/models/_utils.py
@@ -5,9 +5,11 @@
 
 import numpy as np
 import cv2
+from math import floor
 from typing import List
+from statistics import median_low
 
-__all__ = ['extract_crops', 'extract_rcrops', 'rotate_page', 'get_bitmap_angle', 'rotate_boxes']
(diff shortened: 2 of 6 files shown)


doctr/documents/reader.py
```diff
@@ -17,11 +19,55 @@
...
+
+    angles = []
+    for contour in contours[:n_ct]:
+        _, (w, h), angle = cv2.minAreaRect(contour)
+        if w / h > ratio_threshold_for_lines:  # select only contours with ratio like lines
+            angles.append(angle)
+        elif w / h < 1 / ratio_threshold_for_lines:  # if lines are vertical, substract 90 degree
+            angle -= 90
+            angles.append(angle)
+    angle = median(angles)
```
fg-mindee, line 51: This part of the function might be better as a standalone function `estimate_orientation` and we'll need to add some typing annotations :sweat_smile:

doctr/documents/reader.py
```diff
@@ -17,11 +19,55 @@
...
+    thresh = cv2.dilate(thresh, kernel, iterations=1)
+
+    # extract contours
+    contours, _ = cv2.findContours(thresh, cv2.RETR_LIST, cv2.CHAIN_APPROX_SIMPLE)
+
+    # Sort contours
+    contours = sorted(contours, key=cv2.contourArea, reverse=True)
+
+    n_ct = 50
+    ratio_threshold_for_lines = 5
```
fg-mindee, line 41: Let's make those default values of function's arguments

doctr/documents/reader.py
fg-mindee: Since this isn't a "reading" feature, I'd suggest either moving it to `doctr.models.utils` or `doctr.documents.utils`

doctr/documents/reader.py
```diff
@@ -17,11 +19,55 @@
...
+        # if image is closer to portrait mode, go for a full 90° rotation
+        img = cv2.rotate(img, cv2.cv2.ROTATE_90_CLOCKWISE)
+        angle += 90 #track the rotation by updating angle
+
+    (h, w) = img.shape[:2]
+    center = (w // 2, h // 2)
+    M = cv2.getRotationMatrix2D(center, angle, 1.0)
+    img = cv2.warpAffine(img, M, (w, h),
+                         flags=cv2.INTER_CUBIC, borderMode=cv2.BORDER_REPLICATE)
+    return img
```
fg-mindee, line 63: this part will need to be integrated in the predictor, but it's dangerous sending back only the rotated image. It means that we won't be able to reproject our predictions in the original image.

doctr/documents/reader.py
```diff
@@ -56,6 +102,8 @@ def read_img(
     # Switch the channel order
     if rgb_output:
         img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
+    if orient_img:
+        img = orient_image(img)
```
fg-mindee, line 106: Let's make this an option in the predictor, but it's not a reading feature

fg-mindee (changes requested):
Thanks a lot for the PR!

Two high-level comments:
- might be better to split your main function & focus this PR on angle estimation (as a standalone feature with your method, which we'll integrate in the predictors in another PR)
- as mentioned, the angle estimation part is not a reading feature so we'd better move this to `doctr.documents.utils` or `doctr.models.utils` :man_shrugging: 

Once this is done, we're gonna have to add typing annotations & a small unittest :+1: 

Let me know what you think!

doctr/documents/reader.py
Rob192: OK, then I think it will fit well inside `doctr.models._utils`

doctr/documents/reader.py
Rob192: OK will do

doctr/documents/reader.py
Rob192: I am not quite sure about where/how to do that ... Should we change the `DocumentBuilder` to accept a page_angle parameter that would describe the global page orientation ? We would then modify `visualize_page` in `doctr.utils.visualization` to rotate back everything.

doctr/documents/reader.py
Rob192: ok will do

Rob192:
Hello @fg-mindee 
Thank you for the throrough review. I implemented most of you comments. I am still thinking how to do two things : 
- I need to find a way to build back the original image at the end of the OCRPredictor. My first idea would be to change the DocumentBuilder to accept a page_angle parameter that would describe the global page orientation ? I would then modify visualize_page in doctr.utils.visualization to rotate back everything. Please tell me your thoughts on this matter, I am afraid I do not have the complete picture here. 
- I still need to add some testing. I will do that ASAP.

doctr/models/_utils.py
```diff
@@ -87,9 +89,9 @@ def extract_rcrops(img: np.ndarray, boxes: np.ndarray) -> List[np.ndarray]:
 
 
 def rotate_page(
-    image: np.ndarray,
-    angle: float = 0.,
-    min_angle: float = 1.
+        image: np.ndarray,
+        angle: float = 0.,
+        min_angle: float = 1.
```
fg-mindee, line 94: Would you mind reverting that change? It will throw linter check errors

doctr/models/_utils.py
```diff
@@ -148,9 +196,9 @@ def get_bitmap_angle(bitmap: np.ndarray, n_ct: int = 20, std_max: float = 3.) ->
 
 
 def rotate_boxes(
-    boxes: np.ndarray,
-    angle: float = 0.,
-    min_angle: float = 1.
+        boxes: np.ndarray,
+        angle: float = 0.,
+        min_angle: float = 1.
```
fg-mindee, line 201: same here

doctr/models/utils.py
```diff
@@ -16,17 +16,16 @@
...
-    hash_prefix: Optional[str] = None,
-    overwrite: bool = False,
-    internal_name: str = 'weights',
-    **kwargs: Any,
+        model: Model,
+        url: Optional[str] = None,
+        hash_prefix: Optional[str] = None,
+        overwrite: bool = False,
+        internal_name: str = 'weights',
+        **kwargs: Any,
```
fg-mindee, line 28: same here

doctr/models/utils.py
```diff
@@ -58,12 +57,12 @@ def load_pretrained_params(
...
-    bn: bool = False,
-    padding: str = 'same',
-    kernel_initializer: str = 'he_normal',
-    **kwargs: Any,
+        out_channels: int,
+        activation: str = None,
+        bn: bool = False,
+        padding: str = 'same',
+        kernel_initializer: str = 'he_normal',
+        **kwargs: Any,
```
fg-mindee, line 65: same here

doctr/models/utils.py
```diff
@@ -83,7 +82,7 @@ def conv_sequence(
         list of layers
     """
     # No bias before Batch norm
-    kwargs['use_bias'] = kwargs.get('use_bias', not(bn))
+    kwargs['use_bias'] = kwargs.get('use_bias', not (bn))
```
fg-mindee, line 85: same here

doctr/models/utils.py
```diff
@@ -112,10 +111,11 @@ class IntermediateLayerGetter(Model):
...
         layer_names: the list of layers to retrieve the feature map from
     """
+
     def __init__(
-        self,
-        model: Model,
-        layer_names: List[str]
+            self,
+            model: Model,
+            layer_names: List[str]
```
fg-mindee, line 118: same here

doctr/models/utils.py
```diff
@@ -112,10 +111,11 @@ class IntermediateLayerGetter(Model):
         model: the model to extract feature maps from
         layer_names: the list of layers to retrieve the feature map from
     """
+
```
fg-mindee, line 114: let's avoid modifying files that aren't related to this PR :)

doctr/models/utils.py
```diff
@@ -16,17 +16,16 @@
 
 logging.getLogger("tensorflow").setLevel(logging.DEBUG)
 
-
```
fg-mindee, line 19: let's revert this

doctr/models/core.py
```diff
@@ -42,13 +42,22 @@ def __init__(
...
             raise ValueError("incorrect input shape: all pages are expected to be multi-channel 2D images.")
 
+        # Detect document rotation and rotate pages
+        if rotate_document:
+            page_angles = []
+            for i,page in enumerate(pages):
+                page_angle = estimate_orientation(page)
+                page_angles.append(page_angle)
+                pages[i] = rotate_page(page,page_angle)
+
```
fg-mindee, line 60: in this PR, I suggest discarding the integration in the predictor and keeping this only about orientation estimation :+1:

doctr/models/_utils.py
```diff
@@ -103,10 +105,56 @@ def rotate_page(
     """
     if abs(angle) < min_angle or abs(angle) > 90 - min_angle:
         return image
+    elif abs(angle) > 45:
+        # if image is closer to portrait mode, go for a full 90° rotation
+        image = cv2.rotate(image, cv2.cv2.ROTATE_90_CLOCKWISE)
+        angle += 90  # track the rotation by updating angle
```
fg-mindee, line 111: I might be missing something but: if the angle is higher than 45, and we perform first-hand a 90° rotation, then the remaining rotation to perform is `angle - 90` and not `angle + 90` ?

doctr/models/_utils.py
```diff
@@ -103,10 +105,56 @@ def rotate_page(
...
+
     height, width = image.shape[:2]
     center = (height / 2, width / 2)
     rot_mat = cv2.getRotationMatrix2D(center, angle, 1.0)
-    return cv2.warpAffine(image, rot_mat, (width, height))
+    return cv2.warpAffine(image, rot_mat, (width, height),
+                          flags=cv2.INTER_CUBIC, borderMode=cv2.BORDER_REPLICATE)
```
fg-mindee, line 117: the borderMode should be a new option: I suggest not modifying this function in this PR for now (otherwise, we'd need to benchmark the impact of this change on performances right away :upside_down_face: )

doctr/models/_utils.py
```diff
@@ -103,10 +105,56 @@ def rotate_page(
...
+
+
+def estimate_orientation(img: np.ndarray, n_ct: int = 50, ratio_threshold_for_lines: int = 5) -> int:
+    """Estimate the angle of the general document orientation based on the
+     lines of the document and the assumption that they should be horizontal.
+
+
```
fg-mindee, line 124: You can remove the second blank line ;)

doctr/models/_utils.py
```diff
@@ -103,10 +105,56 @@ def rotate_page(
...
     height, width = image.shape[:2]
     center = (height / 2, width / 2)
     rot_mat = cv2.getRotationMatrix2D(center, angle, 1.0)
-    return cv2.warpAffine(image, rot_mat, (width, height))
+    return cv2.warpAffine(image, rot_mat, (width, height),
+                          flags=cv2.INTER_CUBIC, borderMode=cv2.BORDER_REPLICATE)
+
+
+def estimate_orientation(img: np.ndarray, n_ct: int = 50, ratio_threshold_for_lines: int = 5) -> int:
```
fg-mindee, line 120: the return type is a `float`

fg-mindee: `line_aspect_ratio` might be better for the last argument? (and it's a `float` as well)

doctr/models/_utils.py
```diff
@@ -103,10 +105,56 @@ def rotate_page(
...
+    k_x = max(1, (floor(w / 100)))
+    k_y = max(1, (floor(h / 100)))
+    kernel = cv2.getStructuringElement(cv2.MORPH_RECT, (k_x, k_y))
+    thresh = cv2.dilate(thresh, kernel, iterations=1)
+
+    # extract contours
+    contours, _ = cv2.findContours(thresh, cv2.RETR_LIST, cv2.CHAIN_APPROX_SIMPLE)
+
+    # Sort contours
+    contours = sorted(contours, key=cv2.contourArea, reverse=True)
```
fg-mindee, line 147: don't you think we should sort by `max(w/h, h/w)` to consider boxes that are further away from a 1:1 ratio? Here we give priority to big boxes but a big square bounding box will not help up to estimate the orientation

doctr/models/_utils.py
```diff
@@ -103,10 +105,56 @@ def rotate_page(
...
+
+    angles = []
+    for contour in contours[:n_ct]:
+        _, (w, h), angle = cv2.minAreaRect(contour)
+        if w / h > ratio_threshold_for_lines:  # select only contours with ratio like lines
+            angles.append(angle)
+        elif w / h < 1 / ratio_threshold_for_lines:  # if lines are vertical, substract 90 degree
```
fg-mindee, line 154: Perhaps there is an extra condition to be added here but I'm not sure: a condition to discard single character boxes that are much higher than wide (e.g. `l`)

doctr/models/_utils.py
```diff
@@ -103,10 +105,56 @@ def rotate_page(
...
+            angle -= 90
+            angles.append(angle)
```
fg-mindee, line 156:
```
angles.append(angle - 90)
```

fg-mindee (changes requested):
Thanks for the edits! First, could you merge "main" into your branch? There are some conflicts apparently. Sorry my previous review wasn't clear enough, but I believe it would be better to consider several PRs here: 1 to estimate the orientation (this one), another later on to integrate it in the predictor and making sure we can reproject the end results, and a last one to check whether we should make this a default by doing a performance check.

I might be wrong but I suspect that this method might work quite well for some specific documents with well-defined lines (french ID cards for instance), but it would lack generalization to other use cases. So let's investigate this carefully :)

doctr/models/_utils.py
Rob192: ok will do

doctr/models/_utils.py
Rob192: ok will do

doctr/models/utils.py
Rob192: ok will do

doctr/models/utils.py
Rob192: ok will do

doctr/models/utils.py
Rob192: ok will do. sorry about that, i let Pycharm correct everything without checking :/

doctr/models/_utils.py
Rob192: ok

doctr/models/_utils.py
Rob192: ok done

doctr/models/_utils.py
Rob192: yes good point !

doctr/models/_utils.py
Rob192: The issue here is that this function is blind from what's inside the boxes. I though that taking the median would remove this kind of outliers.

doctr/models/_utils.py
Rob192: In `cv2.getRotationMatrix2D`, positive angle values mean counter-clockwise rotation. Here I assume the image needs clockwise rotation.

doctr/models/_utils.py
Rob192: OK, I will move this away. BTW, how do you benchmark the performances ?

doctr/models/core.py
Rob192: ok got it !!

doctr/models/utils.py
Rob192: Sorry about that !

doctr/models/_utils.py
Rob192: @fg-mindee is it OK for you if I leave it as it is ?

Rob192:
Hello @fg-mindee
thanks for your review. I split the PRs into three different one. 
Could you please point me to a dataset for benchmarking the changes for the image rotation ? Or even better, to an evaluation script that you would use for the release of an update. 
Thanks :)

doctr/models/_utils.py
```diff
@@ -103,12 +105,64 @@ def rotate_page(
...
     if abs(angle) < min_angle or abs(angle) > 90 - min_angle:
         return image
+
     height, width = image.shape[:2]
     center = (height / 2, width / 2)
     rot_mat = cv2.getRotationMatrix2D(center, angle, 1.0)
     return cv2.warpAffine(image, rot_mat, (width, height))
 
 
+def get_max_width_length_ratio(contour: np.ndarray):
```
fg-mindee, line 115:
```python
def get_max_width_length_ratio(contour: np.ndarray) -> float:
```
to get a complete typing annotation ;)

doctr/models/_utils.py
fg-mindee: Sure, if we split this into several PRs, it's quite OK!

fg-mindee (commented):
Almost good to merge! A few adjustments and the rest looks good to me!

test/test_models.py
```diff
@@ -235,11 +235,27 @@ def mock_bitmap(tmpdir_factory):
...
 
+@pytest.fixture(scope="function")
+def mock_image(tmpdir_factory):
+    url = 'https://github.com/mindee/doctr/releases/download/v0.2.1/bitmap30.png'
+    file = BytesIO(requests.get(url).content)
+    tmp_path = str(tmpdir_factory.mktemp("data").join("mock_bitmap.jpg"))
+    with open(tmp_path, 'wb') as f:
+        f.write(file.getbuffer())
+    image = reader.read_img(tmp_path)
+    return image
```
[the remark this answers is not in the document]
Rob192: Yes will do !

Rob192:
Great !

fg-mindee (approved):
Thanks a lot for the edits!

fg-mindee:
> Hello @fg-mindee
> Thank you for your detailled review. I split the PRs into three different one.
> Could you please point me to a dataset for benchmarking the changes for the image rotation ? Or even better, to an evaluation script that you would use for the release of an update.
> Thanks :)

The repo already has an evaluation script in `scripts/evaluate.py` with support of several public datasets! But the tricky part is that those aren't rotated, so we'll have to add some test-time augmentation to check this

PR #329, opened 2021-06-25
feat: added Pytorch support of MASTER

charlesmindee:
- moved` transformer.py` and `master.py` to` master/transformer_tf.py `and `master/tensorflow.py`
- added a `master/base.py`  framework-agnostic meta-class, which is used to instantiate both tensorflow and pytorch master models.
- added a `master/transformer_pt.py` and `master/pytorch.py` for both master and transformer decoder in pytorch.
- updated tests in `pytorch/test_models_recognition.py`

For now, the master model is implemented (both forward and predict methods), another PR will follow to add loss and model in the model zoo.

Any feedback is welcome!


doctr/models/recognition/master/base.py
```diff
@@ -0,0 +1,37 @@
...
+
+import numpy as np
+from typing import List, Tuple
+from ....datasets import VOCABS, encode_sequences
+
+
+class _MASTER:
+
+    vocab: str = VOCABS['french']
+    max_length: int = 50
```
fg-mindee, line 14: I think it's a bad idea to assign values here. If the constructor doesn't do it, it will hold values that are not easily accessible to the developper.

I guess this was for typing issues ? If so you can type them without assigning a value:
```python
vocab: str
max_length: int
```

doctr/models/recognition/master/base.py
```diff
@@ -0,0 +1,37 @@
+# Copyright (C) 2021, Mindee.
+
+# This program is licensed under the Apache License version 2.
+# See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0.txt> for full license details.
+
+import numpy as np
+from typing import List, Tuple
+from ....datasets import VOCABS, encode_sequences
```
fg-mindee, line 8: You can remove the VOCABS import if my next comment sounds relevant to you

doctr/models/recognition/master/pytorch.py
```diff
@@ -0,0 +1,290 @@
...
+
+__all__ = ['MASTER']
+
+
+default_cfgs: Dict[str, Dict[str, Any]] = {
+    'master': {
+        'mean': (.5, .5, .5),
+        'std': (1., 1., 1.),
+        'input_shape': (3, 48, 160),
+        'post_processor': 'MASTERPostProcessor',
```
fg-mindee, line 26: You can remove this (since it's instantiated in the model by default, the information is no longer useful)

doctr/models/recognition/master/pytorch.py
```diff
@@ -0,0 +1,290 @@
...
+        context = torch.matmul(shortcut, context_mask)
+
+        # [N, headers * C', 1, 1]
+        context = context.view(batch, self.headers * self.single_header_inplanes, 1, 1)
+
+        return context
+
+    def forward(self, inputs: torch.Tensor, **kwargs) -> torch.Tensor:
+        # Context modeling: B, C, H, W  ->  B, C, 1, 1
+        context = self.context_modeling(inputs)
```
fg-mindee, line 106: Mmmh this is a tricky topic, but for clarity, I'd suggest writing the commands in forward. I might be wrong but here are my thoughts:
if I read through the forward, I can see two calls to methods/attributes. I'm mentally guessing those are modules you instantiated in the constructor, which isn't the case for context_modeling here. If we unroll this function, the reader can see that all `self....` calls are from modules (with parameters). Mixing both looks odd to me. What do you think?

doctr/models/recognition/master/pytorch.py
```diff
@@ -0,0 +1,290 @@
...
+        max_length: maximum length of character sequence handled by the model
+        input_size: size of the image inputs
+    """
+
+    def __init__(
+        self,
+        vocab: str,
+        d_model: int = 512,
+        headers: int = 1,
+        dff: int = 2048,
```
fg-mindee, line 173: the arg name is far from being explicit :sweat_smile:

doctr/models/recognition/master/pytorch.py
```diff
@@ -0,0 +1,290 @@
...
+            num_layers=num_layers,
+            d_model=d_model,
+            num_heads=num_heads,
+            dff=dff,
+            vocab_size=self.vocab_size,
+            maximum_position_encoding=max_length,
+        )
+        self.feature_pe = positional_encoding(input_shape[1] * input_shape[2], d_model)
+        self.linear = nn.Linear(d_model, self.vocab_size + 1)
+        nn.init.kaiming_uniform_(self.linear.weight)
```
fg-mindee, line 201: In PyTorch, the default initialization is already well done: https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/linear.py#L92

Apart from if you're using a custom activation afterwards, I don't suggest changing that

doctr/models/recognition/master/pytorch.py
```diff
@@ -0,0 +1,290 @@
...
+        for m in self.modules():
+            if isinstance(m, nn.Conv2d):
+                nn.init.kaiming_normal_(m.weight, mode='fan_out', nonlinearity='relu')
+            elif isinstance(m, (nn.BatchNorm2d, nn.GroupNorm)):
+                nn.init.constant_(m.weight, 1)
+                nn.init.constant_(m.bias, 0)
+
+    def make_mask(self, target: torch.Tensor) -> torch.Tensor:
+        look_ahead_mask = create_look_ahead_mask(target.size(1))
+        target_padding_mask = create_padding_mask(target, self.vocab_size)  # Pad with EOS
```
fg-mindee, line 212: As mentioned later on, I think it's not worth making functions for those

doctr/models/recognition/master/pytorch.py
```diff
@@ -0,0 +1,290 @@
...
+        Return:
+            A torch tensor, containing logits
+        """
+
+        # Encode
+        feature = self.feature_extractor(x, **kwargs)
+        b, c, h, w = (feature.size(i) for i in range(4))
+        feature = torch.reshape(feature, shape=(b, c, h * w))
+        feature = feature.permute(0, 2, 1)  # shape (b, h*w, c)
+        encoded = feature + self.feature_pe[:, :h * w, :]
```
fg-mindee, line 241: Not sure I get it, if `feature` is of shape (N, H*W, C), how does it work that we slice the positional encoding features that way? (I mean: we may lose some information in the process if the tensor is bigger?)

doctr/models/recognition/master/pytorch.py
```diff
@@ -0,0 +1,290 @@
...
+        return torch.tile(combined_mask.permute(1, 0, 2), (self.num_heads, 1, 1))
+
+    def forward(
+        self,
+        x: torch.Tensor,
+        target: Optional[List[str]] = None,
+        return_model_output: bool = False,
+        return_preds: bool = False,
+        **kwargs: Any,
+    ) -> torch.Tensor:
```
fg-mindee, line 223: would you mind adopting the same output type as other recognition models please ? :pray: (dict with the same keys)

fg-mindee (commented):
Thanks for the edits! I did a more thorough review and left some comments! Also, you'll need to merge main to resolve the conflicts, sorry :/

doctr/models/recognition/master/transformer_pt.py
```diff
@@ -0,0 +1,97 @@
...
+def create_look_ahead_mask(size: int) -> torch.Tensor:
+    # With torch transformers, True for pad and 0 False for sequences
+    mask = ~ (torch.triu(torch.ones(size, size)) == 1).transpose(0, 1)
+    return mask[:, None, :]
+
+
+def create_padding_mask(seq: torch.Tensor, padding: int = 0) -> torch.Tensor:
+    # With torch transformers, True for pad and 0 False for sequences
+    seq = ~ torch.eq(seq, padding)
+    return seq  # (batch_size, seq_len)
```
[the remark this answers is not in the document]
charlesmindee: I think it is clearer to define the function for both masks in the transformer module as they did in the tensorflow transformer tutorial, so that you just have to call both masks from there when building master model, instead of performing low-level operations specific to transformer in MASTER.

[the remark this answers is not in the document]
charlesmindee: Tell me if I'm wrong

doctr/models/recognition/master/pytorch.py
charlesmindee: I see, I can change it. Do I need to change it in tensorflow as well ?

doctr/models/recognition/master/pytorch.py
charlesmindee: It is called that way in the paper :/

doctr/models/recognition/master/pytorch.py
charlesmindee: We compute the embedding matrix in the constructor (shape of the input) but then the input will be down-sized in the feature extractor and we need to keep only the n-th first embedding positions. I think that way you have more flexibility if you want to change the down sizing for instance

test/pytorch/test_models_recognition.py
```diff
@@ -43,3 +43,17 @@ def test_reco_postprocessors(post_processor, input_shape, mock_vocab):
...
+    input_tensor = torch.rand((batch_size, 3, 32, 128))
+    target = ["i", "am", "a", "jedi"]
+    logits = master(input_tensor, target)["out_map"]
+    assert isinstance(logits, torch.Tensor)
+    assert logits.shape == (batch_size, max_len, 1 + len(mock_vocab))  # 1 more for EOS
+    prediction, logits = master.decode(input_tensor)
+    assert isinstance(prediction, torch.Tensor)
+    assert isinstance(logits, torch.Tensor)
+    assert prediction.shape == (batch_size, max_len)
+    assert logits.shape == (batch_size, max_len, 1 + len(mock_vocab))
```
fg-mindee, line 59: To keep it similar to other unittests, I suggest creating a `master_magcresnet` function to instantiate the model and adding the test case on the `mark.parametrize` on Line 12

doctr/models/recognition/transformer/pytorch.py
```diff
@@ -0,0 +1,97 @@
...
+def create_look_ahead_mask(size: int) -> torch.Tensor:
+    # With torch transformers, True for pad and 0 False for sequences
+    mask = ~ (torch.triu(torch.ones(size, size)) == 1).transpose(0, 1)
+    return mask[:, None]
+
+
+def create_padding_mask(seq: torch.Tensor, padding: int = 0) -> torch.Tensor:
+    # With torch transformers, True for pad and 0 False for sequences
+    seq = ~ torch.eq(seq, padding)
+    return seq  # (batch_size, seq_len)
```
fg-mindee, line 42: Considering those are basically one-liner, it might be better to have everything in `master`

doctr/models/recognition/master/pytorch.py
```diff
@@ -0,0 +1,284 @@
...
+from ...backbones import resnet_stage
+from ..transformer import Decoder, positional_encoding, create_look_ahead_mask, create_padding_mask
+from .base import _MASTER
+
+__all__ = ['MASTER']
+
+
+default_cfgs: Dict[str, Dict[str, Any]] = {
+    'master': {
+        'vocab': VOCABS['french'],
```
fg-mindee, line 23: Here I meant that we had to remove `post_processor` but the rest is relevant, the predictor will need the normalization stats and it doesn't hurt keeping the input shape :)

doctr/models/recognition/master/pytorch.py
```diff
@@ -0,0 +1,284 @@
...
+from ...utils import conv_sequence_pt
+from ...backbones import resnet_stage
+from ..transformer import Decoder, positional_encoding, create_look_ahead_mask, create_padding_mask
+from .base import _MASTER
+
+__all__ = ['MASTER']
+
+
+default_cfgs: Dict[str, Dict[str, Any]] = {
+    'master': {
```
fg-mindee, line 22: we're actually missing the function to instantiate the model :sweat_smile:

fg-mindee (changes requested):
Almost there! The function to instantiate the model is still to be added :ok_hand:

doctr/models/recognition/master/pytorch.py
```diff
@@ -0,0 +1,284 @@
...
+            output = self.decoder(torch.from_numpy(gt), encoded, tgt_mask, None)
+            logits = self.linear(output)
+
+        else:
+            _, logits = self.decode(encoded)
+
+        out: Dict[str, torch.Tensor] = {}
+        out['out_map'] = logits
+
+        return out
```
fg-mindee, line 248: Could you check how the forward is implemented in the other archs ?

From what I read, the loss computation is not implemented, but still you have some lines using the "gt", what about the preds ?

test/pytorch/test_models_recognition.py
```diff
@@ -49,7 +49,7 @@ def test_master(mock_vocab, max_len=50, batch_size=4):
     master = recognition.MASTER(vocab=mock_vocab, input_shape=(3, 32, 128))
     input_tensor = torch.rand((batch_size, 3, 32, 128))
     target = ["i", "am", "a", "jedi"]
-    logits = master(input_tensor, target)["out_map"]
+    logits = master(input_tensor, target, return_model_output=True, return_preds=True)['out_map']
```
fg-mindee, line 52: Considering what's implemented yet, you don't need the `return_preds=True` (it doesn't do anything)

doctr/models/recognition/master/pytorch.py
```diff
@@ -231,19 +236,25 @@ def forward(
...
         else:
             _, logits = self.decode(encoded)
 
-        out: Dict[str, torch.Tensor] = {}
-        out['out_map'] = logits
+        if return_model_output:
+            out['out_map'] = logits
+
+        if return_preds:
+            out['preds'] = logits
```
fg-mindee: Could you remove this while the preds are not in proper format? (next PR)

fg-mindee (changes requested):
Small adjustments to make and we're good to go!

doctr/models/recognition/master/pytorch.py
```diff
@@ -0,0 +1,327 @@
...
+        out: Dict[str, torch.Tensor] = {}
+
+        if target is not None:
+            # Compute target: tensor of gts and sequence lengths
+            gt, seq_len = self.compute_target(target)
+            tgt_mask = self.make_mask(torch.from_numpy(gt))
+            # Compute logits
+            output = self.decoder(torch.from_numpy(gt), encoded, tgt_mask, None)
+            logits = self.linear(output)
+            out['out_map'] = logits
```
fg-mindee, line 248: duplicate populating of `out_map` entry, let's remove this

fg-mindee (approved):
Thanks for the edits, looks good to me!

PR #332, opened 2021-06-28
feat: add filter_boxes method

### LINKED ISSUE #330: Duplicate words in OCR result
## 🐛 Bug

Running the sample code:
```
from doctr.documents import DocumentFile

model = ocr_predictor(pretrained=True)
# PDF
doc = DocumentFile.from_images(["table.png"])
# Analyze
result = model(doc)

result.show(doc)
```

I get this result:

image

Everything looks fine but there is some overlap between different words. The mouse is pointing to the word "Header4" and there is another word with the content "4". In that case I'm not able to reconstruct properly the table header as there is either an extra "4".

## To Reproduce

1. download this image

table

2. Run the following code
```
from doctr.documents import DocumentFile

model = ocr_predictor(pretrained=True)
# PDF
doc = DocumentFile.from_images(["table.png"])
# Analyze
result = model(doc)

result.show(doc)
```

charlesmindee:
This PR adds a filter_boxes method in the core DetectionPostProcessor which performs the following:

- NMS with a high threshold to eliminate boxes included in others. NMS is implemented in Numpy (Fast-RCNN implementation).
- Merging for boxes with a consequent IOU, to avoid overlapping of letters which would appears in both words.

For now, only straight bounding boxes are supported, as the computation of NMS/boxes to merge is slightly harder when dealing with rotated boxes. In case of rotated boxes, it will simply return the array of boxes without filtering.

Any feeback is welcome!


doctr/models/detection/core.py
```diff
@@ -82,6 +83,108 @@ def bitmap_to_boxes(
     ) -> np.ndarray:
         raise NotImplementedError
 
+    @staticmethod
+    def nms(boxes: np.ndarray, thresh: float = .5) -> List[int]:
```
fg-mindee, line 87: I would argue that nms should be put in `doctr.utils.metrics` since this will be reused in many architectures and it's a common operation to perform :+1:

doctr/models/detection/core.py
```diff
@@ -82,6 +83,108 @@ def bitmap_to_boxes(
...
+
+        def enclosing(box_a: List[float], box_b: List[float]) -> List[float]:
+            """Compute enclosing bbox from 2 list of 5 floats [xmin, ymin, xmax, ymax, score].
+            Return a list of [xmin, ymin, xmax, ymax, score].
+            """
+            x1, y1, x2, y2 = box_a[:4]
+            x3, y3, x4, y4 = box_b[:4]
+            score = (box_a[4] + box_b[4]) / 2
+            enclosing = [min(x1, x3), min(y1, y3), max(x2, x4), max(y2, y4), score]
+            return enclosing
```
fg-mindee, line 169: Might be better to define it as a static class method or move it to `doctr.utils.geometry` as a function

doctr/models/detection/core.py
```diff
@@ -82,6 +83,108 @@ def bitmap_to_boxes(
...
+            return boxes
+
+        # First perform NMS on array of boxes
+        keep = self.nms(boxes, thresh=nms_thresh)
+
+        # Update box list
+        box_list = [list(box) for i, box in enumerate(list(boxes)) if i in keep]
+
+        # Compute iou beween boxes
+        boxes = np.asarray(box_list)
```
fg-mindee, line 157:
```python
boxes = boxes[keep, :]
```
to save some execution duration and recreating a numpy array ;)

doctr/models/detection/core.py
```diff
@@ -82,6 +83,108 @@ def bitmap_to_boxes(
...
+        for [i, j] in list(to_merge):
+            # Resolve enclosing box, update dictionnary
+            enclosing_box = enclosing(merged[i], merged[j])
+            merged[i] = enclosing_box
+            merged[j] = enclosing_box
+
+        # Update box list: remove duplicates from merged dictionnary
+        box_list = []
+        for box in merged.values():
+            if box not in box_list:
```
fg-mindee, line 183: That comparison doesn't look very clean to me, perhaps we should consider a better way than duplicating the enclosing box computation in the dict earlier on?

fg-mindee (changes requested):
Thanks for the PR! I added a few improvement suggestions

fg-mindee:
Also, this PR modifies the default output behaviour (cf the failed unittest in the API)

doctr/models/detection/core.py
charlesmindee: Do you have any suggestion ? I think we need to store the index of the original box for each merged box: suppose you want to merge 3 boxes i, j and k : first you merge i and j, and then you want to merge k with the merged-box associated to j, so you need to store for each index the updated merged result in a structure.

charlesmindee:
Changes: 
- nms is now implemented in `metrics`, but not used here.
- the `filter_boxes` method uses` box_ioa` (which is implemented in `metrics`) to merge boxes above a threshold. This will remove boxes included in others or merge those who overlap significantly.
- a `resolve_enclosing_boxarray` function is implemented in geometry.
- added corresponding tests.

doctr/utils/geometry.py
```diff
@@ -44,3 +45,13 @@ def resolve_enclosing_bbox(bboxes: List[BoundingBox]) -> BoundingBox:
...
+
+
+def resolve_enclosing_bboxarray(bboxarray: np.ndarray) -> np.ndarray:
+    """Compute enclosing bbox from an array of boxes: (*, 5), where boxes have this shape:
+    (xmin, ymin, xmax, ymax, score).
+    Return a (1, 5) array (enclosing boxarray)
+    """
+    xmin, ymin, xmax, ymax, score = np.split(bboxarray, 5, axis=1)
+    enclosing = [np.min(xmin), np.min(ymin), np.max(xmax), np.max(ymax), np.mean(score)]
+    return np.asarray(enclosing)
```
fg-mindee, line 57: return np.array([xmin.min(), ymin.min(), xmax.max(), ymax.max(), score.mean()])

doctr/utils/geometry.py
```diff
@@ -44,3 +45,13 @@ def resolve_enclosing_bbox(bboxes: List[BoundingBox]) -> BoundingBox:
 def resolve_enclosing_rbbox(rbboxes: List[RotatedBbox]) -> RotatedBbox:
     pts = np.asarray([pt for rbbox in rbboxes for pt in rbbox_to_polygon(rbbox)], np.float32)
     return fit_rbbox(pts)
+
+
+def resolve_enclosing_bboxarray(bboxarray: np.ndarray) -> np.ndarray:
```
fg-mindee, line 50: I suggest naming it `enclose_boxes` :) 

I was wondering how it differs from `resolve_enclosing_bbox`

fg-mindee (approved):
Nice, thanks!

For further reference, since this changes the default behaviour of the detection part, should we update the benchmark tables as well?

doctr/utils/geometry.py
charlesmindee: it has the same purpose but the computation is from an array of boxes instead of a list of `BoundingBox`

doctr/utils/geometry.py
```diff
@@ -44,3 +45,12 @@ def resolve_enclosing_bbox(bboxes: List[BoundingBox]) -> BoundingBox:
...
+    """Compute enclosing bbox from an array of boxes: (*, 5), where boxes have this shape:
+    (xmin, ymin, xmax, ymax, score).
+    Return a (1, 5) array (enclosing boxarray)
+    """
+    xmin, ymin, xmax, ymax, score = np.split(bboxarray, 5, axis=1)
+    return np.array([xmin.min(), ymin.min(), xmax.max(), ymax.max(), score.mean()])
```
fg-mindee, line 56: Actually, since both functions are doing the same, could you update the original function and only add support for numpy arrays?

something like that for the signature 
```python
def resolve_enclosing_bbox(bboxes: Union[List[BoundingBox], np.ndarray]) -> Union[BoundingBox, np.ndarray]:
```

That would make things simpler I guess

test/test_utils_geometry.py
```diff
@@ -30,3 +30,8 @@ def test_resolve_enclosing_rbbox():
     pred = geometry.resolve_enclosing_rbbox([(.2, .2, .05, .05, 0), (.2, .2, .2, .2, 0)])[:4]
     target = (.2, .2, .2, .2)
     assert all(abs(i - j) <= 1e-7 for (i, j) in zip(pred, target))
+
+
+def test_resolve_bboxarray():
+    pred = geometry.resolve_enclosing_bboxarray(np.array([[0.1, 0.1, 0.2, 0.2, 0.9], [0.15, 0.15, 0.2, 0.2, 0.8]]))
+    assert pred.all() == np.array([0.1, 0.1, 0.2, 0.2, 0.85]).all()
```
fg-mindee, line 37: If we merge both functions, some modifications will be required here

doctr/utils/geometry.py
fg-mindee: using `if isinstance(bboxes, np.ndarray):` to check the type

fg-mindee (commented):
Sorry, I had another thought of merging both bbox resolution function since they indeed have the same purpose. The rest looks good to me!

doctr/utils/geometry.py
charlesmindee: yes

fg-mindee (approved):
Thanks a lot!

Issue #331, opened 2021-06-28
MASTER training


fracav:
Hello  and good morning,

I'm trying your fantastic repo to try different OCR architectures, I have successfully managed to train both the crnn model (in its various versions) and also the SAR model. I am trying to train the MASTER model, with no great results. I get all the time Validation loss: + -3.72135 (Exact: 0.00% | Partial: 0.00%)

I am using a multilingual dataset with about 250k training images and with the other architectures I managed to get to Exact about 85%. (with this specific alphabet  ('3K}7eé;5àÎYho]QwV6qU~W"XnbBvcADfËmy.9ÔpÛ{CôïE%M4#ÈR:g@T$x?0î£|za1ù8,OG€P-''kçHëÀÂ2É/ûIJ\'j(LNÙFut[)èZs+&°Sd=Ï!<â_Ç>r i`l'))  

I am using the doctr/reference/recognition/train.py script. Any suggestions?

charlesmindee:
Hi @fracav, 

Thank you for your report & the attention you pay to doctr.
Concerning the MASTER model, the implementation is very recent and we are training the model for the first time this week. I already spotted a mistake in the encoding process of sequences, and there may be more of them since it is not very stable yet. We will include support for Pytorch within a few days, and finish debugging/training the architecture this week, I invite you to wait 1 week or 2 to use this model if you want work on a definitely stable version!

fracav:
Hi @charlesmindee,

thanks for your very fast response! Hope to try the new architecture soon and good luck with debugging :).
Have a good day. See you soon.

charlesmindee:
This should be fixed now, since #335.
Can you confirm @fracav ?

fracav:
dear @charlesmindee i am glad you solved it quickly. 
Tomorrow I plan to try, I'll let you know soon.

charlesmindee:
The model converged with a batch size of 64 and a learning rate of 1e-4 on my GPU, in the paper they used a learning rate of 1e-4 x number of GPUs, and I believe it is important to respect that: I first tried with a learning rate of 4e-4 and it resulted in very poor results.

fracav:
Hi @charlesmindee  and good morning,

I'm running my experiments on Colab pro (Tesla P100) with input size (48, 160,3), french vocabulary with also whitespace, lr = 0.001 and batch_size = 64. I am using the doctr/references/recognition/train.py script. Over the weekend I was not able to get good results, the predictions are always empty or the repetition of a character max_length times.
[...]

PR #335, opened 2021-07-01
fix: bug in MASTER loss

charlesmindee:
MASTER model was not trained properly due to a misalignment in the loss function, this is now fixed. More precisely:

- Loss function is updated (right shift for gt sequences)
- `encode_sequences` is  improved to support not only `EOS` but also `SOS` and `PAD` symbols, which are mandatory to train MASTER (`SOS` to shift gt to the right, and `PAD` because we need to mask where `PAD` in the loss but nos `EOS`). It is important to note that default behavior of this function is NOT modified, when pad and sos arguments are left to `None`, the function still pad with `EOS` at the end of each sequence. When `PAD` is provided, sequences are concatenated with `EOS` before adding `PAD` to fill the matrix.
- Updated corresponding unitests.
- When gt are given in the call, but `training=False`, the model is now using the prediction decoder to compute the loss, instead of using gts to feed the decoder. This is a more accurate way to evaluate the model and to be sure that we are training the model 'properly': we can check at evaluation time if it is able to predict _from the images only_ the right labels.

Any feedback is welcome!

### DIFF
--- a/doctr/models/recognition/master/tensorflow.py
+++ b/doctr/models/recognition/master/tensorflow.py
@@ -293,17 +293,17 @@ class MASTER(_MASTER, Model):
             gt, seq_len = self.compute_target(target)
 
         if kwargs.get('training', False):
-            # When not training, we want to compute logits in with the decoder, although
-            # we have access to gts (we need gts to compute the loss, but not in the decoder)
-            _, logits = self.decode(encoded, **kwargs)
-
-        else:
             if target is None:
                 raise AssertionError("In training mode, you need to pass a value to 'target'")
             tgt_mask = self.make_mask(gt)
             # Compute logits
             output = self.decoder(gt, encoded, tgt_mask, None, **kwargs)
             logits = self.linear(output, **kwargs)
+
+        else:
+            # When not training, we want to compute logits in with the decoder, although
+            # we have access to gts (we need gts to compute the loss, but not in the decoder)
+            _, logits = self.decode(encoded, **kwargs)
 
         if target is not None:
             out['loss'] = self.compute_loss(logits, gt, seq_len)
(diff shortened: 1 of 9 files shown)


fg-mindee:
For reference, #331 is related to this PR

doctr/models/recognition/master/tensorflow.py
```diff
@@ -298,8 +303,13 @@ def call(
             out['loss'] = self.compute_loss(logits, gt, seq_len)
 
         else:
+            # When not training, we want to compute logits in with the decoder, although
+            # we have access to gts (we need gts to compute the loss, but not in the decoder)
             _, logits = self.decode(encoded)
 
+            if target is not None:
+                out['loss'] = self.compute_loss(logits, gt, seq_len)
+
```
fg-mindee, line 312: This needs to be unindented ;)

fg-mindee: And you can remove line 302-303 to avoid duplicate loss computation

doctr/models/recognition/master/tensorflow.py
```diff
@@ -290,6 +292,9 @@ def call(
         if target is not None:
             # Compute target: tensor of gts and sequence lengths
             gt, seq_len = self.compute_target(target)
+
+        if training:
+            assert target is not None
```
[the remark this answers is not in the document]
fg-mindee: Actually start with the shortest option:
```python
if target is None:
```
those two conditional branches need to only compute the logits

fg-mindee (changes requested):
I think there is a small mistake in the implementation and then we can merge!

fg-mindee (approved):
Looks good to me!

PR #339, opened 2021-07-02
fix: invert_colorize

charlesmindee:
This PR modify the` invert_colorize` test in Pytorch to make sure the test won't randomly crash when running CI.

Any feedback is welcome!


test/pytorch/test_transforms.py
```diff
@@ -59,5 +59,5 @@ def test_invert_colorize(rgb_min):
 
     input_t = torch.full((8, 3, 32, 32), 255, dtype=torch.uint8)
     out = transfo(input_t)
-    assert torch.all(out <= int(math.ceil(255 * (1 - rgb_min))))
+    assert torch.all(out <= int(1 + math.ceil(255 * (1 - rgb_min))))
```
fg-mindee, line 62: Mmmh, but that's way less accurate, don't you think we should add something like `1e-4`?

fg-mindee (commented):
Thanks for the PR! Small suggestion in the comment

test/pytorch/test_transforms.py
charlesmindee: This is still an issue with 1e-4, because we have a gap of 1 between max(out) and `int(math.ceil(255 * (1 - rgb_min)))`

fg-mindee: Mmmh, but if it's not a rounding error, I'm not sure it's supposed to happen :thinking: 
Any idea why?

charlesmindee: I happens only with non-normalized images, I fixed it the same way we did for normalized images

fg-mindee (approved):
Nice, thanks!

PR #340, opened 2021-07-02
feat: add loss + MASTERPostProcessor in Pytorch

charlesmindee:
This PR adds `compute_loss` and `MASTERPostProcessor` in Pytorch, updated unitests and forward method as well.
Any feedback is welcome!


doctr/models/recognition/master/pytorch.py
```diff
@@ -14,9 +14,9 @@
 from ...utils import conv_sequence_pt, load_pretrained_params
 from ...backbones import resnet_stage
 from ..transformer import Decoder, positional_encoding
-from .base import _MASTER
+from .base import _MASTER, _MASTERPostProcessor
 
-__all__ = ['MASTER', 'master']
+__all__ = ['MASTER', 'MASTERPostProcessor', 'master']
```
fg-mindee, line 19: No need to export `MASTERPostProcessor` since it's wrapped by the model

doctr/models/recognition/master/pytorch.py
```diff
@@ -239,21 +272,33 @@ def forward(
...
+            if target is None:
+                raise AssertionError("In training mode, you need to pass a value to 'target'")
+            tgt_mask = self.make_mask(gt)
             # Compute logits
-            output = self.decoder(torch.from_numpy(gt), encoded, tgt_mask, None)
+            output = self.decoder(gt, encoded, tgt_mask, None)
             logits = self.linear(output)
 
         else:
             _, logits = self.decode(encoded)
```
fg-mindee, line 287: If we don't need the `ys`, perhaps we should avoid computing them in `self.decode`?

doctr/models/recognition/master/pytorch.py
```diff
@@ -204,10 +206,41 @@ def __init__(
...
+        Returns:
+            The loss of the model on the batch
+        """
+        # Input length : number of timesteps
+        input_len = model_output.shape[1]
+        # Add one for additional <eos> token (sos disappear in shift!)
+        seq_len = seq_len + 1
+        # Compute loss: don't forget to shift gt! Otherwise the model learns to output the gt[t-1]!
+        # The "masked" first gt char is <sos>. Delete last logit of the model output.
+        cce = F.cross_entropy(model_output.permute(0, 2, 1)[:, :, :-1], gt[:, 1:], reduction='none')
```
fg-mindee, line 236: Two things:
- for the slicing, you can use ellipsis if this is a 3-D tensor: `input_t[:, :, :-1]` is the same as `input_t[..., :-1]` (basically the ellipsis will fill the remaining dimensions)
- better to slice before the permutation

fg-mindee (commented):
Thanks for the PR! Some modifications suggestions in the comments

fg-mindee:
For reference, this is also related to #331

doctr/models/recognition/master/pytorch.py
charlesmindee: we need to iterate over ys in decode, it is a kind of buffer

fg-mindee (approved):
Thanks for the edits!

PR #350, opened 2021-07-05
feat: add bottleneck ratio in MASTER

charlesmindee:
This PR adds the bottleneck ratio of 1/16 in both tf & pt MASTER models, and set the number of multi aspect context to 8 as described in paper. 
Any feedback is welcome!


fg-mindee (approved):
Thanks!

PR #355, opened 2021-07-06
feat: Added box coord clipping and target validation for detection tasks

fg-mindee:
- added box coord clipping to avoid rounding errors in detection datasets
- added target validation for detection models to avoid silent failures
- added unittests for target validation

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #356, opened 2021-07-06
refactor: Renamed MASTER feature extractor attribute

fg-mindee:
This PR simply renames the `feature_extractor` attribute to `feat_extractor` to keep the same naming conventions across models.

Any feedback is welcome!


charlesmindee (approved):
thanks!

PR #357, opened 2021-07-06
fix: Fixed Resize transfo in PyTorch when aspect ratio matches the target

fg-mindee:
In rare cases, when the original image has the same aspect ratio as the target size, the resize was not performed correctly. This PR fixes that and adds a test case in corresponding pytorch unittests.

Any feedback is welcome!


charlesmindee (approved):
Thanks

Issue #352, opened 2021-07-06
[transforms] Add geometric transformations (train detection models with rotated samples)


charlesmindee:
We need to train our detection models with rotated samples, because for now we support rotation but our models are not fitted to predict rotated boxes, which yields to very poor results.

Ismail-Maj:
Hello,
I don't know if you're open to outsider contribution, but I think there might be a bug in your function (I don't know enough about the flow of the program to tell), line 46 doctr/transforms/functional/pytorch.py.

It's related to this problem: https://www.pyimagesearch.com/2017/01/02/rotate-images-correctly-with-opencv-and-python/
I had the same problem when training my OCR on rotated images, some part of the text went out of bound.

If it is in fact a problem, you could replace rotate from torchvision to imutils.rotate_bound, it's not ideal (extra import, tensor to array) so I think the best way to do it is to calculate the needed padding, pad the image then rotate with torchvision.

fg-mindee:
Hi @Ismail-Maj :wave: 

Thanks for reporting this! I'll take a look at it indeed, fortunately for now we haven't used rotation in our training (only as a preprocessing step in some cases) so there aren't any major consequences on existing checkpoints.

But this could get messy later on so thanks for this :pray: 
We certainly do accept outsider contributors, this is an open-source project, feel free to open issues/PRs :)

charlesmindee:
![Screenshot_2021-09-16 app · Streamlit(1)](https://user-images.githubusercontent.com/70526046/133616968-3d0b20d8-aa64-4b92-a021-d5f182cd02b7.png)
![Screenshot_2021-09-16 app · Streamlit](https://user-images.githubusercontent.com/70526046/133616985-35a965c1-c153-495a-af18-fd73b95b5da4.png)

fg-mindee:
Closed by #765 :rocket:

PR #358, opened 2021-07-06
feat: add rotate in transforms

charlesmindee:
This PR adds a rotate functional transformation in both pt and tf transforms modules.
The function `rotate_boxes` is moved to `utils/geometry.py`, since it is no longer useful only for models (otherwise we need to deal with circular imports).
Updated unitests as well.

Any feedback is welcome!

linked to #352

### DIFF
--- a/doctr/models/detection/differentiable_binarization/base.py
+++ b/doctr/models/detection/differentiable_binarization/base.py
@@ -271,7 +271,7 @@ class _DBNet:
             abs_boxes[:, [1, 3]] *= output_shape[-2]
             abs_boxes = abs_boxes.round().astype(np.int32)
 
-            if self.rotated_bbox:
+            if self.rotated_bbox or abs_boxes.shape[1] == 5:
                 boxes_size = np.minimum(abs_boxes[:, 2], abs_boxes[:, 3])
                 polys = np.stack([
                     rbbox_to_polygon(tuple(rbbox)) for rbbox in abs_boxes  # type: ignore[arg-type]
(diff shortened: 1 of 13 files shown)


doctr/models/detection/differentiable_binarization/base.py
```diff
@@ -271,7 +271,7 @@ def compute_target(
             abs_boxes[:, [1, 3]] *= output_shape[-2]
             abs_boxes = abs_boxes.round().astype(np.int32)
 
-            if self.rotated_bbox:
+            if self.rotated_bbox or abs_boxes.shape[1] == 5:
```
fg-mindee, line 274: If we do this switch, I suggest removing `self.rotated_bbox`

doctr/models/detection/linknet/base.py
```diff
@@ -128,7 +128,7 @@ def compute_target(
             abs_boxes[:, [1, 3]] *= output_shape[-2]
             abs_boxes = abs_boxes.round().astype(np.int32)
 
-            if self.rotated_bbox:
+            if self.rotated_bbox or abs_boxes.shape[1] == 5:
```
fg-mindee, line 131: same here

setup.py
```diff
@@ -84,6 +85,7 @@ def deps_list(*pkgs):
 extras = {}
 extras["tf"] = deps_list("tensorflow")
 extras["tf-cpu"] = deps_list("tensorflow-cpu")
+extras["tfa"] = deps_list("tensorflow-addons")
```
fg-mindee, line 88:
```python
extras["tf"] = deps_list("tensorflow", "tensorflow-addons")
extras["tf-cpu"] = deps_list("tensorflow-cpu", "tensorflow-addons")
```

fg-mindee (changes requested):
Thanks for the PR! I added a few comments

fg-mindee (approved):
Thanks for the edits!

fg-mindee:
@charlesmindee I didn't see it but the mypy job failed :sweat_smile: 
Could you add the corresponding entry to the missing import ignore in `mypy.ini` please? :pray:

PR #360, opened 2021-07-06
feat: add learning rate scheduler in TF training scripts

charlesmindee:
This PR adds a CosineDecay lr scheduler in both tensorflow training scripts.
Any feedback is welcome!


fg-mindee (approved):
Nice, thanks for this improvement!

PR #361, opened 2021-07-06
fix: Fixed recognition model PyTorch versions and added GPU compatibility

fg-mindee:
- fixed MASTER input_shape (causes failure for predictor)
- fixed multiple buffer registration (otherwise changing device doesn't propagate to this tensor)
- fixed LSTM dimensioning in CRNN (needed to be dynamic with backbone output shape)
- added `crnn_resnet31` which was missing from PyTorch support
- removed `include_top` constructor argument from backbones
- fixed compatibility with GPU input
- re-enabled GPU for unittests (when CUDA available)
- added missing tests for recognition zoo

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #366, opened 2021-07-07
feat: add crop functional transformation

charlesmindee:
This PR adds a `crop_detection` function in both tf & pt `transforms.functional`. This function crops a tf or pt image according a given crop_box, and crops as well the given array of bboxes (relative or absolutes).
Any feedback is welcome!


fg-mindee (approved):
Thanks a lot!

Issue #263, opened 2021-05-17
[references] Add FP16 support for training


fg-mindee:
The reference training script should have an option to switch from a FP32 training to a FP16 training.
It would raise a question: how to harmonize model loading from both FP?

A few suggestions:
- simply cast the FP16 checkpoint to FP32 before hashing + uploading
- for a given model, each checkpoint is marked with its FP. The user may choose how the model needs to be instantiated: by default the output FP is FP32, but this can be changed by the user. Depending on the checkpoint FP, it will be cast to its counterpart FP for loading.

The second option would be cleaner and avoid unnecessary large checkpoints when they can be kept in FP16

Here is a proposition:
- [x] add support of FP16 to `doctr.datasets` (#367)
- [x] add support of FP16 to `doctr.models` (#382)
- [x] add support of FP16 to `doctr.transforms` (#388)
- [x] add support of FP16 to `references` with PyTorch (#604)
- [ ] add support of FP16 to `references` with TensorFlow

fg-mindee:
Quick update: this design was good on paper but ML frameworks actually uses AMP instead of full fp16. And other data types yield issues with post processing (cv2 doesn't like things that are not in fp32 among others)

[ ] - Adding AMP support to PyTorch trainings (#604)
[ ] - Adding AMP support to TensorFlow trainings

fg-mindee:
Moving this to a later release for TensorFlow support of AMP

PR #367, opened 2021-07-07
feat: Added FP16 support to doctr.datasets

fg-mindee:
Following up on #263, this PR introduces the following modifications:
- switched default image dtype of TF datasets to FP32 (before it was `tf.uint8`)
- added FP16 mode on all datasets
- updated references script to reflect changes (no need to perform 255 division manually anymore)

Please note that there is a breaking change here, the default image dtype of datasets in TF has been switched from `tf.uint8` (between 0 and 255) to `tf.float32` (between 0 and 1).

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #368, opened 2021-07-07
fix: Removed PDF forced square resize from demo & analysis script

fg-mindee:
This PR removes the resizing step of PDF image rendering to square images. The reason behind that is that while it's indeed optimized if the predictor target size is that specific size if there are multiple preprocessors inside, it yields double interpolation.

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #369, opened 2021-07-07
feat: Added page selection & JSON display to demo

### LINKED ISSUE #234: [demo] Improve UI for OCR result display
For very dense documents, since the predicted text value is plotted statically, there can be some readability issues. We should try to improve this

fg-mindee:
- added tip to enlarge images
- switched to 4 columns display
- added page selection (for multi-page PDF)
- speeded up analysis by only analyzing the selected page
- added JSON output display (foldable, and can be copied to clipboard)
- updated demo display in README

Here is how it renders:
Demo

Closes #234

Any feedback is welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ You can then easily run your app in your default browser by running:
 streamlit run demo/app.py
 ```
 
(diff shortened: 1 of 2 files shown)


charlesmindee (approved):
Thanks for the tuning

PR #370, opened 2021-07-07
fix: Fixed preprocessor support to torch.uint8, tf.uint8 and np.float32

fg-mindee:
- added a much more comprehensive strategy for input preprocessing that now correctly processes tensors in `uint8` or numpy in `float32`
- also added multithreading for sample processing 

Any feedback is welcome!


charlesmindee (dismissed):
Thanks

charlesmindee (approved):
Thanks for that!

PR #375, opened 2021-07-08
fix: Removed extra 255 division

fg-mindee:
This PR fixes a bug introduced in #367:
- `tf.image.convert_image_dtype` already performs the 255 division when it receives a float dtype.
- the extra 255 division was wrongly placed afterward and is removed in this PR

Any feedback is welcome!


charlesmindee (approved):
thanks

PR #376, opened 2021-07-08
fix: Recognition dataset merging

charlesmindee:
This PR allows recognition datasets to be merged.
Any feedback is welcome!

### DIFF
--- a/doctr/datasets/recognition.py
+++ b/doctr/datasets/recognition.py
@@ -6,6 +6,7 @@
 import os
 import json
 from typing import Tuple, List, Optional, Callable, Any
+from pathlib import Path
 
 from .datasets import AbstractDataset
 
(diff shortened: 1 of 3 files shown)


doctr/datasets/recognition.py
```diff
@@ -46,3 +47,19 @@ def __init__(
             if not isinstance(label, str):
                 raise KeyError("Image is not in referenced in label file")
             self.data.append((img_path, label))
+
+    def merge_dataset(self, ds: AbstractDataset) -> None:
+        # Update data with new root for self
+        new_data: List[Tuple[str, str]] = []
+        for img_path, label in self.data:
+            new_data.append((str(Path(self.root).joinpath(img_path)), label))
+        self.data = new_data
```
fg-mindee, line 56:
```python
self.data = [(str(Path(self.root).joinpath(img_path)), label) for img,path, label in self.data]
```

list comprehension for the win

doctr/datasets/recognition.py
```diff
@@ -46,3 +47,19 @@ def __init__(
...
+        # Define new root
+        self.root = Path("/")
+        # Merge with ds data
+        for img_path, label in ds.data:
+            self.data.append((str(Path(ds.root).joinpath(img_path)), label))
+
+    def merge_datasets(self, ds_list: List[AbstractDataset]) -> None:
+        for ds in ds_list:
+            self.merge_dataset(ds)
```
fg-mindee, line 65: I'm having second thought about whether this method is useful (without it, we only ave one extra line with the "for" loop to add)

fg-mindee (commented):
Thanks for the PR! 2 comments where I'd like to have your opinion

doctr/datasets/recognition.py
charlesmindee: I agree, as you wish

fg-mindee (approved):
Thanks!

PR #378, opened 2021-07-12
fix: transform test bug in CI

charlesmindee:
This PR fixes the rotation in the transform module.
(When `min_angle` > 0, if the angle was less than `min_angle` boxes were not rotated and returned in a straight way (4 coords), now boxes are always composed of 5 coords, even if the angle is 0.)


fg-mindee (approved):
Thanks!

PR #379, opened 2021-07-12
fix: Torch loss dimension

charlesmindee:
This PR fixes both SAR & MASTER loss in pytorch which were composed of more than 1 scalar, which was triggering an error when training:
`RuntimeError: grad can be implicitly created only for scalar outputs`

### DIFF
--- a/doctr/models/recognition/master/pytorch.py
+++ b/doctr/models/recognition/master/pytorch.py
@@ -245,7 +245,7 @@ class MASTER(_MASTER, nn.Module):
         cce[mask_2d] = 0
 
         ce_loss = cce.sum(1) / seq_len.to(dtype=torch.float32)
-        return torch.mean(ce_loss)
+        return ce_loss.mean()
 
     def forward(
         self,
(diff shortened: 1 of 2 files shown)


doctr/models/recognition/master/pytorch.py
```diff
@@ -245,7 +245,7 @@ def compute_loss(
         cce[mask_2d] = 0
 
         ce_loss = cce.sum(1) / seq_len.to(dtype=torch.float32)
-        return ce_loss.unsqueeze(1)
+        return torch.mean(ce_loss)
```
fg-mindee, line 248: `return ce_loss.mean()`
so that we don't necessarily need to import torch for this line :+1:

doctr/models/recognition/sar/pytorch.py
```diff
@@ -242,7 +242,7 @@ def compute_loss(
         cce[mask_2d] = 0
 
         ce_loss = cce.sum(1) / seq_len.to(dtype=torch.float32)
-        return ce_loss.unsqueeze(1)
+        return torch.mean(ce_loss)
```
fg-mindee, line 245: `return ce_loss.mean()` for the same reasons

fg-mindee (commented):
Thanks for spotting this! I made a small adjustment request

fg-mindee (approved):
Thanks for the edit!

PR #380, opened 2021-07-13
fix: Removed deprecated arg in training script

charlesmindee:
This PR removes a deprecated arg in the training script.


fg-mindee (approved):
Cheers!

PR #382, opened 2021-07-13
feat: Added FP16 support for doctr.models

fg-mindee:
Following up on #263, this PR introduces the following modifications:
- made dtype dynamic whenever possible
- added error message for unknown dtype in preprocessors
- added FP16 support to all models
- updated unittests

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #384, opened 2021-07-18
feat: add a draw_boxes function for detection task only

### LINKED ISSUE #89: [utils] Add visualization capabilities for independent tasks
Visualization is end-to-end for the moment dynamic, but this means that a static version is not currently available, nor that there is a visualization option for text detection or text recognition only. We should discuss and add visualization for the following blocks:
- [ ] Text detection: display bounding boxes of detected items over the image
- [ ] Text recognition: display the label and confidence in a corner of the crop

charlesmindee:
This PR adds a draw_boxes function to draw boxes on images when performing detection task.
Partially resolves #89 
Any feedback is welcome!


doctr/utils/visualization.py
```diff
@@ -229,3 +229,24 @@ def synthetize_page(
...
+        image: np array
+    """
+    h, w = image.shape[:2]
+    for box in boxes.tolist():
+        xmin, ymin, xmax, ymax = box
+        image = cv2.rectangle(
+            image,
+            (int(xmin * w), int(ymin * h)),
+            (int(xmax * w), int(ymax * h)),
+            color=(0, 0, 255),
```
fg-mindee, line 248: Let's make the color a default argument (None by default) of the function:

`color=color if isinstance(color, tuple) else (0, 0, 255)`

doctr/utils/visualization.py
```diff
@@ -229,3 +229,24 @@ def synthetize_page(
...
+def draw_boxes(boxes: np.ndarray, image: np.ndarray, **kwargs) -> None:
+    """Draw an array of relative straight boxes on an image
+
+    Args:
+        boxes: array of relative boxes, of shape (*, 4)
+        image: np array
+    """
+    h, w = image.shape[:2]
+    for box in boxes.tolist():
+        xmin, ymin, xmax, ymax = box
```
fg-mindee, line 243: the conversion to absolute coords could be vectorized by doing this for all boxes at once

test/common/test_utils_visualization.py
```diff
@@ -15,3 +15,14 @@ def test_draw_page():
...
+
+def test_draw_boxes():
+    image = np.ones((256, 256, 3))
+    boxes = [
+        [0.1, 0.1, 0.2, 0.2],
+        [0.15, 0.15, 0.19, 0.2],  # to suppress
+        [0.5, 0.5, 0.6, 0.55],
+        [0.55, 0.5, 0.7, 0.55],  # to suppress
+    ]
+    visualization.draw_boxes(boxes=np.array(boxes), image=image, block=True)
```
fg-mindee, line 28: How can this pass? The `block=True` makes this test a blocking one :thinking:

doctr/utils/visualization.py
```diff
@@ -229,3 +229,24 @@ def synthetize_page(
...
 
     return response
+
+
+def draw_boxes(boxes: np.ndarray, image: np.ndarray, **kwargs) -> None:
+    """Draw an array of relative straight boxes on an image
+
+    Args:
+        boxes: array of relative boxes, of shape (*, 4)
+        image: np array
```
fg-mindee, line 239: can you specify if there is a required dtype? (np.uint8 or np.float32 ?)

fg-mindee (changes requested):
Thanks for the PR! I added some comments

doctr/utils/visualization.py
```diff
@@ -294,3 +294,33 @@ def synthetize_page(
...
+    """Draw an array of relative straight boxes on an image
+
+    Args:
+        boxes: array of relative boxes, of shape (*, 4)
+        image: np array, float32 or uint8
+    """
+    h, w = image.shape[:2]
+    # Convert boxes to absolute coords
+    boxes[:, [0, 2]] *= w
+    boxes[:, [1, 3]] *= h
```
fg-mindee, line 314: I think we should do a copy of boxes before doing inplace ops :sweat_smile:

fg-mindee (commented):
Thanks! Just a small thing about an inplace ops

fg-mindee (approved):
Thanks!

PR #397, opened 2021-08-05
fix: LR scheduler's final LR in TF scripts

charlesmindee:
This PR updates the lr scheduler in tf scripts to have the same training parameters in both tensorflow and pytorch scripts. The changes were already made locally on the training machine before launching trainings, nothing to worry about.


fg-mindee (approved):
Thanks!

PR #403, opened 2021-08-09
feat: add export signatures for saved_models

charlesmindee:
This PR defines export signatures for serving models.
Any feedback is welcome


fg-mindee:
Not entirely sure, but either we:
- transform this PR to make it flexible enough to integrate it
- close the PR

What do you think @charlesmindee? :pray:

PR #407, opened 2021-08-10
feat: Added support of new label format in datasets

fg-mindee:
- adds a root folder check for all datasets
- adds support of new label format for private datasets (including US tax forms)

Any feedback is welcome!


charlesmindee (approved):
Thanks!

Issue #255, opened 2021-05-12
[models] Add detection & recognition models with much lighter backbones


fg-mindee:
Comparing similar vision tasks, the segmentation part of text detection is easier than real-life scene segmentation (on Cityscapes for instance) which uses ResNet-like backbones, and even easier than medical segmentation which uses often lighter feature extractor (cf. UNet).

Additionally, the feature extraction of text recognition is similar to image classification for a single character image (MNIST but with more classes) which uses very light backbones.

So here is a suggestion:
- [ ] train a very light image classifier for character classification on a specific vocab (it will be used as a pretrained backbone for text recognition)
- [ ] train a text recognition model with this
- [ ] train a light text segmentation model

This means that DocTR will extend the list of supported tasks by adding:
- character classification
- text segmentation

fg-mindee:
Closed by #516 & #517

PR #408, opened 2021-08-11
feat: add MNIST-like characters dataset generator

charlesmindee:
This PR implements a generate_character function to generate random single character images from a vocabulary to train our recognition backbones.

Linked to #255 

Any feedback is welcome!

Samples:
mnist


fg-mindee:
Also we need some unittests, and mypy isn't happy :sweat_smile:

fg-mindee (commented):
A few improvement suggestions in comments!

fg-mindee:
Also quick suggestion: let's produce straight characters (we'll be able to use transforms afterwards to rotate the characters if needed)

PR #420, opened 2021-08-19
fix: tf mobilenet classification head + loss classifier

charlesmindee:
This PR fixes the tf mobilenet classification head (removed the last softmax included in the loss) and fixes the loss function: tf.nn.softmax_ce is designed for one-hot labels (to allow a non binary distribution for the labels across the classes) whereas tf.nn.sparse_softmax_ce is designed for a unique class for each entry, embedded by the index in the label tensor.

Any feedback is welcome!


fg-mindee (approved):
Well spotted, thanks!

Issue #422, opened 2021-08-19
Pytorch models not implemented


fmobrj:
Hello all!

Thanks, for the great library!

I am trying to test the pytorch code persented here: [,](https://github.com/mindee/doctr/releases) , but I get a NotImplementedError message. Inspecting the code for the function _dbnet, it seems that the code is raising the error:

`    if pretrained:
        raise NotImplementedError

    return model` 

So, actually, are there pretrained models for use with pytorch?

Best regards,
Fabio.

charlesmindee:
Hi @fmobrj,

Thanks for your feedback, as a matter of facts we have implemented all our models in Pytorch but we are behind schedule for the trainings: pytorch trainings are still in progress or to be scheduled and we cannot provide pretrained models yet. Sorry for that, we are working on it and within a few weeks it should be OK.

However you can use the pytorch architectures in the meantime and train the models with your own datasets, it should work fine.

Have a nice day!

fg-mindee:
Hi @fmobrj,

To give you extra details, I see two different aspects in this issue:
- [ ] having a more specific user message when pretrained checkpoints are not available (should be tackled by #425)
- [ ] catching up with TF checkpoints

The second aspect will be one of the focus of the `0.4.0` release (and thus the coming weeks)!

PR #425, opened 2021-08-23
fix: Fixed PyTorch param loading

fg-mindee:
Following up on #422, this PR switches back to the common parameter loading mechanism. 

Previously, some pytorch models' factory functions were throwing errors when `pretrained=True`. With this PR, it throws a warning and proceeds with param initialization when no checkpoint is available.

Any feedback is welcome!


charlesmindee (approved):
Thanks

charlesmindee (approved):
Thanks!

PR #427, opened 2021-08-25
fix: Fixed GPU selection in PyTorch training scripts

### LINKED ISSUE #409: [training] Pytorch training of recognition models is very slow
Pytorch trainings are very slow for recognition models, far more slower than tensorflow trainings with the same dataset.
It could either be:
- A torch dataloading issue (dataset/dataloader)
- A torch model issue
- A hardware issue (CUDA/GPU/CPU lack of compatibility)

fg-mindee:
- added GPU selection by default if available in PyTorch scripts
- added console warning when running on CPU
- fixed missing GPU support on text recognition training (considerable speedup)

Resolves #409

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #432, opened 2021-08-25
fix: deprecated clipping + removed 'None' metrics

charlesmindee:
This PR fixes: 

- a deprecation warning in both pytorch training scripts with gradient clipping function.
- a bug occurring when no boxes at all are found (edge case): the current LocalizationMetric gives back a None for both precision & IoU and this is blowing up the training script because metrics can't be logged properly.

Any feedback is welcome!

### DIFF
--- a/references/detection/train_pytorch.py
+++ b/references/detection/train_pytorch.py
@@ -223,12 +223,11 @@ def main(args):
             print(f"Validation loss decreased {min_loss:.6} --> {val_loss:.6}: saving state...")
             torch.save(model.state_dict(), f"./{exp_name}.pt")
             min_loss = val_loss
-        if not all(recall, precision, mean_iou):
-            mb.write(f"Epoch {epoch + 1}/{args.epochs} - Validation loss: {val_loss:.6} "
-                     "Either recall, precision or mean_iou can't be defined (empty gts or empty preds)")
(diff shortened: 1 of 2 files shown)


doctr/utils/metrics.py
```diff
@@ -344,13 +344,13 @@ def summary(self) -> Tuple[Optional[float], Optional[float], Optional[float]]:
...
-        recall = self.matches / self.num_gts if self.num_gts > 0 else None
+        recall = self.matches / self.num_gts if self.num_gts > 0 else 0
 
         # Precision
-        precision = self.matches / self.num_preds if self.num_preds > 0 else None
+        precision = self.matches / self.num_preds if self.num_preds > 0 else 0
 
         # mean IoU
-        mean_iou = self.tot_iou / self.num_preds if self.num_preds > 0 else None
+        mean_iou = self.tot_iou / self.num_preds if self.num_preds > 0 else 0
```
fg-mindee, line 353: I understand that keeping it this way would crash the script, but perhaps logging a `0` isn't a good idea either?
There is a difference between having a metric being undefined and being 0 (I'm afraid we might miss something later on)

fg-mindee (commented):
Thanks for the PR!
Agree with the inplace clipping, but I added a comment for the metric

Issue #442, opened 2021-08-27
[transforms] Add an option to preserve aspect ratio during rotation when expand=True


fg-mindee:
With #438, we can now avoid losing image information during rotation. However, this doesn't preserve the initial aspect ratio by default.

This snippet:
```python
import tensorflow as tf
import numpy as np
from doctr.transforms.functional import rotate
import matplotlib.pyplot as plt

img = tf.ones((256, 128, 3), dtype=tf.float32)
box = [95, 40, 105, 60]
res = rotate(img, np.array([box], dtype=int), angle=30, expand=True)
_, axes = plt.subplots(1, 2)
# Mark box center
img = img.numpy()
img[int((box[3] + box[1]) / 2), int((box[2] + box[0]) / 2), 1] = 0
axes[0].imshow(img)
# Mark box center
rbox = res[1][0]
res = res[0].numpy()
res[int(rbox[1]), int(rbox[0]), 1] = 0
axes[1].imshow(res)
plt.show()
```

yields:
![aratio_issue](https://user-images.githubusercontent.com/76527547/131155165-ff54bccb-6457-4ef2-bfa7-b4aaece6c8ae.png)

fg-mindee:
Considering our earlier discussion about combining this transfo with `Resize` while `preserve_aspect_ratio=True`, if you agree @charlesmindee, we can close this :ok_hand:

fg-mindee:
Closing this @charlesmindee  as this looks resolved (cf. comment above)!
Feel free to reopen if I missed anything :ok_hand:

PR #448, opened 2021-08-30
feat: add RandomCrop module in transforms

charlesmindee:
This PR adds RandomCrop in `transforms/base.py`, and updates the detection dataset  with a `geometric_transforms` args to compose both RandomCrop and RandomRotate. 

Any feedback is welcome!

NB: cropping should be performed BEFORE rotation, because it is lighter that way (less information to deal with), and also because the rotation transformation converts boxes to rotated boxes whereas cropping is performed with straight bounding boxes.


doctr/transforms/functional/pytorch.py
```diff
@@ -32,7 +32,7 @@ def rotate(
     img: torch.Tensor,
     boxes: np.ndarray,
     angle: float,
-    expand: bool = False,
+    expand: bool = True,
```
fg-mindee, line 35: I would advise against it: I didn't enable it on purpose in #438.
This has consequences on predictors which we still need to investigate carefully as mentioned in #225

doctr/transforms/functional/tensorflow.py
```diff
@@ -33,7 +33,7 @@ def rotate(
     img: tf.Tensor,
     boxes: np.ndarray,
     angle: float,
-    expand: bool = False,
+    expand: bool = True,
```
fg-mindee, line 36: same here

doctr/transforms/modules/base.py
```diff
@@ -89,21 +89,45 @@ def __call__(self, img: Any) -> Any:
...
+        return f"min_wh={self.min_wh}, max_wh={self.max_wh}"
+
+    def __call__(self, img: Any, boxes: np.ndarray) -> Tuple[Any, np.ndarray]:
+        h, w = img.shape[:2]
+        crop_w = random.uniform(self.min_wh, self.max_wh)
+        crop_h = random.uniform(self.min_wh, self.max_wh)
+        start_x, start_y = random.uniform(0, 1 - crop_w), random.uniform(0, 1 - crop_h)
+        crop_box = (int(start_x * w), int(start_y * h), int((start_x + crop_w) * w), int((start_y + crop_h) * h))
+        croped_img, crop_boxes = F.crop_detection(img, boxes, crop_box)
+        return croped_img, crop_boxes
```
fg-mindee, line 133: A bare `RandomCrop` cannot be used as an augmentation to have consistent size inside a batch. I would suggest to go for `RandomResizedCrop` similarly to https://pytorch.org/vision/stable/transforms.html#torchvision.transforms.RandomResizedCrop

(variation to the scale & the aspect ratio before the resizing should be decorrelated in my opinion)

doctr/datasets/detection.py
```diff
@@ -33,11 +33,13 @@ def __init__(
         img_folder: str,
         label_folder: str,
         sample_transforms: Optional[Callable[[Any], Any]] = None,
+        geometric_transforms: Optional[Callable[[Any], Any]] = None,
```
fg-mindee, line 36: for now I suggest only implementation the augmentation without integrating in the datasets. I would argue that the integration should only have up to two types of transforms : sample transforms (before batching), and batch_transforms (after batching).

Currently, the latter being minimal, we haven't implemented that but we might come back to this in the future. Regarding geometric transforms, for the more general purpose of a transformation,  which isn't necessarily geometric, I think we should keep sample_transforms and do the following:
- change expected input/output types of sample_transforms to `img, target` instead of `img`
- for existing transforms applying only to the image, we had a wrapper to take `img, target` as input, but, under-the-hood, only applying a given transform to the img before returning the resulting pair

What do you think?

fg-mindee (changes requested):
Thanks for the PR!

I think we should keep this PR about the sole implementation of the cropping augmentation (and handle the integration of img+target transforms in the datasets in another PR), cf my comments :)

doctr/transforms/modules/base.py
```diff
@@ -89,21 +90,45 @@ def __call__(self, img: Any) -> Any:
...
-    def __init__(self, max_angle: float = 25., expand: bool = False) -> None:
+    def __init__(self, max_angle: float = 5., expand: bool = True) -> None:
         self.max_angle = max_angle
         self.expand = expand
 
     def extra_repr(self) -> str:
         return f"max_angle={self.max_angle}, expand={self.expand}"
 
-    def __call__(self, img: Any, target: Dict[str, np.ndarray]) -> Tuple[Any, Dict[str, np.ndarray]]:
+    def __call__(self, img: Any, boxes: np.ndarray) -> Tuple[Any, np.ndarray]:
```
fg-mindee, line 107: Mmmh we should keep target (imagine if we start including the segmentation target, we would also need to transform that part of the target)

doctr/transforms/modules/base.py
```diff
@@ -89,21 +90,45 @@ def __call__(self, img: Any) -> Any:
...
-    """Randomly rotate a tensor image
+    """Randomly rotate a tensor image and its boxes
 
     Args:
         max_angle: maximum angle for rotation, in degrees. Angles will be uniformly picked in
             [-max_angle, max_angle]
         expand: whether the image should be padded before the rotation
     """
-    def __init__(self, max_angle: float = 25., expand: bool = False) -> None:
+    def __init__(self, max_angle: float = 5., expand: bool = True) -> None:
```
fg-mindee, line 100: While #442 isn't dealt with, do you think we should switch to `expand=True` ?
- advantages: we don't lose information
- inconvenient: aspect ratio is messed up

Since the resizing would be done afterwards, it's possible. But we have to understand the consequences

fg-mindee (commented):
Thanks for the edits! Only a few comments left

fg-mindee (commented):
Thanks ! Looks good to me, could you add an entry in the documentation please? :pray:

fg-mindee (approved):
Cheers!

PR #453, opened 2021-08-31
fix: Torch detection preprocessing shape

charlesmindee:
This PR fixes the detection predictor preprocessing, the shape is now correct when using torch.


fg-mindee (approved):
Thanks!

Issue #429, opened 2021-08-25
[tests] Add CI jobs to check all scripts outside of the core codebase


fg-mindee:
As of now, the CI runs unittests, the analyze script. However there are a bunch of other things that would need testing:
- [x] package unittests
- [x] API template
- [x] browser demo (#456)
- [ ] reference training scripts _(#457, #469, #491)_
- scripts
    - [x] `analyze.py` (#315, #351, #458)
    - [x] `collect_env.py` (#456)
    - [x] `evaluate.py` (#456, #458)

PR #456, opened 2021-09-03
chore: Added CI jobs to check evaluate, collect_env & demo

fg-mindee:
Following up on #429, this PR introduces the following modifications:
- adds a job to run the `evaluate.py` script on FUNSD
- adds a job to check the `collect_env.py` script
- adds a job to check whether we can run the demo
- adds an option to run the evaluation script on the first N samples (the CI job checks it for the 10 first samples)

Any feedback is welcome!

### DIFF
--- a/.github/workflows/demo.yml
+++ b/.github/workflows/demo.yml
@@ -1,0 +1,49 @@
+name: demo
+
+on:
+  push:
+    branches: main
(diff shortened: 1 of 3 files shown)


charlesmindee (approved):
Thanks for that!

PR #458, opened 2021-09-03
fix: Fixed PyTorch tensor cropping and extended script support

### LINKED ISSUE #428: [scripts] Add a PyTorch analysis & evaluation script
This project has had an evaluation script and its analysis counterpart early on for TensorFlow only. However, with DocTR entering the multi-framework sphere, we need to accommodate users in the same fashion for all of them.

- PyTorch analysis script
- PyTorch evaluation script

fg-mindee:
As per #428, this PR introduces the following modifications:
- added option to correctly crop images that are channels_first
- updated predictors accordingly
- added support of PyTorch in the evaluation script
- added support of PyTorch in the analysis script
- added CI job to check Pytorch evaluation
- added metric formatter when a metric is undefined
- fixed predictor call when some crops have a size of 0

Resolves #428

Any feedback is welcome!

### DIFF
--- a/.github/workflows/scripts.yml
+++ b/.github/workflows/scripts.yml
@@ -14,6 +14,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
         python: [3.7, 3.8]
+        framework: [tensorflow, pytorch]
     steps:
       - if: matrix.os == 'macos-latest'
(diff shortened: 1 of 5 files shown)


charlesmindee (approved):
Thanks!

PR #461, opened 2021-09-09
fix: Fixed cropping mode in predictor

fg-mindee:
This PR fixes the cropping mode when both frameworks are available. #458 introduced an issue occurring for users with both frameworks installed. SInce the priority (default framework mode) is TensorFlow, this PR switches the condition to ensure proper cropping.

Any feedback is welcome!


charlesmindee (approved):
Thanks for the fix!

PR #462, opened 2021-09-10
docs: Fixed broken URL in README

fg-mindee:
This PR fixes a broken URL that was referring to the former `documents` module which has been renamed to `io`.


charlesmindee (approved):
Thanks for the fix!

PR #464, opened 2021-09-13
feat: Added portuguese entry in VOCAB

fmobrj:
PR for portuguese entry in Vocab.

### DIFF
--- a/doctr/datasets/vocabs.py
+++ b/doctr/datasets/vocabs.py
@@ -16,5 +16,5 @@ VOCABS: Dict[str, str] = {
     'currency': '£€¥¢฿',
     'latin': string.digits + string.ascii_letters + string.punctuation + '°',
     'french': string.digits + string.ascii_letters + string.punctuation + '°' + 'àâéèêëîïôùûçÀÂÉÈËÎÏÔÙÛÇ' + '£€¥¢฿',
-    'portuguese': string.digits + string.ascii_letters + string.punctuation + '°' + 'ª' + 'àâáãéêíïóôõúüçÀÂÃÁÉÊÍÏÔÓÕÚÜÇ' + '£€¥¢฿',
+    'portuguese': string.digits + string.ascii_letters + string.punctuation + '°' + 'àâáãéêíïóôõúüçÀÂÃÁÉÊÍÏÔÓÕÚÜÇ' + '£€¥¢฿',
 }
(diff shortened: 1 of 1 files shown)


doctr/datasets/vocabs.py
```diff
@@ -16,4 +16,5 @@
     'currency': '£€¥¢฿',
     'latin': string.digits + string.ascii_letters + string.punctuation + '°',
     'french': string.digits + string.ascii_letters + string.punctuation + '°' + 'àâéèêëîïôùûçÀÂÉÈËÎÏÔÙÛÇ' + '£€¥¢฿',
+    'portuguese': string.digits + string.ascii_letters + string.punctuation + '°' + 'ª' + 'àâáãéêíïóôõúüçÀÂÃÁÉÊÍÏÔÓÕÚÜÇ' + '£€¥¢฿',
```
fg-mindee, line 19: A few things:
- would you mind removing the ordinal mark `'ª'`? We'll try to handle this without
- I'm not a Portuguese native, but it looks like there are some missing accented characters: `èìòù`
- similarly, I checked and couldn't find the following characters being valid portuguese characters: `ïü`

(I used the Diacritics section of https://en.wikipedia.org/wiki/Portuguese_orthography)

fg-mindee (changes requested):
Thanks a lot for the PR :pray: 
I added one comment about the vocab, let me know what you think!

doctr/datasets/vocabs.py
fmobrj: No problem removing the 'ª'  symbol. It as as common as ° in portuguese, but I don't think it is crucial to have. When adding the portuguese symbols I was in doubt if I should include it or not, but included because of the °. In portuguese we don't have èìòù. Only à. 

Until very recently we used to have ïü. By the last "portuguese language agreement" the ¨ was abolished, but a lot of not so old texts will have ïü. What do you think? Should I remove it?

What should I do? Update the file and resubmit the PR?

fg-mindee: Regarding `'ª'` I fully understand, the only reason we have `'°'` in other alphabets is for it being the symbol of "degrees" (not to be confused with the ordinal `º`) ! And to be honest we think about removing it, since it's only an `o` in smaller and placed differently.

The rest looks good :+1: 
So you only have to edit your file locally (removing the ordinal), then
```
git commit -m "refactor: Edited portuguese VOCAB"
```
And finally push your commit (it will automatically update the PR):
```
git push origin portuguese-vocab
```

fmobrj: Done!

fg-mindee (approved):
Looks good to me thanks!
I'll fix the flake8 check later on :ok_hand:

PR #465, opened 2021-09-13
feat: Add crop splitting in recognition predictor for high aspect-ratio boxes

charlesmindee:
This PR adds a crop splitting feature in the recognition predictor to split too long boxes in small ones before feeding the recognition model. This should prevent the predictor to be fed with very long (and thus highly vertically padded) sequences, which was leading to very poor result. 

The splitting is activated by default when aspect ratio is > 8, which corresponds to 50% of the image padded (aspect ratio of 4 in the model).

The splitted crops are dilated so they overlap, thus it is easy to merge back full words with a high precision.
We propose a `merge_sequences` function in the` recognition.utils` module, which is smart enough to reconstruct a full word from its samples (computing the min. Levestein distance between the prefixes/suffixes). It takes a dilation factor as parameter, corresponding to the geometrical overlap between samples of one word, which is used only to determine the number of occurrences of a character when the word is splitted on a repetition of more than 2 characters. Thus it is robust to long repetitions, such as MRZ codes. The dilation factor is set to 1.4, which often corresponds to an overlap of 2 characters, depending on the font.

After checking, it does slow down the inference speed on CPU of less than 7% on documents with many boxes to split, and of less than 1% on documents with few boxes to split.

Any feedback is welcome!

Samples:

mrz


doctr/models/recognition/utils.py
```diff
@@ -0,0 +1,90 @@
...
+        'abcdefgh'
+        >>> merge_sequences('abcdi', 'cdefgh', 1.4)
+        'abcdefgh'
+    """
+    seq_len = min(len(a), len(b))
+    if seq_len == 0:  # One sequence is empty, return the other
+        if len(a) == 0:
+            return b
+        else:
+            return a
```
fg-mindee, line 37:
```
return b if len(a) == 0 else b
```

doctr/models/recognition/utils.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # Edge case (split in the middle of char repetitions): if it starts with 2 or more 0
+    if (scores[0], scores[1]) == (0, 0):
+        # Compute n_overlap (number of overlapping chars, geometrically determined)
+        n_overlap = round(len(b) * (dil_factor - 1) / dil_factor)
+        # Find the number of consecutive zeros in the scores list
+        n_zeros = 0
+        for i, score in enumerate(scores):
+            if score == 0:
+                n_zeros += 1  # Impossible to have a zero after a non-zero score in that case
```
fg-mindee, line 52:
```
n_zeros = sum(val == 0 for val in scores)
```

fg-mindee (commented):
Thanks for the PR, that's a really useful feature here!

I added a few comments :ok_hand:

fg-mindee:
Apart from if I misunderstood something, this PR doesn't introduce a breaking change. But we will have to update the benchmarks in the documentation if we merge it :)

charlesmindee:
I am finishing the benchmark this morning to open a new PR

doctr/models/recognition/core.py
```diff
@@ -105,4 +128,25 @@ def __call__(
             # Process outputs
             out = [charseq for batch in raw for charseq in batch]
 
+            # Find if crops were splitted, and if so merge back the predictions
+            if len(splitted_idxs):
```
fg-mindee, line 132: Let's be specific:
```
if len(splitted_idxs) > 0:
```

fg-mindee:
It looks like the new dependency is breaking a few things (the docker fails because building this specific deps requires `gcc`)
And the deps needs to be added to `requirements-pt.txt`

doctr/models/recognition/utils.py
```diff
@@ -0,0 +1,85 @@
...
+# See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0.txt> for full license details.
+
+from typing import List
+from Levenshtein import distance
+
+
+__all__ = ['merge_sequences', 'merge_multi_sequences']
+
+
+def merge_sequences(a: str, b: str, dil_factor: float) -> str:
```
fg-mindee, line 13: for clarity I suggest: `merge_strings`

doctr/models/recognition/utils.py
```diff
@@ -0,0 +1,85 @@
...
+    """Compute the (best) overlap between 2 character sequences and merge them.
+
+    Args:
+        a: first char seq, suffix should be close to b's prefix.
+        b: second char seq, prefix should be close to a's suffix.
+        dil_factor: dilation factor of the boxes to overlap, should be > 1. This parameter is
+        only used when the mother sequence is splitted on a character repetition
```
fg-mindee, line 20: you need to indent that by 4 spaces for the documentation to render correctly ;)

doctr/models/recognition/utils.py
```diff
@@ -0,0 +1,85 @@
...
+
+from typing import List
+from Levenshtein import distance
+
+
+__all__ = ['merge_sequences', 'merge_multi_sequences']
+
+
+def merge_sequences(a: str, b: str, dil_factor: float) -> str:
+    """Compute the (best) overlap between 2 character sequences and merge them.
```
fg-mindee, line 14: Suggestion:
"Merges 2 character sequences in the best way to maximize the alignment of their overlapping characters"

doctr/models/recognition/utils.py
```diff
@@ -0,0 +1,85 @@
...
+
+    Args:
+        a: first char seq, suffix should be close to b's prefix.
```
fg-mindee, line 17: "close" --> "similar"

doctr/models/recognition/utils.py
```diff
@@ -0,0 +1,85 @@
...
+        b: second char seq, prefix should be close to a's suffix.
```
fg-mindee, line 18: same

doctr/models/recognition/utils.py
```diff
@@ -0,0 +1,85 @@
...
+            if score < min_score:
+                min_score, index = score, i + 1  # Add one because first index is an overlap of 1 char
+
+    # Merge with correct overlap
+    if index == 0:
+        return a + b
+    return a[:-1] + b[index - 1:]
+
+
+def merge_multi_sequences(seq_list: List[str], dil_factor: float) -> str:
```
fg-mindee, line 62: Suggestion "merge_string_sequences"

doctr/models/recognition/utils.py
```diff
@@ -0,0 +1,85 @@
...
+    """Wrapper for the resursive version of merge_sequences
+    Compute the merged string from a list of strings:
```
fg-mindee, line 64: Recursively merges consecutive string sequences with overlapping characters

doctr/models/recognition/utils.py
```diff
@@ -0,0 +1,85 @@
...
+
+    inputs:
```
fg-mindee, line 66: Args:

doctr/models/recognition/utils.py
```diff
@@ -0,0 +1,85 @@
...
+        seq_list: list of sequences to merge. Sequences need to be ordered from left to right.
+        dil_factor: dilation factor of the boxes to overlap, should be > 1. This parameter is
+        only used when the mother sequence is splitted on a character repetition
```
fg-mindee, line 69: same here

doctr/models/recognition/utils.py
```diff
@@ -0,0 +1,85 @@
...
+
+    Returns:
+        A merged character sequence
+
+    Example::
+        >>> from doctr.model.recognition.utils import merge_multi_sequences
+        >>> merge_multi_sequences(['abc', 'bcdef', 'difghi', 'aijkl'], 1.4)
+        'abcdefghijkl'
+    """
+    def merge_sequences_rec(a: str, seq_list: List[str], dil_factor: float) -> str:
```
fg-mindee, line 79: this is a private recursive function, let's take a private name to make it clear:
"_recursive_merge"

doctr/models/recognition/core.py
```diff
@@ -80,6 +80,8 @@ def __init__(
 
         self.pre_processor = pre_processor
         self.model = model
+        self.critical_ar = 8  # Critical aspect ratio
```
fg-mindee, line 83: Mind adding an option to use the crop_split?
Like `use_crop_splitting` as an arg that is saved as an attribute?

fg-mindee (commented):
Thanks again, I added a few suggestions!

doctr/models/recognition/core.py
```diff
@@ -93,8 +95,29 @@ def __call__(
...
+            splitted_crops: List[np.ndarray] = []
+            splitted_idxs: List[List[int]] = []
+            for crop in crops:
+                h, w = crop.shape[:2]
+                aspect_ratio = w / h
+                if aspect_ratio > self.critical_ar:
+                    # Determine the number of crops, reference aspect ratio = 4 = 128 / 32
+                    n_crops = int(aspect_ratio // 4)
+                    # Find the new widths, additional dilation factor to overlap crops
+                    width = int(self.dil_factor * w / n_crops)
```
fg-mindee, line 107: Better to avoid casting it to int right now (will be done on line 113)
This could lead to rounding errors otherwise

doctr/models/recognition/core.py
```diff
@@ -93,8 +95,29 @@ def __call__(
...
+                    centers = [int((w / n_crops) * (1 / 2 + i)) for i in range(n_crops)]
```
fg-mindee, line 108: same here, let's avoid int casting

doctr/models/recognition/core.py
```diff
@@ -93,8 +95,29 @@ def __call__(
...
+                    # Crop and keep track of indexes
+                    splitted_idxs.append([len(splitted_crops) + i for i in range(n_crops)])
+                    splitted_crops.extend([
+                        crop[
+                            :, max(0, round(centers[i] - width / 2)):min(w - 1, round(centers[i] + width / 2)), :
+                        ]
+                        for i in range(n_crops)])
```
fg-mindee, line 115: No need for the range
```python
crop[:, max(0, int(round(center - width / 2))): min(w - 1, int(round(center + width / 2)))]
for center in centers
])
```

doctr/models/recognition/core.py
```diff
@@ -105,4 +128,25 @@ def __call__(
...
+            # Find if crops were splitted, and if so merge back the predictions
+            if len(splitted_idxs):
+                merged_out = []
+                out_idx = 0
+                for splitted_list in splitted_idxs:
+                    # Iterate over words to reconstruct
+                    while out_idx < splitted_list[0]:
+                        # Append unsplitted words
+                        merged_out.append(out[out_idx])
+                        out_idx += 1
```
fg-mindee, line 140: When `while` can be avoided, let's do it :)
```python
merged_out.extend(out[out_idx: splitted_list[0]])
out_idx = splitted_list[0]
```

doctr/models/recognition/core.py
```diff
@@ -105,4 +128,25 @@ def __call__(
...
+                    # Merge splitted words
+                    merged = merge_multi_sequences([out[i][0] for i in splitted_list], self.dil_factor)
+                    merged_score = min([out[i][1] for i in splitted_list])
+                    merged_out.append((merged, merged_score))
+                    out_idx += len(splitted_list)
+                # Append last unsplitted words after the last reconstructed word
+                while out_idx < len(out):
+                    merged_out.append(out[out_idx])
+                    out_idx += 1
```
fg-mindee, line 149: Same:
```python
merged_out.extend(out[out_idx: len(out)])
```

fg-mindee (commented):
A few additional improvement suggestions :sweat_smile:

doctr/models/recognition/core.py
```diff
@@ -93,8 +95,29 @@ def __call__(
             if any(crop.ndim != 3 for crop in crops):
                 raise ValueError("incorrect input shape: all crops are expected to be multi-channel 2D images.")
 
+            splitted_crops: List[np.ndarray] = []
+            splitted_idxs: List[List[int]] = []
+            for crop in crops:
+                h, w = crop.shape[:2]
```
fg-mindee, line 101: In case we have a channels_first tensor:
```python
channels_first = isinstance(crops[0], np.ndarray) or is_tf_available()
h, w = crop.shape[:2] if channels_first else crop.shape[-2:]
```

doctr/models/recognition/core.py
fg-mindee, line 115: Actually:
```python
if channels_last:
    splitted_crops.extend([
        crop[:, max(0, int(round(center - width / 2))): min(w - 1, int(round(center + width / 2)))]
        for center in centers
    ])
else:
    splitted_crops.extend([
        crop[:, :, max(0, int(round(center - width / 2))): min(w - 1, int(round(center + width / 2)))]
        for center in centers
    ])
```

fg-mindee (commented):
And final comments to make sure it works with PyTorch tensors or any channels_last tensors

doctr/models/recognition/__init__.py
```diff
@@ -3,3 +3,4 @@
 from .master import *
 from .sar import *
 from .zoo import *
+from ...file_utils import is_tf_available
```
fg-mindee, line 6: we don't need this?

doctr/models/recognition/core.py
```diff
@@ -9,6 +9,7 @@
 from ..preprocessor import PreProcessor
```
fg-mindee, line 9: You'll need to import `is_tf_available`:

```python
from doctr.file_utils import is_tf_available
```

doctr/models/recognition/utils.py
fg-mindee: still missing the indentation

fg-mindee (changes requested):
Thanks, just a few things left!

fg-mindee:
@charlesmindee  On line 83 of `setup.py`, you're missing:
```python
deps["python-Levenshtein"],
```
(the docker job is failing because of this)

fg-mindee:
While being at it, you're missing a docstring indentation https://github.com/mindee/doctr/pull/465#discussion_r708322349

fg-mindee (approved):
Thrilled about this feature :tada: Thanks!

PR #466, opened 2021-09-13
docs: Added favicon to the documentation

fg-mindee:
This PR simply adds the favicon to the documentation so that tabs are web browser display the logo instead of the default one.

Here is how it renders now:
tab_doctr

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #467, opened 2021-09-13
refactor: Refactored VOCABs

fg-mindee:
- fixed the french vocab (to avoid breaking change, the `french` vocab was remapped to `legacy_french`)
- added multiple european vocabs in a more flexible way (english, spanish, german)
- updated documentation accordingly

_Note: the `legacy_french` vocab will be deprecated as soon as all models have been retrained with the correct `french` vocab_

Any feedback is welcome!


charlesmindee (approved):
Thanks for that!

PR #469, opened 2021-09-14
chore: Added CI job to check recognition training

fg-mindee:
As per #429, this PR introduces the following modifications:
- refactors the CI job to check character classification training
- added a CI job to check text recognition training

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #470, opened 2021-09-14
chore: Removed some CI jobs with macOS

fg-mindee:
Unfortunately, the CI is getting slower by the way and the suspect is mostly macOS jobs. This PR keeps macOS jobs to build the project and collect environment but removes the rest to speed up PR CI jobs.

Any feedback is welcome!


charlesmindee (approved):
Thanks for the cut!

PR #471, opened 2021-09-14
docs: performances update following up the crop splitting feature

charlesmindee:
This PR adds the benchmark of all models with the new crop splitting feature activated, for some datasets it changes almost nothing, but for our private datasets (especially IDS with a lot of MRZ) it improves significantly performances and beyond that results are far more consistent when dealing with extremely long sequences, even if the exact match metric does not always reflect that (one small mistake over 45 chars is better than 45 mistakes...).

Any feedback is welcome!


fg-mindee (approved):
Looks good to me!

PR #473, opened 2021-09-15
fix: Prevents rounding errors in RandomCrop

fg-mindee:
In very rare cases, the generated coordinates of the crop from `RandomCrop` were out of bounds (negative usually). This PR ensures that proper rounding is used and adds clipping to avoid issues.

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #474, opened 2021-09-15
refactor: evaluate scripts and OCRDataset

charlesmindee:
This PR changes the evaluation script so it can be run with rotated boxes annotations, and the OCRDataset object has been changed to work with all test sets. 

DocDataset has been removed, and all private test sets have been cleaned up and refactored to work with and OCRDataset object and the evaluation script.

Any feedback is welcome!


fg-mindee (approved):
Thanks!

PR #481, opened 2021-09-20
refactor: Refactored API of doctr.models

fg-mindee:
- cleaned the `__init__` files of `doctr.models` to keep only the relevant modules accessible (private modules that will see their API change in the future are better left out of this)
- moved `DocumentBuilder` from `doctr/models/core.py` to `doctr/models/builder.py`
- merged `DetectionModel` and `RecognitionModel` into `BaseModel`
- each predictor class is now split into its implementations in PyTorch & TensorFlow via the same dispatch mechanism as previously (for more specific operations, typing and interface)
- move crop split from `doctr/models/recognition/core.py` to `doctr/models/recognition/base.py`
- updated unittests
- updated API
- updated demo
 
Any feedback is welcome!


charlesmindee (dismissed):
thanks for the PR!

charlesmindee (approved):
Thanks!

PR #483, opened 2021-09-20
feat: Added rectangular stride MobileNets

fg-mindee:
- adds MobileNet V3 counterparts with 3 rectangular strides
- replaces mobilenet backbones in CRNN

I checked locally and the training now yields 92%+ of exact match after a single epoch (vs. 20% before)

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #484, opened 2021-09-21
fix: extract_(r)crops for pytorch predictor (channels_last=False)

charlesmindee:
This PR fixes the extract_crops and extract_rcrops functions which were not working for pytorch predictors (when channels_last=False)

Any feedback is welcome!


fg-mindee (dismissed):
Thanks! Perhaps we should add some unittests for crop extraction?
This PR would be a good occasion if you feel like it :ok_hand:

fg-mindee (approved):
Thanks!

PR #485, opened 2021-09-21
feat: add db_mobilenet_v3_large (TF) ckpt & benchmark

charlesmindee:
This PR adds both checkpoints & benchmark in docs for DB with mobilenet_v3_large in TF.

Any feedback is welcome!


fg-mindee (approved):
Thanks!

PR #488, opened 2021-09-22
Rotate page

Rob192:
This is part of the solutions discussed in #225 and address the point "Integrate the feature in the predictor while being disabled by default" of the "Page-level orientation" part. 

It implements the function `rotate_page` and tracks the  'page_angles'


Dockerfile
```diff
@@ -17,3 +17,7 @@ RUN apt-get update \
     && apt-get autoremove -y \
     && rm -rf /var/lib/apt/lists/* \
     && rm -rf /root/.cache/pip
+
+COPY ./demo .
+RUN pip install -r requirements.txt
+CMD [ "streamlit", "run", "app.py" ]
```
fg-mindee, line 23: This Dockerfile is meant to use the python library, we might add a dockerfile specific for the demo but for now, let's not modify this file :pray:

doctr/models/_utils.py
```diff
@@ -101,6 +101,31 @@ def extract_rcrops(
...
+        # if image is closer to portrait mode, go for a full 90° rotation
+        image = cv2.rotate(image, cv2.cv2.ROTATE_90_CLOCKWISE)
+        angle += 90  # track the rotation by updating angle
+
+    height, width = image.shape[:2]
+    center = (height / 2, width / 2)
+    rot_mat = cv2.getRotationMatrix2D(center, angle, 1.0)
+    return cv2.warpAffine(image, rot_mat, (width, height))
+
+
```
fg-mindee, line 128: I'm not entirely sure, but it looks like `doctr.utils.geometry.rotate_image` is performing the same operation?

doctr/models/predictor/tensorflow.py
```diff
@@ -43,13 +42,22 @@ def __init__(
     def __call__(
         self,
         pages: List[Union[np.ndarray, tf.Tensor]],
+        rotate_document: bool = False,
```
fg-mindee, line 45: For consistency, I would suggest setting this as a class attribute in the constructor

As perhaps rename it to something rather descriptive like `straighten_pages`

doctr/models/predictor/tensorflow.py
```diff
@@ -43,13 +42,22 @@ def __init__(
...
         if any(page.ndim != 3 for page in pages):
             raise ValueError("incorrect input shape: all pages are expected to be multi-channel 2D images.")
 
+        # Detect document rotation and rotate pages
+        if rotate_document:
+            page_angles = []
+            for i, page in enumerate(pages):
+                page_angle = estimate_orientation(page)
+                page_angles.append(page_angle)
+                pages[i] = rotate_page(page, page_angle)
```
fg-mindee, line 59: Let's do the same with list comprehensions :) 

```python
page_orientations = [estimate_orientation(page) for page in pages]
pages = [rotate_page(page, angle) for page, angle in zip(pages, page_orientations)]
```

fg-mindee (changes requested):
Thanks a lot to jump back into this :pray: 

I added a few comments but for me, there is only one major part missing:
- with this PR, we rotate the page firsthand before any DL model forward
- the process is almost identical, but the localizations of words have to be remapped into the original pages (meaning if we detect that the page has a 30° orientation, we forward its straightened version, if we detect a straight bounding box in this, we have to reproject it correctly so that the localization refers to the original page)

Let me know what you think :)

Dockerfile
Rob192: Yep, sorry about this. Will do

doctr/models/_utils.py
Rob192: oh yes indeed !

doctr/models/predictor/tensorflow.py
Rob192: ok !

doctr/models/predictor/tensorflow.py
Rob192: sure !

fg-mindee:
@Rob192 also, if you don't minde merging the `main` branch, there seems to be some conflicts :sweat_smile:

Rob192:
> @Rob192 also, if you don't minde merging the `main` branch, there seems to be some conflicts 😅
Yes ! you guys are mooving too fast for me 😮

Rob192:
Hello,

Thanks for the review ! My objective is now to use geometry.rotate_image to rotate the documents before going into the predictor and then use the same function to rotate every box, compare to the center of the document previously used. 

However, I am feeling a bit confused regarding the `expand` functionnality you added in several functions of `Doctr`. I hoped you could help me clarify. I think you goal was to do the same for every function, i.e. given a image of shape (10,20), the rotation by an angle of 90° with `expand=True` would give a result image of shape (20,10). This is not the case for all the function in `utils.geometry` :
 
```python
import numpy as np

import torch
import tensorflow as tf

from doctr.utils import geometry
from doctr.transforms.functional.tensorflow import rotate as rotate_tf
from doctr.transforms.functional.pytorch import rotate as rotate_to

_, axes = plt.subplots(3, 1)

img = np.ones((32, 64, 3), dtype=np.float32)
rotated = geometry.rotate_image(img, 45., expand=True)
print(rotated.shape)
axes[0].imshow(rotated)

input_t = torch.ones((3, 32, 64), dtype=torch.float32)
boxes = np.array([
    [15, 20, 35, 30]
])
r_img, r_boxes = rotate_to(input_t, boxes, angle=45., expand=True)
print(r_img.shape)
axes[1].imshow(r_img.permute((1,2,0)))

input_t = tf.ones((32, 64, 3), dtype=tf.float32)
boxes = np.array([
    [15, 20, 35, 30]
])
r_img, r_boxes = rotate_tf(input_t, boxes, angle=45., expand=True)
print(r_img.shape)
axes[2].imshow(r_img)
```

Do you think I could modify `geometry.rotate_image` so that `expand=True` has the same behaviour as the two other functions in `doctr.transforms.functional` ?

Dockerfile
```diff
@@ -16,4 +16,4 @@ RUN apt-get update \
     && pip cache purge \
     && apt-get autoremove -y \
     && rm -rf /var/lib/apt/lists/* \
-    && rm -rf /root/.cache/pip
+    && rm -rf /root/.cache/pip
```
fg-mindee, line 19: You have to add an EOF to remove the change :)

doctr/models/predictor/tensorflow.py
```diff
@@ -41,13 +44,20 @@ def __init__(
...
         **kwargs: Any,
     ) -> Document:
 
         # Dimension check
         if any(page.ndim != 3 for page in pages):
             raise ValueError("incorrect input shape: all pages are expected to be multi-channel 2D images.")
 
+        # Detect document rotation and rotate pages
+        if straighten_pages:
+            page_angles = []
```
fg-mindee, line 57: You can remove this line

doctr/models/predictor/tensorflow.py
```diff
@@ -41,13 +44,20 @@ def __init__(
...
+            page_orientations = [estimate_orientation(page) for page in pages]
+            pages = [rotate_image(page, angle) for page, angle in zip(pages, page_orientations)]
+
```
fg-mindee, line 60: We're still missing something to reproject the detections into the original image after `self._process_predictions`:
- right now if we merge, the pages will be rotated for inference
- but we need to map them back after inference

test/common/test_models.py
```diff
@@ -92,3 +92,10 @@ def test_get_bitmap_angle(mock_bitmap):
 def test_estimate_orientation(mock_image):
     angle = estimate_orientation(mock_image)
     assert abs(angle - 30.) < 1.
+
+
+def test_rotate_page(mock_image):
```
fg-mindee, line 97: I would argue it's actually `test_estimate_orientation` :)

fg-mindee (changes requested):
Thanks for the edits! Regarding your question about `rotate_image`, it's still an ongoing implementation so we can add optional features in it. But to keep the PR short, perhaps this should be implemented in the predictor directly? (we can refactor in another PR once we get this working)

test/common/test_models.py
Rob192: Indeed ! I am being a scatterbrain !

doctr/models/predictor/tensorflow.py
Rob192: Yes done !

Rob192:
Hello @fg-mindee ! 
Here are my modifications to remap the localizations of the words back into the original pages. Here are some details of the choices I did : 
- I created a new `rotate_image` function inside `doctr/models/_utils.py` per your recommandation. I removed the last bit of the function to keep the size of the original image. I introduced a mask functionality to crop the padding at the end of the + then - rotation.  
- I had to modify `rotate_boxes` to follow the `expand` that is done in `rotate_image`. The calculation for the rotation of the boxes was not right and I fixed that. There is also a mask functionality in `rotate_boxes`
I hope this makes sense to you guys

doctr/models/predictor/tensorflow.py
Rob192: ok

doctr/utils/geometry.py
```diff
@@ -129,10 +133,41 @@ def rotate_abs_boxes(boxes: np.ndarray, angle: float, img_shape: Tuple[int, int]
     return rotated_boxes
 
 
+def remap_boxes(boxes: np.ndarray, orig_shape: Tuple[int, int], dest_shape: Tuple[int, int]) -> np.ndarray:
```
fg-mindee, line 136: So actually, I realized a week ago that this is an issue:
if the shapes are different and the rbox is in relative coords, this doesn't work
If the aspect ratio is different, simply rescaling doesn't work, and the angle may change :/

fg-mindee (commented):
Thanks again Rob!
I think there are still a few edges cases we are not covering but this is taking a very solid shape!

doctr/models/_utils.py
```diff
@@ -188,3 +190,47 @@ def get_bitmap_angle(bitmap: np.ndarray, n_ct: int = 20, std_max: float = 3.) ->
             angle = 90 + angle
 
     return angle
+
+
+def rotate_image(
```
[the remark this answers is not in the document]
Rob192: Yeah, I think also that we should merge them. I will propose something.

doctr/models/predictor/tensorflow.py
```diff
@@ -56,5 +66,15 @@ def __call__(
         word_preds = self.reco_predictor([crop for page_crops in crops for crop in page_crops], **kwargs)
 
         boxes, text_preds = self._process_predictions(loc_preds, word_preds, self.doc_builder.rotated_bbox)
+
+        # Rotate back pages and boxes while keeping original image size
+        if self.straighten_pages:
+            pages = [rotate_image(page, angle, expand=True, mask_shape=mask) for page, angle, mask in
```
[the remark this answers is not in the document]
Rob192: yes indeed, we don't really need that. I used it for debug purposes, I just removed it.

doctr/utils/visualization.py
```diff
@@ -99,7 +99,7 @@ def polygon_patch(
     # Switch to absolute coords
     x, w = x * width, w * width
     y, h = y * height, h * height
-    points = cv2.boxPoints(((x, y), (w, h), a))
+    points = cv2.boxPoints(((x, y), (w, h), -a))
```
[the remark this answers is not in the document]
Rob192: Mmm ... It is clockwise angle with `cv2.boxPoints` .. and the angles I am passing for the boxes are also given in the trigonometric referencial ... however the image is not properly reconstructed if I do not put this minus here.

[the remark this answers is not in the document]
fg-mindee: Yes thanks for that!

doctr/utils/geometry.py
fg-mindee: Well I think it actually is:
- imagine first a squared image, with an orientation of 45° (box behind read in the direction of the top right corner)
- now you switch to relative coords the rotated bounding box
- now the image is resized to a shape of (100, 200)
- remapping this is a warping transformation, and the box would still be pointing to the top right corner which is now about ~26.6°

So it's actually a bit more complex than it sounds :upside_down_face:

doctr/utils/geometry.py
```diff
@@ -3,15 +3,15 @@
...
+from math import ceil
 from typing import List, Union, Tuple, Optional
 import numpy as np
 import cv2
 from .common_types import BoundingBox, Polygon4P, RotatedBbox, Bbox
 
 __all__ = ['rbbox_to_polygon', 'bbox_to_polygon', 'polygon_to_bbox', 'polygon_to_rbbox',
            'resolve_enclosing_bbox', 'resolve_enclosing_bbox', 'fit_rbbox', 'rotate_boxes', 'rotate_abs_boxes',
-           'compute_expanded_shape', 'rotate_image']
+           'compute_expanded_shape']
```
fg-mindee, line 14: typo here I think, we need "rotate_image" over here :)

doctr/utils/geometry.py
```diff
@@ -213,14 +213,18 @@ def rotate_boxes(
 def rotate_image(
     image: np.ndarray,
     angle: float,
-    expand=False,
+    expand: bool = False,
+    keep_original_size: bool = False,
```
fg-mindee, line 217: Don't you think this can be done by adding a resize transformation after the rotation instead?

doctr/utils/geometry.py
```diff
@@ -250,4 +254,15 @@ def rotate_image(
...
+    if mask_shape is not None:
+        if len(mask_shape) != 2:
+            raise ValueError(f"Mask length should be 2, was found at: {len(mask_shape)}")
+        h_crop, w_crop = int(height - ceil(mask_shape[0])), int(ceil(width - mask_shape[1]))
+        if h_crop > 0 and w_crop > 0:
+            rot_img = rot_img[h_crop // 2: - h_crop // 2, w_crop // 2: - w_crop // 2]
+        elif w_crop <= 0:
+            rot_img = rot_img[h_crop // 2: - h_crop // 2, ]
+        elif h_crop <= 0:
+            rot_img = rot_img[:, w_crop // 2: - w_crop // 2]
```
fg-mindee, line 266: Careful here, by cropping, we will be loosing some information :/

fg-mindee (commented):
Still have a few comments, but the major one being about the reprojection of rotated bounding boxes

doctr/models/zoo.py
```diff
@@ -12,15 +12,15 @@
 __all__ = ["ocr_predictor"]
 
 
-def _predictor(det_arch: str, reco_arch: str, pretrained: bool, det_bs=2, reco_bs=128) -> OCRPredictor:
+def _predictor(det_arch: str, reco_arch: str, pretrained: bool, det_bs=2, reco_bs=128, *kwargs) -> OCRPredictor:
```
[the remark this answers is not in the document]
Rob192: This is a mistake, I will remove it. For the moment, I don't know how to integrate the `rotated_bbox ` nor the `straighten_pages` args to this function. I was thinking that we will do it as part of the third part of this PR when we will be evaluating the impact on performances of this work. Is that OK with you ?

Rob192:
@fg-mindee I did remove the cropping of the image as I used it only for debugging. I tried to address your other comments few days ago, please advise if you still have concerns.

doctr/utils/geometry.py
```diff
@@ -3,11 +3,11 @@
...
 # See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0.txt> for full license details.
 
-import math
-from typing import List, Union, Tuple
+from math import ceil
+from typing import List, Union, Tuple, Optional
 import numpy as np
 import cv2
-from .common_types import BoundingBox, Polygon4P, RotatedBbox
+from .common_types import BoundingBox, Polygon4P, RotatedBbox, Bbox
```
fg-mindee, line 10: I don't think we're using this extra import

doctr/utils/geometry.py
```diff
@@ -143,10 +143,40 @@ def rotate_abs_boxes(boxes: np.ndarray, angle: float, img_shape: Tuple[int, int]
     return rotated_boxes
 
 
+def remap_boxes(boxes: np.ndarray, orig_shape: Tuple[int, int], dest_shape: Tuple[int, int]) -> np.ndarray:
+    """ Remaps a batch of RotatedBbox (x, y, w, h, alpha) expressed for an origin_shape to a destination_shape,
+    This does not impact the absolute shape of the boxes
```
fg-mindee, line 148: Could you specify the description please? Remap from origin shape to destination shape, but for which operation (resize / rotation) ?

doctr/utils/geometry.py
```diff
@@ -168,29 +206,31 @@ def rotate_boxes(
...
+        rotated_boxes = remap_boxes(rotated_boxes, orig_shape=orig_shape, dest_shape=mask_shape)
     return rotated_boxes
 
 
 def rotate_image(
     image: np.ndarray,
     angle: float,
-    expand=False,
+    expand: bool = False,
+    preserve_aspect_ratio: bool = False
```
fg-mindee, line 225: could you add a comma at the end of the line please?

doctr/utils/geometry.py
```diff
@@ -168,29 +206,31 @@ def rotate_boxes(
...
 ) -> np.ndarray:
     """Rotate an image counterclockwise by an given angle.
 
     Args:
         image: numpy tensor to rotate
         angle: rotation angle in degrees, between -90 and +90
         expand: whether the image should be padded before the rotation
+        preserve_aspect_ratio: whether the image should be resized to the original image size after the rotation
```
fg-mindee, line 233: The meaning is a bit different from the preserve_aspect_ratio, I would suggest "preserve_origin_shape" instead

fg-mindee (commented):
Alright, we're almost there!
I added some comments but here are the general things to take care of:
- there seems to be some conflicts with `main`, could you merge it into this branch and resolve the conflicts please?
- flake8 isn't happy, could you fix the lint please?
- do you think you could apply your modifications to the Pytorch predictor as well?
- we're gonna need to extend the unittest of predictors since this introduces a new flag & feature for them

Let me know if you have any questions!

fg-mindee:
Might be worth having your opinion as well @charlesmindee (no hurry)

Rob192:
Hello ! Sorry for the delay in answering this @fg-mindee, it was holiday time for me. 

So I did the merge and had to fix a few things to make it work. Mainly the changes were done for two things. 
- I think the `boxes` formats were not consistent, I hope they are now. Arrays of `boxes` with `boxes.shape[1] == 4` are straight boxes, `boxes.shape[1] == 5` are rotated boxes. 
- In `rotate_image` I had to remove the conversion to `float32` of the image. Indeed, I rotate first the image and with this format, the predictors were not happy and couldn't find the texts in the image. I am not sure of the implications of this change so I let you guys tell me if it should be done differently.

doctr/models/predictor/base.py
```diff
@@ -79,7 +79,10 @@ def _process_predictions(
             boxes, angles = zip(*loc_preds)
             # Rotate back boxes if necessary
             if allow_rotated_boxes:
-                boxes = [rotate_boxes(page_boxes, angle) for page_boxes, angle in zip(boxes, angles)]
+                boxes = [rotate_boxes(page_boxes[:, :-1], angle) for page_boxes, angle in zip(boxes, angles)]
+            # if straight boxes, convert to bbox
+            else:
+                boxes = [page_boxes[:, :-1] for page_boxes in boxes]
```
fg-mindee, line 85: sorry about this, but since #641, I think this needs to be updated :sweat_smile:

doctr/models/predictor/pytorch.py
```diff
@@ -25,6 +27,9 @@ class OCRPredictor(nn.Module, _OCRPredictor):
     Args:
         det_predictor: detection module
         reco_predictor: recognition module
+        straighten_pages: if True, evaluates the page general orientation based on the median of each line orientation.
+            Then, rotates page before using Detection and Recognition Predictors. Then rotates page back to original
```
fg-mindee, line 31: "before using..." --> "before passing it to the deep learning modules"

doctr/models/predictor/pytorch.py
```diff
@@ -25,6 +27,9 @@ class OCRPredictor(nn.Module, _OCRPredictor):
...
+            orientation. This improves the Detection and Recognition Predictors predictions.
```
fg-mindee, line 32: "Then rotates page back to original orientation" --> "The final predictions will be remapped accordingly"

doctr/models/predictor/pytorch.py
```diff
@@ -25,6 +27,9 @@ class OCRPredictor(nn.Module, _OCRPredictor):
     Args:
         det_predictor: detection module
         reco_predictor: recognition module
+        straighten_pages: if True, evaluates the page general orientation based on the median of each line orientation.
```
fg-mindee, line 30: "evaluates" --> "estimates"
"median of each line orientation" --> "median line orientation"

doctr/models/predictor/pytorch.py
fg-mindee: "This improves the Detection and Recognition Predictors predictions." --> "Doing so will improve performances for documents with page-uniform rotations"

doctr/models/predictor/tensorflow.py
```diff
@@ -25,6 +27,9 @@ class OCRPredictor(NestedObject, _OCRPredictor):
     Args:
         det_predictor: detection module
         reco_predictor: recognition module
+        straighten_pages: if True, evaluates the page general orientation based on the median of each line orientation.
+            Then, rotates page before using Detection and Recognition Predictors. Then rotates page back to original
+            orientation. This improves the Detection and Recognition Predictors predictions.
```
fg-mindee, line 32: same here

tests/common/test_models.py
```diff
@@ -93,3 +94,8 @@ def test_get_bitmap_angle(mock_bitmap):
 def test_estimate_orientation(mock_image):
     angle = estimate_orientation(mock_image)
     assert abs(angle - 30.) < 1.
+
+    angle = estimate_orientation(mock_image)
```
fg-mindee, line 98: no need to recompute this (cf. line 95)

tests/common/test_models.py
```diff
@@ -93,3 +94,8 @@ def test_get_bitmap_angle(mock_bitmap):
...
+    rotated = geometry.rotate_image(mock_image, -angle)
+    angle_rotated = estimate_orientation(rotated)
+    assert abs(angle_rotated - 0.) < 1.
```
fg-mindee, line 101:
```python
assert abs(angle_rotated) < 1.
```

doctr/utils/geometry.py
```diff
@@ -145,22 +145,62 @@ def rotate_abs_boxes(boxes: np.ndarray, angle: float, img_shape: Tuple[int, int]
...
+
+    if len(dest_shape) != 2:
+        raise ValueError(f"Mask length should be 2, was found at: {len(dest_shape)}")
+    if len(orig_shape) != 2:
+        raise ValueError(f"Image_shape length should be 2, was found at: {len(orig_shape)}")
+    orig_height, orig_width = orig_shape
+    dest_height, dest_width = dest_shape
+    mboxes = boxes.copy()
+    mboxes[:, 0] = ((boxes[:, 0] * orig_width) + (dest_width - orig_width) / 2) / dest_width
+    mboxes[:, 1] = ((boxes[:, 1] * orig_height) + (dest_height - orig_height) / 2) / dest_height
```
fg-mindee, line 171: I'm not sure I follow: the 2D relative coords of the box center should remain the same, or am I missing something? :thinking:

fg-mindee: Oh my bad, I think I get it now, but we have to specify it's because of the expansion occurring during the rotation :ok_hand:

doctr/utils/geometry.py
```diff
@@ -145,22 +145,62 @@ def rotate_abs_boxes(boxes: np.ndarray, angle: float, img_shape: Tuple[int, int]
...
+    mboxes[:, 2] = boxes[:, 2] * orig_width / dest_width
+    mboxes[:, 3] = boxes[:, 3] * orig_height / dest_height
```
fg-mindee, line 173: I will have to redo all the calculations to ensure, but aren't the remapped width & height depending on the alpha? Because they represent the width along the orientation (if the box has an alpha of 45°, the width is along the diagonal axis)

doctr/utils/geometry.py
```diff
@@ -145,22 +145,62 @@ def rotate_abs_boxes(boxes: np.ndarray, angle: float, img_shape: Tuple[int, int]
...
     Returns:
         A batch of rotated boxes (N, 5): (x, y, w, h, alpha) or a batch of straight bounding boxes
     """
+    # Change format of the boxes to rotated boxes
+    if boxes.shape[1] == 4:
+        boxes = np.column_stack(((boxes[:, 0] + boxes[:, 2]) / 2,
+                                 (boxes[:, 1] + boxes[:, 3]) / 2,
+                                 boxes[:, 2] - boxes[:, 0],
+                                 boxes[:, 3] - boxes[:, 1],
+                                 np.zeros(boxes.shape[0])))
```
fg-mindee, line 203: I agree with the conversion but I would argue it's better to avoid inplace modifications of the input arg

doctr/utils/geometry.py
```diff
@@ -201,15 +243,16 @@ def rotate_image(
...
         exp_img = np.pad(image, ((h_pad // 2, h_pad - h_pad // 2), (w_pad // 2, w_pad - w_pad // 2), (0, 0)))
     else:
         exp_img = image
 
     height, width = exp_img.shape[:2]
     rot_mat = cv2.getRotationMatrix2D((width / 2, height / 2), angle, 1.0)
-    rot_img = cv2.warpAffine(exp_img.astype(np.float32), rot_mat, (width, height))
-    if expand:
+    rot_img = cv2.warpAffine(exp_img, rot_mat, (width, height))
+    if preserve_origin_shape:
```
fg-mindee, line 255: I think this needs to be reverted

`if preserve_origin_shape:` should be the conditional check for the rescale inside this "if", or am I missing something?

doctr/utils/geometry.py
```diff
@@ -170,29 +210,31 @@ def rotate_boxes(
...
+    expand: bool = False,
+    preserve_origin_shape: bool = False,
 ) -> np.ndarray:
     """Rotate an image counterclockwise by an given angle.
 
     Args:
         image: numpy tensor to rotate
         angle: rotation angle in degrees, between -90 and +90
         expand: whether the image should be padded before the rotation
+        preserve_origin_shape: whether the image should be resized to the original image size after the rotation
```
fg-mindee, line 237: let's specify that this only does something if `expand=True` (if there is no expansion, the output shape is identical to the input)

fg-mindee (changes requested):
Thanks a lot Robin!

I think due to a recent PR, you're gonna have to resolve conflicts again. But I added comments to try to merge this PR quickly! Let me know if you have any questions :ok_hand:

doctr/models/predictor/base.py
Rob192: merging done ..

doctr/models/predictor/pytorch.py
Rob192: ok done

doctr/models/predictor/pytorch.py
Rob192: Done

doctr/models/predictor/pytorch.py
Rob192: Done

doctr/models/predictor/tensorflow.py
Rob192: Done

doctr/utils/geometry.py
Rob192: Yes we discussed that, I tried to answer your concerns with a stricter testing in `test_remap_boxes` with a 45° inclined box. Width and Height are independent from alpha.

doctr/utils/geometry.py
Rob192: oh yes, sorry, i changed this

doctr/utils/geometry.py
Rob192: Not sure about what you mean : `if preserve_origin_shape:` is indeed the conditional check for the rescale inside this `if`. Or do you prefer it to be changed to `if expand and preserve_origin_shape:` ?

doctr/utils/geometry.py
Rob192: Yes done !

doctr/utils/geometry.py
Rob192: the box should stay at the same position from the center of the box. the test associated is in `test_remap_boxes` with the following snippet: 
```python
pred = geometry.remap_boxes(np.array([[0.25, 0.5, 0.5, 0.33, 0.]]), (80, 30), (160, 30))
target = np.array([[0.375, 0.5, 0.25, 0.1, 0.]])
assert pred.all() == target.all()
```

Rob192:
Hello @fg-mindee and @charlesmindee 
I merged #641 and adressed your points. Hope we are good !

tests/common/test_models.py
Rob192: done

tests/common/test_models.py
Rob192: done

Rob192:
Hello @fg-mindee, I also merged the last commits, we should be good

doctr/utils/geometry.py
fg-mindee: mmmh but that test doesn't work I think, shouldn't it be:
```python
assert np.all(pred == target)
```
or 
```python
assert (pred == target).all()
```

because otherwise that only checks that no value in the array, once casted to a boolean will be False

doctr/utils/geometry.py
fg-mindee: The second option yes

But this means that this line should be kept as is, and only to create an additional `if preserve_origin_shape` inside this loop :+1:

doctr/models/predictor/base.py
```diff
@@ -64,12 +64,13 @@ def _process_predictions(
         word_preds: List[Tuple[str, float]],
     ) -> Tuple[List[np.ndarray], List[List[Tuple[str, float]]]]:
 
-        text_preds = []
+        boxes, text_preds = [], []
         if len(loc_preds) > 0:
             # Text
             _idx = 0
             for page_boxes in loc_preds:
+                boxes.append(page_boxes[:, :-1])
```
fg-mindee, line 72: Sorry I think I lost a bit the reason why we need to remove the last item on the 2nd axis, can you remind me?

doctr/models/predictor/pytorch.py
```diff
@@ -57,6 +64,13 @@ def forward(
         if any(page.ndim != 3 for page in pages):
             raise ValueError("incorrect input shape: all pages are expected to be multi-channel 2D images.")
 
+        origin_page_shapes = [page.shape[:2] for page in pages]
```
fg-mindee, line 67: If this is a PyTorch tensor and not a numpy one, the axis order will be C, H, W so I think this won't work in that case

doctr/models/predictor/pytorch.py
```diff
@@ -70,6 +84,11 @@ def forward(
 
         boxes, text_preds = self._process_predictions(loc_preds, word_preds)
 
+        # Rotate back pages and boxes while keeping original image size
+        if self.straighten_pages:
+            boxes = [rotate_boxes(page_boxes, angle, orig_shape=page.shape[:2], mask_shape=mask) for
```
fg-mindee, line 89: same here, the `.shape` indexing won't work properly on pytorch tensor which are channel first

doctr/utils/geometry.py
```diff
@@ -145,54 +145,101 @@ def rotate_abs_boxes(boxes: np.ndarray, angle: float, img_shape: Tuple[int, int]
...
+    return mboxes
+
+
 def rotate_boxes(
     boxes: np.ndarray,
     angle: float = 0.,
-    min_angle: float = 1.
+    min_angle: float = 1.,
+    orig_shape: Optional[Tuple[int, int]] = None,
+    mask_shape: Optional[Tuple[int, int]] = None,
```
fg-mindee, line 184: target_shape might be more suitable perhaps?

doctr/utils/geometry.py
```diff
@@ -145,54 +145,101 @@ def rotate_abs_boxes(boxes: np.ndarray, angle: float, img_shape: Tuple[int, int]
...
+    centers = np.stack((_boxes[:, 0], _boxes[:, 1]), axis=-1)
+    rotated_centers = .5 + np.matmul(centers - .5, rotation_mat)
     x_center, y_center = rotated_centers[:, 0], rotated_centers[:, 1]
     # Compute rotated boxes
-    rotated_boxes = np.stack((x_center, y_center, width, height, angle * np.ones_like(boxes[:, 0])), axis=1)
+    rotated_boxes = np.stack((x_center, y_center, _boxes[:, 2], _boxes[:, 3], angle * np.ones_like(_boxes[:, 0])),
+                             axis=1)
+    # Apply a mask if requested
+    if mask_shape is not None and orig_shape is not None:
+        rotated_boxes = remap_boxes(rotated_boxes, orig_shape=orig_shape, dest_shape=mask_shape)
```
fg-mindee, line 225: I think we need to explain somewhere (perhaps in arg description) what happens when either of those is None

fg-mindee (changes requested):
Thanks again :pray: One of the unittests seems to fail, could you take a look at it please? I added a few comments for clarification!

doctr/utils/geometry.py
Rob192: :O ! you are right !! I will change other assertions like that !!

Rob192: We have this kind of assertion in other tests, you guys should maybe have a deeper look at that

Rob192: Hello, this kind of assertion can be found in other places in the code (`test_rotate` and `test_resolve_enclosing_bbox`). You guys should have a look !

doctr/models/predictor/base.py
Rob192: Boxes should either be of dimension 4 or 5. as far as I understood `loc_preds` have a probability parameter and can be of dimension 5 or 6. If you don't remove the last item on the 2nd axis, you are passing the rest of the code an array of boxes with wrond dimension. Downstream, there are many checks of the dimension of the array to verify is the boxes are straight or rotated. This gets confusing if you pass them the probability parameter.

doctr/utils/geometry.py
Rob192: OK I added some description that was indeed missing

doctr/utils/geometry.py
Rob192: ok

doctr/models/predictor/pytorch.py
Rob192: Could you please help me here ? I don't know how to debug this. I used the following snippet to force using PyTorch but it seems that the channel is still last. When do we have PyTorch tensors ?
 
```python
from pathlib import Path

from doctr.models.detection.zoo import detection_predictor
from doctr.models.predictor.pytorch import OCRPredictor
from doctr.models.recognition.zoo import recognition_predictor
from doctr.utils.visualization import visualize_page

file = Path(r"myfile.jpg")

doc = DocumentFile.from_images(file)
det_bs=2
reco_bs=128

det_predictor = detection_predictor('db_resnet50', pretrained=True, batch_size=det_bs, assume_straight_pages=True)
reco_predictor = recognition_predictor('crnn_vgg16_bn', pretrained=True, batch_size=reco_bs)

predictor = OCRPredictor(det_predictor=det_predictor, reco_predictor=reco_predictor, assume_straight_pages=True, straighten_pages=True)

# Plot OCR output
out = predictor(doc)

fig = visualize_page(out.pages[0].export(), doc[0], interactive=False)
plt.show()

```

doctr/models/predictor/pytorch.py
```diff
@@ -70,6 +84,11 @@ def forward(
 
         boxes, text_preds = self._process_predictions(loc_preds, word_preds)
 
+        # Rotate back pages and boxes while keeping original image size
+        if self.straighten_pages:
+            boxes = [rotate_boxes(page_boxes, angle, orig_shape=page.shape[:2], target_shape=mask) for
```
fg-mindee, line 89: `page.shape[:2]` --> `page.shape[:2] if isinstance(page, np.ndarray) else page.shape[-2:]`

tests/common/test_utils_geometry.py
```diff
@@ -35,19 +38,75 @@ def test_resolve_enclosing_rbbox():
...
+
+    orig_dimension = (100, 100)
+    dest_dimensions = (100, 200)
+    orig_box = np.array([[0.5, 0.5, 0.2, 0., 45, 0.5]])
+    # Unpack
+    height_o, width_o = orig_dimension
+    height_d, width_d = dest_dimensions
+    pred = geometry.remap_boxes(orig_box, orig_dimension, dest_dimensions)
+
+    x, y, w, h, a, c = orig_box[0]
```
fg-mindee, line 62: Can you replace "c" with "_" to avoid "unused variable warnings" please? :pray:

doctr/models/_utils.py
```diff
@@ -150,7 +150,11 @@ def estimate_orientation(img: np.ndarray, n_ct: int = 50, ratio_threshold_for_li
             angles.append(angle)
         elif w / h < 1 / ratio_threshold_for_lines:  # if lines are vertical, substract 90 degree
             angles.append(angle - 90)
-    return -median_low(angles)
+
+    if len(angles) == 0:
+        return 0  # in case no angles is found
```
fg-mindee, line 155: Mind adding a unittest case for this please?

doctr/models/predictor/pytorch.py
```diff
@@ -57,6 +64,13 @@ def forward(
         if any(page.ndim != 3 for page in pages):
             raise ValueError("incorrect input shape: all pages are expected to be multi-channel 2D images.")
 
+        origin_page_shapes = [page.shape[:2] for page in pages]
+
+        # Detect document rotation and rotate pages
+        if self.straighten_pages:
+            origin_page_orientations = [estimate_orientation(page) for page in pages]
+            pages = [rotate_image(page, -angle, expand=True) for page, angle in zip(pages, origin_page_orientations)]
```
fg-mindee, line 72: Let's add a unittest to check that loop :pray:

doctr/utils/geometry.py
```diff
@@ -145,54 +145,104 @@ def rotate_abs_boxes(boxes: np.ndarray, angle: float, img_shape: Tuple[int, int]
...
+
+    Returns:
+        A batch of rotated loc_preds (N, 6): (x, y, w, h, alpha, c) expressed in the destination referencial
+
+    """
+
+    if len(dest_shape) != 2:
+        raise ValueError(f"Mask length should be 2, was found at: {len(dest_shape)}")
+    if len(orig_shape) != 2:
+        raise ValueError(f"Image_shape length should be 2, was found at: {len(orig_shape)}")
```
fg-mindee, line 166: Mind adding some unittests for this please?

fg-mindee (changes requested):
Thanks for the edits! Everything looks good to me, only two things left:
- updating the spatial shape resolution
- adding a few unittests to avoid the slight decrease in coverage (considering the importance of this feature and the impact it can have on predictors, let's minize risks :sweat_smile:)

Then we can confidently merge :)

tests/common/test_utils_geometry.py
Rob192: ok !

doctr/models/predictor/pytorch.py
Rob192: Ok thx

doctr/utils/geometry.py
Rob192: ok done

doctr/models/_utils.py
Rob192: ok done!

doctr/utils/geometry.py
Rob192: ok great !

doctr/models/predictor/pytorch.py
```diff
@@ -47,8 +47,8 @@ def __init__(
     ) -> None:
 
         super().__init__()
-        self.det_predictor = det_predictor.eval()  # type: ignore[attr-defined]
-        self.reco_predictor = reco_predictor.eval()  # type: ignore[attr-defined]
+        self.det_predictor = det_predictor  # type: ignore[attr-defined]
+        self.reco_predictor = reco_predictor  # type: ignore[attr-defined]
```
fg-mindee, line 51: Is this an error? We should keep this to ensure that predictor are used for inference

tests/common/test_models.py
```diff
@@ -86,12 +86,21 @@ def mock_bitmap(mock_image):
     return bitmap
 
 
+@pytest.fixture(scope="function")
+def mock_unreadable_image(mock_image):
+    image = mock_image * 0
+    return image
+
+
```
fg-mindee, line 94: no need to create a fixture for this :)

tests/common/test_models.py
```diff
@@ -86,12 +86,21 @@ def mock_bitmap(mock_image):
...
 def test_get_bitmap_angle(mock_bitmap):
     angle = get_bitmap_angle(mock_bitmap)
     assert abs(angle - 30.) < 1.
 
 
-def test_estimate_orientation(mock_image):
+def test_estimate_orientation(mock_image, mock_unreadable_image):
+    angle = estimate_orientation(mock_unreadable_image)
+    assert angle == 0
+
```
fg-mindee, line 103:
```python
def test_estimate_orientation(mock_image):
    assert estimate_orientation(mock_image * 0) == 0
```

fg-mindee (changes requested):
One suggestion in comment, and there are some CI jobs to fix :upside_down_face: flake8 and pytest aren't happy

tests/common/test_models.py
Rob192: hoho ! yes it's nicer

doctr/models/predictor/pytorch.py
fg-mindee: Can you specify the command you run to get this error?

It will only run when Pytorch backend is enabled (set `USE_TORCH='1'` as env variable before running python)
The `.eval` method is inherited from `torch.nn.Module` so this part shouldn't be changed

tests/pytorch/test_models_recognition_pt.py
```diff
@@ -54,6 +58,34 @@ def test_reco_postprocessors(post_processor, input_shape, mock_vocab):
...
+    assert all(isinstance(val, str) and isinstance(conf, float) for val, conf in out)
+
+    # Dimension check
+    with pytest.raises(ValueError):
+        input_crop = (255 * np.random.rand(1, 128, 64, 3)).astype(np.uint8)
+        _ = predictor([input_crop])
+
+    return predictor
+
+
```
fg-mindee, line 88: same here

tests/pytorch/test_models_zoo_pt.py
```diff
@@ -1,9 +1,46 @@
 import numpy as np
 import pytest
+from test_models_detection_pt import test_detectionpredictor_pt  # noqa: F401
+from test_models_recognition_pt import test_recognitionpredictor_pt  # noqa: F401
```
fg-mindee, line 4: Let's just recreate the detection & recognition predictors directly here :ok_hand:

tests/pytorch/test_models_zoo_pt.py
```diff
@@ -1,9 +1,46 @@
...
 from doctr import models
-from doctr.io import Document
-from doctr.models.predictor import OCRPredictor
+from doctr.io import Document, DocumentFile
+from doctr.models.predictor.pytorch import OCRPredictor
+
+
+def test_ocrpredictor(
+    mock_pdf, test_detectionpredictor_pt, test_recognitionpredictor_pt  # noqa: F811
+):
```
fg-mindee, line 13: let's define the det_predictor and reco_predictor directly here :ok_hand:

Rob192:
OK @fg-mindee ! Now I got it for the testing !

fg-mindee (approved):
Looks good to me, thanks a lot @Rob192 ! @charlesmindee would you mind taking a look at it as well before we merge? :pray:

charlesmindee (dismissed):
Thanks for the PR !

doctr/utils/geometry.py
```diff
@@ -145,54 +145,104 @@ def rotate_abs_boxes(boxes: np.ndarray, angle: float, img_shape: Tuple[int, int]
...
+    preserve_origin_shape: bool = False,
 ) -> np.ndarray:
     """Rotate an image counterclockwise by an given angle.
 
     Args:
         image: numpy tensor to rotate
         angle: rotation angle in degrees, between -90 and +90
         expand: whether the image should be padded before the rotation
+        preserve_origin_shape: whether the image should be resized to the original image size after the rotation.
+            Only useful if expand is True.
```
fg-mindee, line 245: "if expand is set to True, resizes the final output to the original image size"

doctr/utils/geometry.py
```diff
@@ -201,25 +251,27 @@ def rotate_image(
...
-            # Pad width
-            if (rot_img.shape[0] / rot_img.shape[1]) > (image.shape[0] / image.shape[1]):
-                h_pad, w_pad = 0, int(rot_img.shape[0] * image.shape[1] / image.shape[0] - rot_img.shape[1])
-            # Pad height
-            else:
-                h_pad, w_pad = int(rot_img.shape[1] * image.shape[0] / image.shape[1] - rot_img.shape[0]), 0
-            rot_img = np.pad(rot_img, ((h_pad // 2, h_pad - h_pad // 2), (w_pad // 2, w_pad - w_pad // 2), (0, 0)))
-        # rescale
-        rot_img = cv2.resize(rot_img, image.shape[:-1][::-1], interpolation=cv2.INTER_LINEAR)
+        if preserve_origin_shape:
```
fg-mindee, line 264: Sorry I think I skipped that in my last review: the `preserve_origin_shape` should not condition the padding operation (controlled by `expand`), but only the scaling part.

So could you move this "if" right before the scaling part & unindent this block please? :pray:

fg-mindee (changes requested):
Sorry Robin, a minor thing I missed in my last review to adjust :sweat_smile: 
That shouldn't change much but especially on this feature it's important to have a good understanding of each argument influence on the final output!

doctr/utils/geometry.py
```diff
@@ -145,54 +145,103 @@ def rotate_abs_boxes(boxes: np.ndarray, angle: float, img_shape: Tuple[int, int]
...
+    expand: bool = False,
+    preserve_origin_shape: bool = False,
 ) -> np.ndarray:
     """Rotate an image counterclockwise by an given angle.
 
     Args:
         image: numpy tensor to rotate
         angle: rotation angle in degrees, between -90 and +90
-        expand: whether the image should be padded before the rotation
+        expand: whether the image should be padded before the rotation,
```
fg-mindee, line 243: no need for the final comma :)

doctr/models/predictor/pytorch.py
```diff
@@ -57,6 +64,13 @@ def forward(
         if any(page.ndim != 3 for page in pages):
             raise ValueError("incorrect input shape: all pages are expected to be multi-channel 2D images.")
 
+        origin_page_shapes = [page.shape[:2] if isinstance(page, np.ndarray) else page.shape[-2:] for page in pages]
+
+        # Detect document rotation and rotate pages
+        if self.straighten_pages:
+            origin_page_orientations = [estimate_orientation(page) for page in pages]
+            pages = [rotate_image(page, -angle, expand=True) for page, angle in zip(pages, origin_page_orientations)]
```
fg-mindee, line 72: I think you'll have to pass `preserve_origin_shape=True` over here to fix the failing test :)

fg-mindee (commented):
There seems to be a failing test: you might have to pass `preserve_origin_shape=True` where I added the comment :+1:

doctr/models/predictor/pytorch.py
Rob192: I don't get it :( this test passes on my computer ... I will have a look !

doctr/utils/geometry.py
Rob192: ok done

doctr/models/predictor/pytorch.py
Rob192: seems to be good now

doctr/utils/geometry.py
Rob192: done

Rob192: Yep I did that ! When doing this I realized that the padding was adding much offset. This was because we were using relative boxes with the rotation matrix although here we should be using absolute boxes. I changed that and now it works perfectly.

doctr/utils/geometry.py
```diff
@@ -145,54 +145,104 @@ def rotate_abs_boxes(boxes: np.ndarray, angle: float, img_shape: Tuple[int, int]
...
+    return mboxes
+
+
 def rotate_boxes(
-    boxes: np.ndarray,
+    loc_preds: np.ndarray,
     angle: float = 0.,
-    min_angle: float = 1.
+    min_angle: float = 1.,
+    orig_shape: Tuple[int, int] = (1, 1),
```
fg-mindee, line 183: With your last two commits, `orig_shape` has to be specified, so let's remove the default value :) 

Besides, it's rather dangerous to assign a default value to a immutable objects :sweat_smile:

doctr/utils/geometry.py
```diff
@@ -145,54 +145,104 @@ def rotate_abs_boxes(boxes: np.ndarray, angle: float, img_shape: Tuple[int, int]
...
+    # remaps box dimension for the destination image shape
+    mboxes[:, 2] = loc_preds[:, 2] * orig_width / dest_width
+    mboxes[:, 3] = loc_preds[:, 3] * orig_height / dest_height
+    return mboxes
+
+
 def rotate_boxes(
-    boxes: np.ndarray,
+    loc_preds: np.ndarray,
     angle: float = 0.,
```
fg-mindee, line 181: While being at it, I'd argue this doesn't make sense to make `angle` optional, could you remove the default value, and place `orig_shape` right afterwards? :pray:

fg-mindee (changes requested):
Making that last argument non-optional and we're good!

doctr/utils/geometry.py
Rob192: yes it's better

doctr/utils/geometry.py
Rob192: ok done

fg-mindee (approved):
All good :raised_hands: Thanks a lot!

Rob192:
Yeah ! Thanks a lot for your support in reviewing this !

PR #472, opened 2021-09-15
feat: Improved font resolution and page synthesis

fg-mindee:
This PR introduces the following modifications:
- refactored all font selection code into a `doctr.utils.fonts` module (for character generation & page synthesis)
- fixed typo `synthetize_page` --> `synthesize_page`
- improve page synthesis thanks to the new font module
- added a `synthesize` method to `Document` and `Page`
- added synthesis visualization sample in the README
- updated unittests
- updated documentation

As shown in the README, here is how it renders:
synthesis sample

Any feedback is welcome!

Issue #426, opened 2021-08-24
Integration with Hugging Face Hub


osanseviero:
Hi there,

First of all thanks for a awesome library!

At [Hugging Face](hf.co) we are collaborating with open source libraries in the ecosystem such as [spaCy](https://huggingface.co/spacy), [Sentence Transformers](https://huggingface.co/blog/sentence-transformers-in-the-hub), Timm, ESPNet, and [more](https://huggingface.co/docs/hub/libraries) in order to implement integration in the models hub.

The idea is to make it as easy as possible for your users to try out the models and to share their models as well. I think it would be very great to have some integration with DocTR. Users would get the following benefits:

- Free hosting of models
- Built-in file versioning
- Hosted Inference API and widgets to try out the models (you can find examples of all our widgets [here](https://huggingface-widgets.netlify.app/), and [here](https://huggingface.co/explosion/en_textcat_goemotions?text=This+is+very+cool%21) is an example in a model card.
- Hub features: Code snippets, filters to find models, and other features to help with discoverability

We have [documentation](https://huggingface.co/docs/hub/adding-a-library) on the integration process, but in summary, we usually have 3 things:
  * Downstream support: allow users to download pretrained models from the Hub
  * Upstream support: allow users to upload models to the Hub
  * Hub features: code snippets, widgets, Inference API, etc.

In this case, I think the first and third points could be a nice first step. I would be happy to hear your thoughts.

cc @LysandreJik @NielsRogge

fg-mindee:
Hi @osanseviero :wave: 

Thank you a lot for your message! We certainly appreciate the contribution of HuggingFace to the dev community, so your suggestion is flattering :pray:

I agree that some integration would most likely be very beneficial to the open source world! Especially considering the possibilities of combining an OCR block from DocTR with a NLP block from Huggingface :ok_hand:

There is only one aspect that we need to discuss internally: while Mindee does provide open source material, we do offer commercial services for the orchestration of APIs specialized in OCR-related tasks (similar to your organization tier for your inference API I think). Currently, I do not know the full range of possibilities for Huggingface users if we move forward with this integration (i.e. if future users could take a Huggingface subscription for your inference API on DocTR models, we need to figure how it would work in the long term) :sweat_smile:

That being said, for free tier / non-commercial usage, I fully agree that integrating DocTR in your Hub would be an exciting idea :+1: 

Happy to discuss this further!

osanseviero:
Hi @fg-mindee!

[...]

PR #492, opened 2021-09-27
Misc fixed to Streamlit demo

osanseviero:
Hello!

I built a Space demo hosting your Streamlit example (link at https://huggingface.co/spaces/osanseviero/doctr). It works great and is much faster than I expected.

I did face some issues during the development. I hope this PR fixes them:

1. You suggest to install `doctr` in the `requirements` but I think that's actually an unrelated package, it should be `python-doctr`.
2. But then `out.pages[0].synthesize()` threw an error saying `Page` class has no `synthesize` method. This was introduced in #472 so I don't think it's in the recent release. 

Given 1 and 2, and that #472 will probably be soon in a release, maybe we could change `requirements.txt` to install `python-doctr[tf]` instead of installing directly from GitHub. WDYT?

### DIFF
--- a/demo/requirements.txt
+++ b/demo/requirements.txt
@@ -1,4 +1,2 @@
--e git+https://github.com/mindee/doctr.git#egg=doctr
-tensorflow
-tensorflow-addons
+-e git+https://github.com/mindee/doctr.git#egg=doctr[tf]
 streamlit>=0.65.0
--- a/demo/requirements.txt
+++ b/demo/requirements.txt
@@ -1,2 +1,2 @@
-doctr
+doctr[tf]
 streamlit>=0.65.0
--- a/demo/requirements.txt
+++ b/demo/requirements.txt
@@ -1,2 +1,4 @@
-doctr[tf]
+-e git+https://github.com/mindee/doctr.git#egg=doctr
+tensorflow
+tensorflow-addons
 streamlit>=0.65.0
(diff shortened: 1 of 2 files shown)


demo/requirements.txt
```diff
@@ -1,2 +1,4 @@
-doctr
+-e git+https://github.com/mindee/doctr.git#egg=doctr
+tensorflow
+tensorflow-addons
```
fg-mindee, line 3: Thanks!
In the same spirit, I would suggest so that it stays robust to future evolutions of the dependencies:
```
-e git+https://github.com/mindee/doctr.git#egg=doctr[tf]
```

fg-mindee (commented):
Thanks a lot for the demo & the bug fix @osanseviero :pray: 

To answer your questions:
- we have an upcoming release in a few days so the page synthesis will be on a stable release!
- you're correct about the package ref in the demo, but for the same reason than your first point, I think it might be best to stay aligned with the main branch on Github, don't you think?

I added a comment to use the extras to have a fully compatible TF environment, let me know what you think!

demo/requirements.txt
osanseviero: Alright, good idea! Done!

osanseviero:
Thanks for your input! I changed the requirements as you suggested and removed the TF requirements since they are part of the extras.

I'm glad you liked the demo!

fg-mindee (approved):
Much appreciated Omar!

fg-mindee:
@osanseviero by any chance, is there a markdown badge or something similar for huggingface spaces demo?
That could be a cool thing to include in the README in another PR :grin:

fg-mindee:
For reference: cf #426

PR #497, opened 2021-09-28
fix: Fixed inconsistency of export when nothing is detected

fg-mindee:
So I noticed that when the predictor wasn't detecting anything, the number of pages in the export wasn't matching the input. Even if it doesn't manage to detect anything, the export should reflect the correct number of pages.
- refactors the `DocumentBuilder` to ensure that localization predictions are the same length as the text predictions, as well as the number of pages
- fixed zero-sized crop filtering (on multi-pages cases)
- refactored crop preparation & prediction processing for all framework-specific predictors

Using this image
euro

with the following snippet:
```python
predictor = ocr_predictor(pretrained=True)

doc = DocumentFile.from_images('/home/fg/Downloads/euro.png')
out = predictor(doc)
print(out.export())
```

used to yield:
```
{'pages': []}
```

and now yields:
```
{'pages': [{'page_idx': 0,
   'dimensions': (960, 640),
   'orientation': {'value': None, 'confidence': None},
   'language': {'value': None, 'confidence': None},
   'blocks': []}]}
```

I checked with a sample PDF, both TF & PyTorch are yielding almost identical results.

Any feedback is welcome!

### DIFF
--- a/doctr/models/builder.py
+++ b/doctr/models/builder.py
@@ -250,36 +250,34 @@ class DocumentBuilder(NestedObject):
     def __call__(
         self,
         boxes: List[np.ndarray],
-        word_preds: List[Tuple[str, float]],
+        text_preds: List[List[Tuple[str, float]]],
         page_shapes: List[Tuple[int, int]]
     ) -> Document:
         """Re-arrange detected words into structured blocks
 
         Args:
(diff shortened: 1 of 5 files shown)


charlesmindee (approved):
Thanks!

charlesmindee (approved):
Thanks!

PR #498, opened 2021-09-29
chore: Split common, torch & tensorflow unittests

fg-mindee:
For clarity and CI speedups, this PR splits the unittests of `common`, `pytorch`, and `tensorflow`.
This brings several advantages:
- many unittests will now be done in parallel instead of making `pytest-tf` slower
- when an error occurs, it's easier to spot right now if it involves any of the framework

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #500, opened 2021-09-29
feat: Added ancient Greek vocab

fg-mindee:
This PR simply adds the ancient Greek alphabet to the supported vocabs.

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #502, opened 2021-09-29
feat: Added option to select vocab for training

fg-mindee:
This PR adds the option to select a different vocab for training (character classification and text recognition).

Any feedback is welcome!


charlesmindee (approved):
thanks!

PR #519, opened 2021-10-01
docs: Restructured documentation

fg-mindee:
- removes the mention that PyTorch is in beta in the README
- in the documentation, split benchmark & export model from actual docstrings: "Using DocTR" section doesn't include any function/class documentation but explains how some features work and provides benchmarks. Hyperlinks were added so that users can easily access the signature.

Here is the new navigation menu:
image

Please note that this doesn't include the PyTorch benchmark for now!

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #548, opened 2021-10-26
perf: resolve both lines and blocks by default when building a doc

### LINKED ISSUE #547: set resolve_lines and resolve_blocks to True
## 🐛 Bug

As discussed in #512 it needs an improvement for the `sort_boxes ` to set the `resolve_lines` and `resolve_blocks` in `builder.py `to True as default

In fact of this the `.render()` and `.export_as_xml()` results are not correct 

## To Reproduce

1. test with any document image which has multible text blocks/lines or where the document is slightly crooked in the picture, for example photographed by a mobile phone

## Expected behavior

Correct sorted boxes left-right / top-bottom that the lines and blocks can be resolved correctly

## Environment

-DocTR version: 0.4.1a0
-TensorFlow version: N/A
-PyTorch version: 1.9.1 (torchvision 0.10.1)
-OpenCV version: 4.4.0
-OS: Ubuntu 20.04.3 LTS
-Python version: 3.8
-Is CUDA available (TensorFlow): N/A
-Is CUDA available (PyTorch): No
-CUDA runtime version: Could not collect
-GPU models and configuration: Could not collect
-Nvidia driver version: Could not collect
-cuDNN version: Could not collect

charlesmindee:
This PR sets `resolve_lines` and` resolve_blocks` to True in the builder as suggested and benchmarked in #547
Any feedback is welcome!

closes #547

### DIFF
--- a/doctr/models/builder.py
+++ b/doctr/models/builder.py
@@ -178,7 +178,6 @@ class DocumentBuilder(NestedObject):
             ), axis=-1
         )
         # Compute clusters
-        print(box_features)
         clusters = fclusterdata(box_features, t=0.1, depth=4, criterion='distance', metric='euclidean')
 
         _blocks: Dict[int, List[int]] = {}
@@ -215,7 +214,7 @@ class DocumentBuilder(NestedObject):
         if self.resolve_lines:
             lines = self._resolve_lines(boxes[:, :-1])
             # Decide whether we try to form blocks
-            if self.resolve_blocks:
+            if self.resolve_blocks and len(lines) > 1:
                 _blocks = self._resolve_blocks(boxes[:, :-1], lines)
             else:
                 _blocks = [lines]


fg-mindee:
@Rob192 we're gonna clean the entire interface with the rotation option for the 0.5.0 release

But I don't think the error was caused by this PR, rather by the previous behaviour (which we'll need to fix). But I might be missing something?

Rob192:
Hello @fg-mindee ! I think is comes from the fact that before this PR, `resolve_lines` and `resolve_blocks` were set to `False`. So we were not using this part of the code. I am not sure of what are your thoughts concerning the management of the boxes for the future. Currently I am having a hard time figuring out what type of boxes are returned by the different functions and I feel that testing that they have a length of 4 or 5 is not sufficient to discriminate between the `straight` or `rotated` boxes.

fg-mindee:
> Hello @fg-mindee ! I think is comes from the fact that before this PR, `resolve_lines` and `resolve_blocks` were set to `False`. So we were not using this part of the code. I am not sure of what are your thoughts concerning the management of the boxes for the future. Currently I am having a hard time figuring out what type of boxes are returned by the different functions and I feel that testing that they have a length of 4 or 5 is not sufficient to discriminate between the `straight` or `rotated` boxes.

Yes, we have (hopefully) clarified box operations with 2 new flags, that were integrated in the high-level API in #632. Could you let us know if the error is still happening on the main branch if you install it from source?

PR #552, opened 2021-10-28
make resizing flexible

felixdittrich92:
feat: this request makes the resizing in Documentfile.from_images flexible it is enough to specify width or height, the function keeps ratio of the image it is only an improvement you can still specify both values

Currently no open ticket for this Request


fg-mindee:
> 2. Your variant is very static right now, assuming the user only wants to change the width of the images, then he would have to calculate the height himself in order to keep the ratio, just the other way around for the height and width. In my variant, this is already taken over, so that it is sufficient to specify one of the two values. Otherwise, of course, you still have the option of specifying both values.

Well, we can't offer absolute freedom to users :sweat_smile: Most of the computer vision-related libraries employ tuple of ints to specify shapes. I think we should stick to this

> 3. you mean something like this: target_size: Optional[Tuple[Int, Int]] = tuple(None, None)  -> (height=300, width=None) ?

I meant `Optional[Union[int, Tuple[int, int]]] = None`. If it's an int, we say it will be the size of the longest image side (or shortest for instance, but we need a convention). If it's a tuple, we keep to existing behavior

felixdittrich92:
@fg-mindee 
By stick do you mean no changes ? or that we keep target_size as it is and work with the given values ?
like (1200, None) does only change the height and keep image ratio on width?

felixdittrich92:
@fg-mindee 
By stick do you mean no changes ? or that we keep target_size as it is and work with the given values ?
like (1200, None) does only change the height and keep image ratio on width?

fg-mindee:
> @fg-mindee By stick do you mean no changes ? or that we keep target_size as it is and work with the given values ? like (1200, None) does only change the height and keep image ratio on width?

I mean sticking to tuple of ints, it's almost an implicit standard in the computer vision community :)

PR #571, opened 2021-11-01
fix: Fixed portuguese entry in VOCAB

fmobrj:
Refactored portuguese entry in VOCAB as suggested by @fg-mindee.


fg-mindee (approved):
Thanks for the fix!

fmobrj:
> Thanks for the fix!

You are welcome.

PR #574, opened 2021-11-02
Feature resize (Reopening #552)

felixdittrich92:
As discussed in #552


fg-mindee:
Better to pull/sync your `main` branch before opening another PR to get a clean git history :ok_hand: 
But I'm not sure I understand what you are proposing here?

In #552, we concluded that we would keep the current interface with a tuple of int (not optional int, actual int). So this is already covered. The other alternatives would be confusing to most people using computer vision libraries :man_shrugging:

felixdittrich92:
@fg-mindee 
Then I got it wrong, I assumed that we could only keep output_size as it is and still implement the feature. It would be nice not to be dependent on having to specify both values. for example ` (1200, None)` change height and keep ratio, `(None, 1200)` change only width and `(1200, 1200)` is the current default behaviour.
?

fg-mindee:
My bad then! I suggested at some point to have either a tuple of 2 ints, or a single int (and then resolving dynamically the smaller side). But I think it's best to keep the current interface for now at least. We're not talking about the resizing done to PIL image or tensors here but to the PDF rendering. And I would argue it needs to be carefully handled, so for now perhaps it would be best to keep the current API :)

felixdittrich92:
@fg-mindee 
then i will close it for the moment :)

PR #583, opened 2021-11-04
feat: added DocArtefacts dataset

SiddhantBahuguna:
- Implemented class wrapper for artefact detection dataset

Any feedback is welcome!


fg-mindee:
Looks to me like we're missing two things:
- [ ] Added the corresponding entry in the documentation
- [ ] Adding corresponding unittest

fg-mindee (changes requested):
Thanks for the PR! Looks to me like we're missing two things:
- [ ] Added the corresponding entry in the documentation
- [ ] Adding corresponding unittest

fg-mindee (approved):
Thanks!

PR #586, opened 2021-11-05
docs: Fixed docstring and export as xml dim bug

felixdittrich92:
There was a wrong description in the docstring which results also in the wrong dim order in export_as_xml 
now the parser example runs also correct


fg-mindee (commented):
Thanks would you mind specifying the dimension order in the docstrings of `doctr/utils/visualization.py` as well? (so that we avoid this to happen again)

fg-mindee (approved):
Looks good to me, thanks Felix :pray:

PR #595, opened 2021-11-09
feat: :sparkles: add det/reco_predictor arch in OCRPredictor.__repr__

RBMindee:
Previously, if model is a TF version of OCRPredictor, `print(model) `shows only `OCRPredictor() ` which is not too informative. It now prints the architecture of the underlying det_predictor and reco_predictor.

### DIFF
--- a/doctr/models/predictor/tensorflow.py
+++ b/doctr/models/predictor/tensorflow.py
@@ -26,6 +26,7 @@ class OCRPredictor(NestedObject, _OCRPredictor):
         det_predictor: detection module
         reco_predictor: recognition module
     """
+    _children_names = ['det_predictor', 'reco_predictor']
 
     def __init__(
         self,


doctr/models/predictor/tensorflow.py
```diff
@@ -40,6 +40,7 @@ def __init__(
         self.reco_predictor = reco_predictor
         self.doc_builder = DocumentBuilder(export_as_straight_boxes=export_as_straight_boxes)
         self.assume_straight_pages = assume_straight_pages
+        self._children_names = ['det_predictor', 'reco_predictor']
```
fg-mindee, line 43: Indeed, would you mind moving this to a class attribute rather than instance attribute? (i.e. at line 29 similarly to other TF components)

fg-mindee (commented):
Nice catch, thanks a lot! Would you mind relocating the children name definition to class attributes please? :pray:

fg-mindee (approved):
Thanks a lot Remy!

RBMindee:
What does the check that fails mean ?

fg-mindee:
> What does the check that fails mean ?

That one is a nasty lack of support by TensorFlow of grouped convolution, I'll have to work on that but I left it as it might become important to check when this is finally supported by this major DL framework :sweat_smile:

PR #590, opened 2021-11-08
docs: Fixed README instruction for page synthesis

fg-mindee:
This PR updates the instructions for page synthesis in the README that were failing.

Any feedback is welcome!

PR #596, opened 2021-11-09
docs: :memo: update readme and help in analyse.py

RBMindee:
Slight fixes and modification of the doc:

- fixed typo in readme suggested code using demonstrating synthesize
- Added the mention to tensorflow-addons
- Added model names in front of papers
- Mentionned in help of the noblock option of analyze.py that it's only used for CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,11 +105,10 @@ For Windows users, those dependencies are included in GTK. You can find the late
 
 You can then install the latest release of the package using [pypi](https://pypi.org/project/python-doctr/) as follows:
 
-> :warning: Notice that the basic installation is not standalone, as it does not provide a deep learning framework, which is required for the package to run.
-
 ```shell
 pip install python-doctr
 ```
+> :warning: Please note that the basic installation is not standalone, as it does not provide a deep learning framework, which is required for the package to run.
 
 We try to keep framework-specific dependencies to a minimum. You can install framework-specific builds as follows:
 
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ For examples & further details about the export format, please refer to [this se
 
 ### Prerequisites
 
-Python 3.6 (or higher) and [pip](https://pip.pypa.io/en/stable/) are required to install docTR. Additionally, you will need to install at least one of [TensorFlow](https://www.tensorflow.org/install/) or [PyTorch](https://pytorch.org/get-started/locally/#start-locally). If you use tensorflow, you will need to install the additional package `tensorflow-addons`
+Python 3.6 (or higher) and [pip](https://pip.pypa.io/en/stable/) are required to install docTR. 
 
 Since we use [weasyprint](https://weasyprint.readthedocs.io/), you will need extra dependencies if you are not running Linux.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -69,7 +69,7 @@ Or even rebuild the original document from its predictions:
 ```python
 
-plt.imshow(result.synthesize()); plt.axis('off'); plt.show()
+plt.imshow(result.synthesize()[0]); plt.axis('off'); plt.show()
```
fg-mindee, line 72: Was already fixed in #590 :sweat_smile: Mind reversing that change please? :pray:

README.md
```diff
@@ -89,7 +89,7 @@ For examples & further details about the export format, please refer to [this se
 
 ### Prerequisites
 
-Python 3.6 (or higher) and [pip](https://pip.pypa.io/en/stable/) are required to install docTR. Additionally, you will need to install at least one of [TensorFlow](https://www.tensorflow.org/install/) or [PyTorch](https://pytorch.org/get-started/locally/#start-locally).
+Python 3.6 (or higher) and [pip](https://pip.pypa.io/en/stable/) are required to install docTR. Additionally, you will need to install at least one of [TensorFlow](https://www.tensorflow.org/install/) or [PyTorch](https://pytorch.org/get-started/locally/#start-locally). If you use tensorflow, you will need to install the additional package `tensorflow-addons`
```
fg-mindee, line 92: What do you think about always specifying the extra with the pip install command?
Because here we say that TF needs `tensorflow-addons` but say tomorrow there are 5 extra TF-specific deps, we don't want to leave the user alone to do it :/ 

But if they run `pip install python-doctr[tf]` that will automatically install the remaining dependencies. What do you think?

fg-mindee (changes requested):
thanks!
About the page synthesis instruction, that was fixed in #590! I added some comment about the tf-addons installation :)

README.md
RBMindee: I find it a bit awkward that you have "you should install tensorflow first" and "you can then install the latest release with pip install python-doctr" while it does not produce a working installation. I understand that you propose a solution immediately after,  but doesn't it mean that python-doctr setup alone should always need an argument ?

Anyway, that's a very minor issue and the main correction was the error in the example, so I guess we can cancel this PR if you do not like this correction ^^"

fg-mindee: Yup I would argue it should, and thus we can remove the section about "you need to install a framework first" :+1: 

Oh don't take it the wrong way, I was thinking about trying to harmonize installation instructions. I'm only trying to find the best option to tackle the underlying installation issue here

RBMindee: No worries! I was just saying that messing with setup tool is probably a bit more than twiddling the doc, but we can keep it in the PR if you want. If I understood you correctly, setup_tools should throw an error if an argument is not specified ?

RBMindee: Actually, thinking a bit more, I guess it makes sense to have a framework agnostic version, even if it's not standalone. We could simply have a warning in the installation instruction. What do you think about the current state of things ?

fg-mindee: Yup I was only suggesting to modify the instructions in the README here :) 
We'll have to do the same with the instructions in the documentation but that can wait for another PR

README.md
```diff
@@ -105,11 +105,13 @@ For Windows users, those dependencies are included in GTK. You can find the late
 
 You can then install the latest release of the package using [pypi](https://pypi.org/project/python-doctr/) as follows:
 
+> :warning: Notice that the basic installation is not standalone, as it does not provide a deep learning framework, which is required for the package to run.
+
```
fg-mindee, line 109: One last adjustment: could you rephrase "Notice" to "Please note" ? :pray: 
And move the warning right after the `pip install python-doctr` command?

fg-mindee (commented):
Last adjustments and we're good!

fg-mindee (approved):
Thanks a lot for the edits!

PR #600, opened 2021-11-09
feat: Added new architectures to model selection in demo

fg-mindee:
- adds `db_mobilenet_v3_large` and `crnn_mobilenet_v3_small` as new options for model selection
- updates the column creation to avoid warnings with latest streamlit versions

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #601, opened 2021-11-09
feat: Split DocArtefacts into subsets and updated its class mapping

fg-mindee:
- updates the URL of the zip which now contains a train and val subset
- updates the labels target: rather than being a list of string, it will return a `numpy.array` of class indices
- updates the docstring of the class

Any feedback is welcome!


charlesmindee (approved):
Thanks!

SiddhantBahuguna (commented):
1) Should we add the collate function here?
2) Also, we need to obtain absolute box coordinates. Right now we use relative ones.

fg-mindee:
> 1. Should we add the collate function here?

Oh yeah, you're right, I'll change this

> 2. Also, we need to obtain absolute box coordinates. Right now we use relative ones.

Indeed

fg-mindee:
For the collate function, actually I just checked and it's already working @SiddhantBahuguna :
```python
from doctr.datasets import DocArtefacts
from doctr.datasets import DataLoader

ds = DocArtefacts(train=True, download=True)
train_loader = DataLoader(ds, batch_size=2)
train_iter = iter(train_loader)
x, targets = next(train_iter)
```
And the results look satisfactory to me:
```python
# Check shape of input
print(x.shape)
```
```
TensorShape([2, 1024, 800, 3])
```

```python
print(len(targets))
print(targets[0])
```

```
2
{'boxes': array([[0.94625   , 0.39746094, 0.99375   , 0.4345703 ],
       [0.38375   , 0.5957031 , 0.4275    , 0.6269531 ],
       [0.41875   , 0.28027344, 0.52875   , 0.36621094],
       [0.39625   , 0.36816406, 0.6675    , 0.57910156],
       [0.26625   , 0.5332031 , 0.2975    , 0.5576172 ],
       [0.20125   , 0.1953125 , 0.2575    , 0.23925781],
       [0.105     , 0.00976562, 0.17125   , 0.06152344],
       [0.15875   , 0.078125  , 0.2425    , 0.14355469],
       [0.10125   , 0.5292969 , 0.25875   , 0.6796875 ],
       [0.73625   , 0.7138672 , 0.85375   , 0.8261719 ]], dtype=float32), 'labels': array([1, 2, 1, 1, 1, 1, 3, 3, 4, 4])}
```

SiddhantBahuguna (approved):
The output is following our expectations. Collate problem is resolved now. Thanks:)

PR #604, opened 2021-11-10
feat: Added support of AMP to all PyTorch training scripts

fg-mindee:
- added option to run the training in AMP
- updated information logged into W&B
- fixed sample plotting in trainings
- fixed compatibility of DBNet with data type different from fp32

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #609, opened 2021-11-12
docs: Specified that the API template only supports images for now

fg-mindee:
Following up on some QA feedback, this PR introduces the following modifications:
- specifies that all routes support only images and not PDF for now
- cleaned the expected route status
- improved the CI check: it used to check that the base URL is accessible, and now it will check whether it can access the swagger/docs

Any feedback is welcome!

cc @RBMindee


charlesmindee (approved):
Thanks for the PR!

PR #611, opened 2021-11-12
fix: :bug: deprecated streamlit version in demo app

charlesmindee:
This PR updates the min. Streamlit version to 1.0.0 (major released in october), older versions are deprecated with the use of columns.

Any feedback is welcome!


fg-mindee (approved):
Sounds good to me!

PR #612, opened 2021-11-12
docs: Added README specific for the API with route examples

fg-mindee:
Following on some QA feedback, this PR adds a dedicated README for the API template. It provides curated examples of inputs and outputs on each route.

Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR !

PR #613, opened 2021-11-12
chore: Added checkpoint format to gitignore

fg-mindee:
This PR simply adds checkpoint files to gitignore


charlesmindee (approved):
Thanks!

Issue #587, opened 2021-11-05
[datasets] Extend the range of public datasets supported in docTR


fg-mindee:
Currently, we support `FUNSD`, `CORD` and `SROIE` but we should look at extending the range of supported datasets. Among others, we could include handwritten, and in-the-wild situations.

Here is a list of datasets you can usually find in OCR-related benchmarks:
- [x] IIIT-5k (https://cvit.iiit.ac.in/research/projects/cvit-projects/the-iiit-5k-word-dataset) #589
- [x] SVT (http://vision.ucsd.edu/~kai/svt/) #597 #620
- [x] IC03 (http://www.iapr-tc11.org/mediawiki/index.php?title=ICDAR_2003_Robust_Reading_Competitions) #653
- [x] IC13 (http://dagdata.cvc.uab.es/icdar2013competition/?ch=2&com=downloads) #662
- [x] SVHN (http://ufldl.stanford.edu/housenumbers/) #634
- [x] SynthText (https://github.com/ankush-me/SynthText) #624
- [ ] IMGUR5K (https://github.com/facebookresearch/IMGUR5K-Handwriting-Dataset) #785 

Of course, the list goes on

felixdittrich92:
@fg-mindee 
handwritten:
https://fki.tic.heia-fr.ch/databases/iam-handwriting-database
was also used for TrOCR finetuning

felixdittrich92:
TextOCR dataset
https://textvqa.org/textocr/dataset/

felixdittrich92:
@fg-mindee 
i will start with the IIIT-5K dataset can you assigne me to this :smiley: ?

fg-mindee:
Sure, you can go ahead and I see that you already opened a PR :)

felixdittrich92:
Char74k dataset
[http://www.ee.surrey.ac.uk/CVSSP/demos/chars74k/](http://www.ee.surrey.ac.uk/CVSSP/demos/chars74k/)

felixdittrich92:
TotalText dataset
[https://github.com/cs-chan/Total-Text-Dataset/tree/master/Dataset](https://github.com/cs-chan/Total-Text-Dataset/tree/master/Dataset)

felixdittrich92:
@fg-mindee 
I´m not able to grab the IAM dataset (have registered but after this nothing) so i would say lets change this to IMGUR5K dataset i will prepare this (it´s only a generator script) but we need to upload this (IAM would be the same) 
[...]

PR #624, opened 2021-11-15
SynthText dataset integration

felixdittrich92:
This PR adds the Synthtext dataset 

cf. #587

Problems:

- the unpack part needs a bit longer  (>800k data) so i think we have to implement a check that unpacking is finished before continue
- same for the hash check this will end in a MemoryError

wdyt @fg-mindee how to handle? :)
suggestions are welcome :hugs: 

- [x] tests
- [x] bypass the hash check


doctr/datasets/synthtext.py
```diff
@@ -0,0 +1,103 @@
...
+        rotated_bbox: whether polygons should be considered as rotated bounding box (instead of straight ones)
+        **kwargs: keyword arguments from `VisionDataset`.
+    """
+
+    URL = 'https://thor.robots.ox.ac.uk/~vgg/data/scenetext/SynthText.zip'
+    SHA256 = '28ab030485ec8df3ed612c568dd71fb2793b9afbfa3a9d9c6e792aef33265bf1'
+
+    def __init__(
+        self,
+        train: bool = True,
```
fg-mindee, line 40: There is no train/val split?
If so, we should make it ourselves (say the first 90% goes into train, 10% for val)

doctr/datasets/synthtext.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Example::
+        >>> # NOTE: This dataset has currently no train/test split
+        >>> from doctr.datasets import SynthText
+        >>> data_set = SynthText(download=True)
+        >>> img, target = data_set[0]
+
+    Args:
+        sample_transforms: composable transformations that will be applied to each image
+        rotated_bbox: whether polygons should be considered as rotated bounding box (instead of straight ones)
```
fg-mindee, line 31: missing arg description for train

doctr/datasets/synthtext.py
```diff
@@ -0,0 +1,103 @@
...
+                mat_data['imnames'][0],
+                mat_data['wordBB'][0],
+                mat_data['txt'][0]
+        ), desc='Load SynthText', total=len(mat_data['imnames'][0])):
+
+            # File existence check
+            if not os.path.exists(os.path.join(tmp_root, img_path[0])):
+                raise FileNotFoundError(f"unable to locate {os.path.join(tmp_root, img_path[0])}")
+
+            labels = _text_to_words(txt)
```
fg-mindee, line 66: let's avoid other functions when it's not needed:
```python
labels = txt.tolist()
```

doctr/datasets/synthtext.py
```diff
@@ -0,0 +1,103 @@
...
+
+            self.data.append((img_path[0], dict(boxes=np.asarray(box_targets, dtype=np_dtype), labels=labels)))
+
+        self.root = tmp_root
+
+
+def _text_to_words(txt: np.ndarray) -> List[str]:
+    """Convert np.str-Array to list of str."""
+    line = '\n'.join(txt)
+    return line.split()
```
fg-mindee, line 84: can be removed

doctr/datasets/synthtext.py
```diff
@@ -0,0 +1,103 @@
...
+
+            labels = _text_to_words(txt)
+            word_boxes = word_boxes.transpose(2, 1, 0) if word_boxes.ndim == 3 else np.expand_dims(word_boxes, axis=0)
+
+            if rotated_bbox:
+                # x_center, y_center, w, h, alpha = 0
+                box_targets = [_compute_rotated_box(pts) for pts in word_boxes]
+            else:
+                # xmin, ymin, xmax, ymax
+                box_targets = [_compute_straight_box(pts) for pts in word_boxes]  # type: ignore[misc]
```
fg-mindee, line 74: no need:
```python
box_targets = np.concatenate((word_boxes.min(axis=1), word_boxes.max(axis=1)), axis=1)
```

doctr/datasets/synthtext.py
```diff
@@ -0,0 +1,103 @@
...
+            # File existence check
+            if not os.path.exists(os.path.join(tmp_root, img_path[0])):
+                raise FileNotFoundError(f"unable to locate {os.path.join(tmp_root, img_path[0])}")
+
+            labels = _text_to_words(txt)
+            word_boxes = word_boxes.transpose(2, 1, 0) if word_boxes.ndim == 3 else np.expand_dims(word_boxes, axis=0)
+
+            if rotated_bbox:
+                # x_center, y_center, w, h, alpha = 0
+                box_targets = [_compute_rotated_box(pts) for pts in word_boxes]
```
fg-mindee, line 71: Better to vectorize this:
```python
mins = word_boxes.min(axis=1)
maxs = word_boxes.max(axis=1)
box_targets = np.concatenate(((mins + maxs) / 2, maxs - mins, np.zeros(word_boxes.shape[0])), axis=1)
```

doctr/datasets/synthtext.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def _compute_rotated_box(pts: np.ndarray) -> Tuple[float, float, float, float, int]:
+    # pts: Nx2
+    x = np.min(pts[:, 0])
+    y = np.min(pts[:, 1])
+    width = np.max(pts[:, 0]) - x
+    height = np.max(pts[:, 1]) - y
+    # x_center, y_center, w, h, alpha = 0
+    return x + width / 2, y + height / 2, width, height, 0
```
fg-mindee, line 103: Can be removed

doctr/utils/data.py
```diff
@@ -37,10 +37,14 @@ def _urlretrieve(url: str, filename: Union[Path, str], chunk_size: int = 1024) -
...
-        sha_hash = hashlib.sha256(f.read()).hexdigest()
+    try:
+        with open(file_path, 'rb') as f:
+            sha_hash = hashlib.sha256(f.read()).hexdigest()
 
-    return sha_hash[:len(hash_prefix)] == hash_prefix
+        return sha_hash[:len(hash_prefix)] == hash_prefix
+    # FIX: for big datasets
+    except MemoryError:
+        return True
```
fg-mindee, line 47: We can't afford that, if the dataset causes hash issues, we can consider avoiding the hash check, but this function should not be modified

fg-mindee (changes requested):
Thanks! I added some comments

doctr/datasets/synthtext.py
felixdittrich92: @fg-mindee does not work correctly but now:
`labels = '\n'.join(txt).split()`

felixdittrich92: ok ?

doctr/datasets/synthtext.py
felixdittrich92:
```
if train:
            self.data = self.data[:int(len(self.data) * 0.9)]
        else:
            self.data = self.data[int(len(self.data) * 0.9):]
```
@fg-mindee 
ok ?

felixdittrich92:
@fg-mindee 
Dataset size: ~41 GB

doctr/datasets/synthtext.py
fg-mindee: still missing

doctr/datasets/synthtext.py
fg-mindee: well let's avoid unnecessary iterations:
- if train --> at 90% we break the iteration
- if val --> we start iterating only at 90%

fg-mindee (commented):
Another round of comments :sweat_smile:

felixdittrich92:
@fg-mindee 
I added to bypass the hash check and removed the tests for this dataset ftm also the train/val split now before iterating.
Already any decision about the dataset test ? :)

felixdittrich92:
@fg-mindee 
after attaching the example can you trigger integration again pls ? :smiley:

doctr/utils/data.py
```diff
@@ -49,6 +49,7 @@ def download_from_url(
     hash_prefix: Optional[str] = None,
     cache_dir: Optional[str] = None,
     cache_subdir: Optional[str] = None,
+    check_hash: bool = True,
```
fg-mindee, line 52: As mentioned before, we can't forward to change this file :sweat_smile: 
If at some point, we cannot perform the check, then we just don't do it upstream but this function should not be changed. If the name of the file doesn't include any hash and it's not passed, the check won't be done, so that should be enough :ok_hand:

doctr/datasets/datasets/base.py
```diff
@@ -89,7 +91,7 @@ def __init__(
         if not os.path.exists(archive_path) and not download:
             raise ValueError("the dataset needs to be downloaded first with download=True")
 
-        archive_path = download_from_url(url, file_name, file_hash, cache_subdir='datasets')
+        archive_path = download_from_url(url, file_name, file_hash, check_hash=check_hash, cache_subdir='datasets')
```
fg-mindee, line 94: cf my other comment, let's revert the changes on this file :pray:

tests/tensorflow/test_datasets_tf.py
```diff
@@ -8,26 +8,28 @@
...
+        ['SynthText', True, [512, 512], 27, False, False],
+        ['SynthText', False, [512, 512], 3, True, False],
     ],
 )
-def test_dataset(dataset_name, train, input_size, size, rotate):
+def test_dataset(dataset_name, train, input_size, size, rotate, download):
 
     ds = datasets.__dict__[dataset_name](
-        train=train, download=True, sample_transforms=Resize(input_size), rotated_bbox=rotate,
+        train=train, download=download, sample_transforms=Resize(input_size), rotated_bbox=rotate,
```
fg-mindee, line 32: cf the comments above

tests/pytorch/test_datasets_pt.py
```diff
@@ -18,26 +18,28 @@ def test_visiondataset():
...
+        ['SynthText', True, [512, 512], 27, False, False],
+        ['SynthText', False, [512, 512], 3, True, False],
     ],
 )
-def test_dataset(dataset_name, train, input_size, size, rotate):
+def test_dataset(dataset_name, train, input_size, size, rotate, download):
 
     ds = datasets.__dict__[dataset_name](
-        train=train, download=True, sample_transforms=Resize(input_size), rotated_bbox=rotate,
+        train=train, download=download, sample_transforms=Resize(input_size), rotated_bbox=rotate,
```
fg-mindee, line 42: better to do an exception to create the `ds` object like
```python
if dataset_name.lower() == "synthtext":
...
```

The idea here is the following:
- if it's synthtext, we set `download=False`
- before calling the constructor, we download the subsample, put it in the cache
- when we call the constructor, we need to monkeypatch the expected hash

fg-mindee (commented):
Thanks for the edits! I suggest waiting a bit, because it's gonna be troublesome to get the CI to work nicely with this

doctr/datasets/synthtext.py
```diff
@@ -0,0 +1,84 @@
...
+
+    Args:
+        train: whether the subset should be the training one
+        sample_transforms: composable transformations that will be applied to each image
+        rotated_bbox: whether polygons should be considered as rotated bounding box (instead of straight ones)
+        **kwargs: keyword arguments from `VisionDataset`.
+    """
+
+    URL = 'https://thor.robots.ox.ac.uk/~vgg/data/scenetext/SynthText.zip'
+    #SHA256 = '28ab030485ec8df3ed612c568dd71fb2793b9afbfa3a9d9c6e792aef33265bf1'
```
felixdittrich92, line 35: @fg-mindee 
is it ok to hold this as comment ?

tests/pytorch/test_datasets_pt.py
felixdittrich92: @fg-mindee 
Where to download the subsample (URL) ?? :sweat_smile: 
Can you take a look at the monkeypatch i have never used / needed this before :smiley:

doctr/datasets/synthtext.py
fg-mindee: It would be preferable to uncomment this. But you mentioned the file was too big to be hash?

doctr/datasets/synthtext.py
```diff
@@ -0,0 +1,85 @@
...
+        split = int(len(mat_data['imnames'][0]) * 0.9)
+        paths = mat_data['imnames'][0][:split] if self.train else mat_data['imnames'][0][split:]
+        boxes = mat_data['wordBB'][0][:split] if self.train else mat_data['wordBB'][0][split:]
+        labels = mat_data['txt'][0][:split] if self.train else mat_data['txt'][0][split:]
+
+        self.data: List[Tuple[str, Dict[str, Any]]] = []
+        np_dtype = np.float32
+
+        for img_path, word_boxes, txt in tqdm(iterable=zip(paths, boxes, labels),
+                                              desc='Load SynthText', total=len(paths)):
```
fg-mindee, line 61: "Loading SynthText..."

doctr/datasets/synthtext.py
```diff
@@ -0,0 +1,85 @@
...
+
+            labels = '\n'.join(txt).split()
+            word_boxes = word_boxes.transpose(2, 1, 0) if word_boxes.ndim == 3 else np.expand_dims(word_boxes, axis=0)
+
+            if rotated_bbox:
+                # x_center, y_center, w, h, alpha = 0
+                mins = word_boxes.min(axis=1)
+                maxs = word_boxes.max(axis=1)
+                box_targets = np.concatenate(
+                    ((mins + maxs) / 2, maxs - mins, np.expand_dims(np.zeros(word_boxes.shape[0]), axis=1)), axis=1)
```
fg-mindee, line 75: You can replace `np.expand_dims(np.zeros(word_boxes.shape[0]), axis=1)` with `np.zeros((word_boxes.shape[0], 1))`

tests/pytorch/test_datasets_pt.py
```diff
@@ -32,12 +33,25 @@ def test_visiondataset():
         ['IIIT5K', False, [32, 128], 3000, False],
         ['SVT', True, [512, 512], 100, True],
         ['SVT', False, [512, 512], 249, False],
+        ['SynthText', True, [512, 512], 27, True],  # org: 772875
+        ['SynthText', False, [512, 512], 3, False],  # org: 85875
```
fg-mindee, line 37: "org: " isn't very explicit
--> "Actual set has XXX samples"

tests/pytorch/test_datasets_pt.py
```diff
@@ -32,12 +33,25 @@ def test_visiondataset():
...
+        URL = 'https://github.com/mindee/doctr/releases/download/v0.4.1/synthtext_samples-89fd1445.zip'
+        FILE_HASH = '89fd1445457b9ad8391e17620c6ae1b45134be2bf5449f36e7e4275176cc16ac'
+        FILE_NAME = 'SynthText.zip'
+        download_from_url(URL, FILE_NAME, FILE_HASH, cache_subdir='datasets')
+    else:
+        download = True
+
     ds = datasets.__dict__[dataset_name](
-        train=train, download=True, sample_transforms=Resize(input_size), rotated_bbox=rotate,
+        train=train, download=download, sample_transforms=Resize(input_size), rotated_bbox=rotate,
```
fg-mindee, line 54: You can revert this

tests/pytorch/test_datasets_pt.py
```diff
@@ -32,12 +33,25 @@ def test_visiondataset():
...
+        download = False
+
+        # Download the subsample of SynthText dataset and save in cache
+        URL = 'https://github.com/mindee/doctr/releases/download/v0.4.1/synthtext_samples-89fd1445.zip'
+        FILE_HASH = '89fd1445457b9ad8391e17620c6ae1b45134be2bf5449f36e7e4275176cc16ac'
+        FILE_NAME = 'SynthText.zip'
+        download_from_url(URL, FILE_NAME, FILE_HASH, cache_subdir='datasets')
+    else:
+        download = True
+
```
fg-mindee, line 52:
```python
if dataset_name.lower() == "synthtext":
    # Monkeypatch the class to download a subsample
    datasets.__dict__[dataset_name].URL = 'https://github.com/mindee/doctr/releases/download/v0.4.1/synthtext_samples-89fd1445.zip'
    datasets.__dict__[dataset_name].SHA256 = '89fd1445457b9ad8391e17620c6ae1b45134be2bf5449f36e7e4275176cc16ac'
```

tests/tensorflow/test_datasets_tf.py
```diff
@@ -22,12 +23,25 @@
...
+        URL = 'https://github.com/mindee/doctr/releases/download/v0.4.1/synthtext_samples-89fd1445.zip'
+        FILE_HASH = '89fd1445457b9ad8391e17620c6ae1b45134be2bf5449f36e7e4275176cc16ac'
+        FILE_NAME = 'SynthText.zip'
+        download_from_url(URL, FILE_NAME, FILE_HASH, cache_subdir='datasets')
+    else:
+        download = True
+
     ds = datasets.__dict__[dataset_name](
-        train=train, download=True, sample_transforms=Resize(input_size), rotated_bbox=rotate,
+        train=train, download=download, sample_transforms=Resize(input_size), rotated_bbox=rotate,
```
fg-mindee, line 44: same here

tests/tensorflow/test_datasets_tf.py
```diff
@@ -22,12 +23,25 @@
         ['IIIT5K', False, [32, 128], 3000, False],
         ['SVT', True, [512, 512], 100, True],
         ['SVT', False, [512, 512], 249, False],
+        ['SynthText', True, [512, 512], 27, True],  # org: 772875
+        ['SynthText', False, [512, 512], 3, False],  # org: 85875
```
fg-mindee, line 27: same here

tests/tensorflow/test_datasets_tf.py
```diff
@@ -5,6 +5,7 @@
 from doctr import datasets
 from doctr.datasets import DataLoader
 from doctr.transforms import Resize
+from doctr.utils.data import download_from_url
```
fg-mindee, line 8: can be removed

tests/pytorch/test_datasets_pt.py
```diff
@@ -5,6 +5,7 @@
 
 from doctr import datasets
 from doctr.transforms import Resize
+from doctr.utils.data import download_from_url
```
fg-mindee, line 8: can be removed

tests/pytorch/test_datasets_pt.py
fg-mindee: (we really don't need the FILE_NAME)

fg-mindee (changes requested):
Thanks for taking a look at this Felix!
I added some modification suggestions :ok_hand:

doctr/datasets/synthtext.py
felixdittrich92: yes, i would just like to leave it there in case someone comes up with the idea of ​​wanting to check it out

doctr/datasets/synthtext.py
```diff
@@ -0,0 +1,85 @@
...
+        self.sample_transforms = sample_transforms
+        self.train = train
+
+        # Load mat data
+        tmp_root = os.path.join(self.root, 'SynthText')
+        mat_data = sio.loadmat(os.path.join(tmp_root, 'gt.mat'))
+        split = int(len(mat_data['imnames'][0]) * 0.9)
+        paths = mat_data['imnames'][0][:split] if self.train else mat_data['imnames'][0][split:]
+        boxes = mat_data['wordBB'][0][:split] if self.train else mat_data['wordBB'][0][split:]
+        labels = mat_data['txt'][0][:split] if self.train else mat_data['txt'][0][split:]
```
fg-mindee, line 55: Let's keep things simple:
```python
paths = mat_data['imnames'][0][:split if self.train else split:]
boxes = mat_data['wordBB'][0][:split if self.train else split:]
labels = mat_data['txt'][0][:split if self.train else split:]
```

doctr/datasets/synthtext.py
```diff
@@ -0,0 +1,85 @@
...
+        np_dtype = np.float32
+
+        for img_path, word_boxes, txt in tqdm(iterable=zip(paths, boxes, labels),
+                                              desc='Loading SynthText...', total=len(paths)):
+
+            # File existence check
+            if not os.path.exists(os.path.join(tmp_root, img_path[0])):
+                raise FileNotFoundError(f"unable to locate {os.path.join(tmp_root, img_path[0])}")
+
+            labels = '\n'.join(txt).split()
```
fg-mindee, line 67: why are we joining only to split?
Can't we split + strip directly?

doctr/datasets/synthtext.py
```diff
@@ -0,0 +1,85 @@
...
+from tqdm import tqdm
+
+from .datasets import VisionDataset
+
+__all__ = ['SynthText']
+
+
+class SynthText(VisionDataset):
+    """SynthText dataset from `"Synthetic Data for Text Localisation in Natural Images"
+    <https://arxiv.org/abs/1604.06646>`_.
```
fg-mindee, line 20: I'd argue it's worth mentioning the repo https://github.com/ankush-me/SynthText or the website https://www.robots.ox.ac.uk/~vgg/data/scenetext/ as well

fg-mindee (changes requested):
Thanks, I added a few comments!

doctr/datasets/synthtext.py
felixdittrich92: yeah thats right :)

doctr/datasets/synthtext.py
felixdittrich92: Mhh no `txt ` is a really ugly np.array :sweat_smile: 
example:
 array(['5FG      ', 'the      ', 'either   ', 'Sincerely', 'cars     ',
       'From:    ', 'the      ', 'and\nthat ', 'old      '], dtype='<U9')]

fg-mindee: I don't get how 1box label can have `\n` in it. Especially if we split it afterwards, 1 box = 2 text labels?
for your example, this would be more efficient:
```python
labels = [elt for word in txt.tolist() for elt in word.split()]
```

felixdittrich92: applied

fg-mindee (approved):
Looks good, thanks!

PR #626, opened 2021-11-15
fix: Adds missing console log and removed Tensorboard

fg-mindee:
- adds missing console log of validation metrics in the detection training
- removes support of Tensorboard

Any feedback is welcome!


charlesmindee (approved):
thanks!

references/detection/train_tensorflow.py
```diff
@@ -11,8 +11,6 @@
 import datetime
 import hashlib
 import time
-from collections import deque
-from pathlib import Path
 
 import numpy as np
 import tensorflow as tf
```
SiddhantBahuguna, line 16: May be we also have to import wandb in the script?

fg-mindee: but it is on line19 ;)

PR #627, opened 2021-11-15
refactor: Drops support of np.float16

fg-mindee:
- removed fp16 support from datasets and models' preprocessors
- removed support of np.float16 for targets in text detection models
- adapted unittests

Any feedback is welcome!


charlesmindee (approved):
Thanks!

fg-mindee:
cf #263

PR #629, opened 2021-11-16
feat: Added GPU support for classification and improve memory pinning

fg-mindee:
- removes unnecessary metric reset (it's already done at the beginning of `evaluate`)
- added missing support of GPU for character classification in PyTorch
- improved memory pinning for CPU training (if "True" it preallocates the memory on GPU, but in case of CPU training, we don't want that)

Any feedback is welcome!


charlesmindee (approved):
Thanks!

Issue #602, opened 2021-11-10
problem exporting saved_model


khalidMindee:
### Bug description

When loading a recognition model, I am unable to export the corresponding saved_model after one step of training . 

### Code snippet to reproduce the bug

```python
import os
import string

import numpy as np
import tensorflow as tf
from doctr.models import recognition

# train params
initial_lr = 1e-3
nb_steps = 10
batch_size = 16
input_shape = (20, 150, 3)
max_length = 34
output_dir = "output/test"

os.makedirs(output_dir, exist_ok=True)

# traindata sample
train_ims = tf.convert_to_tensor(np.ones((batch_size, *input_shape)), dtype=tf.float32)
train_labels = ["ABCD123456"] * batch_size

# Optimizer
scheduler = tf.keras.optimizers.schedules.ExponentialDecay(
    initial_lr, decay_steps=nb_steps, decay_rate=1 / (25e4), staircase=False  # final lr as a fraction of initial lr
)
optimizer = tf.keras.optimizers.Adam(learning_rate=scheduler, beta_1=0.95, beta_2=0.99, epsilon=1e-6, clipnorm=5)

# model loading
_model = recognition.__dict__["sar_resnet31"](
    pretrained=False,
    input_shape=input_shape,
    vocab=string.ascii_uppercase + string.digits,
    max_length=max_length
)
# model save works
tf.saved_model.save(_model, output_dir)

# one step training
with tf.GradientTape() as tape:
    train_loss = _model(train_ims, train_labels, training=True)["loss"]
grads = tape.gradient(train_loss, _model.trainable_weights)
optimizer.apply_gradients(zip(grads, _model.trainable_weights))

# model save doesn't work
tf.saved_model.save(_model, output_dir)

```

### Error traceback

```
Traceback (most recent call last):
  File "/home/khalid/Bureau/projects/invoice_v3/venv/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3444, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-12-e53061103513>", line 45, in <module>
    tf.saved_model.save(_model, output_dir)
  File "/home/khalid/Bureau/projects/invoice_v3/venv/lib/python3.8/site-packages/tensorflow/python/saved_model/save.py", line 1032, in save
    _, exported_graph, object_saver, asset_info = _build_meta_graph(
[...]

PR #630, opened 2021-11-16
feat: :sparkles: allow beam width > 1 in the CRNN postprocessor

khalidMindee:
This PR aims to allow beam search decoding in CRNN postprocessing , instead of only greedy decoding ( corresponding to a beam search with beam width equal to 1 )

### DIFF
--- a/doctr/models/recognition/crnn/tensorflow.py
+++ b/doctr/models/recognition/crnn/tensorflow.py
@@ -58,7 +58,9 @@ class CTCPostProcessor(RecognitionPostProcessor):
 
     def __call__(
         self,
-        logits: tf.Tensor
+        logits: tf.Tensor,
+        beam_width: int = 1,
+        top_paths: int = 1,
     ) -> List[Tuple[str, float]]:
         """
         Performs decoding of raw output with CTC and decoding of CTC predictions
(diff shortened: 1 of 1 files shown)


khalidMindee:
This PR solves #602 for exporting CRNN recognition models as saved models .

doctr/models/recognition/crnn/tensorflow.py
```diff
@@ -74,22 +76,35 @@ def __call__(
...
+
+        
+        if top_paths==1 :
+            probs = tf.math.exp(tf.squeeze(_log_prob, axis=1))# dim : batchsize
+            decoded_strings_pred = tf.squeeze(decoded_strings_pred,axis=1)
+        else :
+            probs = tf.math.exp(_log_prob)# dim : batchsize x beamwidth
 
-        return list(zip(word_values, probs.numpy().tolist()))
+        return {
```
charlesmindee, line 104: You are returning a dictionary containing tf.Tensor objects, I think we should keep the same output signature and return a list of string words, otherwise we are going to break the whole pipeline

charlesmindee (changes requested):
Thanks for this feature, but we may want to keep the same input/output format to avoid breaking the predictors

fg-mindee (changes requested):
As Charles mentioned, there are some parts of this PR that could be integrated and others that introduce breaking changes, that cannot be handled right now :/

I think we should split this into 2 PRs:
- one to fix SavedModel compatibility
- one to add topk

fharper:
@khalidMindee is this PR still good? If yes, can you review FG's & @charlesmindee comments, modify in consequences so we can potentially merge it please?

doctr/models/recognition/crnn/tensorflow.py
khalidMindee: yes this signature change has been removed to keep only changes for allowing beam width

khalidMindee:
This PR has been updated accordingly to keep only allowing beam width , postponning the fix for exporting the saved model as it introduces a breaking change due to signature change

charlesmindee (approved):
Thanks!

frgfm:
Missing two labels "module: models" and "ext: test" I think @charlesmindee  :)

PR #633, opened 2021-11-17
refactor: :zap: Speeded up CTC decoding in PyTorch by x10

fg-mindee:
- extends arg type support of `decode_sequence`
- refactored CTC decoding in PyTorch bringing a 10x speedup
- updated unittests

Any feedback is welcome!


charlesmindee (approved):
thanks for the refactor!

PR #632, opened 2021-11-17
feat: add 2 new rotation flags in the ocr_predictor

charlesmindee:
This PR adds the 2 flags introduced in #551 in the highest level predictor so the user can easily choose the parameters of the rotation.

Any feedback is welcome!

PR #641, opened 2021-11-19
refacto: :wrench: postprocessing with rotated boxes

charlesmindee:
Following #632 and #551, this PR removes deprecated flags and deprecated page-rotation steps that are no longer required since we are working with rotated boxes when `assume_straight_pages` is `False`.

Any feedback is welcome!!


fg-mindee (approved):
Thanks a lot Charles :pray: Looks good to me!

PR #644, opened 2021-11-22
docs: Fixed documentation build

fg-mindee:
- cleans up the remaining files after multi-version build
- fixed a silent fail of `git checkout` because of untracked files (which are now removed once the specific doc version is built)

Any feedback is welcome!


charlesmindee (approved):
Thanks

PR #645, opened 2021-11-24
fix: :bug: bug canvas dtype for threshold target

charlesmindee:
The threshold target map (of the DB) was set to be an uint8 tensor, but it should contains values between 0 and 1 (which were cast to 0).
This is now fixed.


fg-mindee (approved):
Thanks a lot!

Issue #646, opened 2021-11-24
Assertion error when using assume_straight_pages=False


Rob192:
### Bug description

When I am using `assume_straight_pages=False` in the `OCRPredictor` I am getting an Assertion error since #641 

### Code snippet to reproduce the bug

```python
from pathlib import Path

import matplotlib.pyplot as plt

from doctr.io import DocumentFile
from doctr.models.detection.zoo import detection_predictor
from doctr.models.predictor.tensorflow import OCRPredictor
from doctr.models.recognition.zoo import recognition_predictor
from doctr.utils.visualization import visualize_page

file = Path(r"C:\Users\Utilisateur\PythonProjects\ocr-xtract\data\aaaaa_straight.jpg") 

doc = DocumentFile.from_images(file)
det_bs=2
reco_bs=128

det_predictor = detection_predictor('db_resnet50', pretrained=True, batch_size=det_bs)
reco_predictor = recognition_predictor('crnn_vgg16_bn', pretrained=True, batch_size=reco_bs)

predictor = OCRPredictor(det_predictor=det_predictor, reco_predictor=reco_predictor, assume_straight_pages=False)

# Plot OCR output
out = predictor(doc)

fig = visualize_page(out.pages[0].export(), doc[0], interactive=False)
plt.show()
```

![aaaaa_straight](https://user-images.githubusercontent.com/33607785/143246053-89b5f292-b8d8-41a7-9730-9d95b0c5a115.jpg)

### Error traceback

```
Traceback (most recent call last):
  File "C:/Users/Utilisateur/PythonProjects/doctr/scripts/test_image.py", line 44, in <module>
    out = predictor(doc)
  File "C:\Users\Utilisateur\PythonProjects\doctr\doctr\models\predictor\tensorflow.py", line 62, in __call__
    crops, loc_preds = self._prepare_crops(
  File "C:\Users\Utilisateur\PythonProjects\doctr\doctr\models\predictor\base.py", line 49, in _prepare_crops
    crops = _OCRPredictor._generate_crops(pages, loc_preds, channels_last, assume_straight_pages)
  File "C:\Users\Utilisateur\PythonProjects\doctr\doctr\models\predictor\base.py", line 35, in _generate_crops
    crops = [
  File "C:\Users\Utilisateur\PythonProjects\doctr\doctr\models\predictor\base.py", line 36, in <listcomp>
    extraction_fn(page, _boxes[:, :-1], channels_last=channels_last)  # type: ignore[operator]
[...]

PR #647, opened 2021-11-24
fix: :bug: assume_straight_pages in predictor

charlesmindee:
This PR fixes the issue #646.
The detection predictor should receive the assume_straight_boxes flag when instantiated. 

Any feedback is welcome!


fg-mindee (approved):
Thanks for the fix!

PR #648, opened 2021-11-24
chore: Added rotations as augmentations

SiddhantBahuguna:
This PR introduces the following changes to the existing training script of object_detection:

1) Adds augmentations (rotations) 

Thank you so much for your feedback!

### DIFF
--- a/references/obj_detection/train_pytorch.py
+++ b/references/obj_detection/train_pytorch.py
@@ -8,6 +8,7 @@ import os
 os.environ['USE_TORCH'] = '1'
 
 import datetime
+import multiprocessing as mp
 import logging
 import random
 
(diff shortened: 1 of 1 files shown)


references/obj_detection/train_pytorch.py
```diff
@@ -185,7 +212,8 @@ def main(args):
             })
         if f1_score > max_score:
             print(f"Validation metric increased {max_score:.6} --> {f1_score:.6}: saving state...")
-            torch.save(model.state_dict(), f"./{exp_name}.pt")
+            # torch.save(model.state_dict(), f"./{exp_name}.pt")
+            torch.save(model.state_dict(), f"./{epoch}.pt")
```
fg-mindee, line 216: I'm not sure that will help. I'd argue we should keep the same naming, if you want to save every epoch, you can add a `torch.save` outside of this "if"

references/obj_detection/train_pytorch.py
```diff
@@ -198,13 +226,13 @@ def parse_args():
...
-    parser.add_argument('-b', '--batch_size', type=int, default=2, help='batch size for training')
+    parser.add_argument('--epochs', type=int, default=20, help='number of epochs to train the model on')
+    parser.add_argument('-b', '--batch_size', type=int, default=8, help='batch size for training')
     parser.add_argument('--device', default=None, type=int, help='device')
-    # parser.add_argument('--input_size', type=int, default=1024, help='model input size, H = W')
+    parser.add_argument('--input_size', type=int, default=1024, help='model input size, H = W')
     parser.add_argument('--lr', type=float, default=0.001, help='learning rate for the optimizer (SGD)')
     parser.add_argument('--wd', '--weight-decay', default=0, type=float, help='weight decay', dest='weight_decay')
-    parser.add_argument('-j', '--workers', type=int, default=14, help='number of workers used for dataloading')
+    parser.add_argument('-j', '--workers', type=int, default=2, help='number of workers used for dataloading')
```
fg-mindee, line 235: you need to merge main into your branch, since #649, the worker number resolution is automatic ;)

references/obj_detection/train_pytorch.py
```diff
@@ -168,7 +195,7 @@ def main(args):
     max_score = 0.
 
     for epoch in mb:
-        fit_one_epoch(model, train_loader, optimizer, scheduler, mb)
+        fit_one_epoch(model, train_loader, optimizer, scheduler, mb, is_transform=True)
```
fg-mindee, line 198: that doesn't sound like an explicit argument name :sweat_smile:

references/obj_detection/train_pytorch.py
```diff
@@ -21,32 +22,57 @@
...
         targets[idx]['boxes'][:, 1::2] = (targets[idx]['boxes'][:, 1::2] * height).round()
-
-    targets = [{
-        "boxes": torch.from_numpy(t['boxes']).to(dtype=torch.float32),
-        "labels": torch.tensor(t['labels']).to(dtype=torch.long)}
-        for t in targets
-    ]
-
-    return targets
+    if is_transform:
```
fg-mindee, line 33: could you add some comments please? I don't get what `is_transform` is supposed to do :thinking:

references/obj_detection/train_pytorch.py
```diff
@@ -21,32 +22,57 @@
...
     # Iterate over the batches of the dataset
     for _ in progress_bar(range(len(train_loader)), parent=mb):
         images, targets = next(train_iter)
         optimizer.zero_grad()
-        targets = convert_to_abs_coords(targets, images.shape)
+        if is_transform:
+            kwargs = {"images": images}
+            images, targets = convert_to_abs_coords(targets, images.shape, is_transform=True, **kwargs)
+        else:
+            targets = convert_to_abs_coords(targets, images.shape)
```
fg-mindee, line 75: Let's make augmentation separate :ok_hand: 
This is called "convert_to_abs_coords" so it's expected to do what it says it does and being deterministic

You can add a `data_augmentation` function that takes images + targets, and returns both, after the conversion :+1:

references/obj_detection/train_pytorch.py
```diff
@@ -21,32 +22,57 @@
...
-        "labels": torch.tensor(t['labels']).to(dtype=torch.long)}
-        for t in targets
-    ]
-
-    return targets
+    if is_transform:
+        if bool(random.getrandbits(1)):
+            angle = random.uniform(0.1, 5.0)
+        else:
+            angle = random.uniform(-0.1, -5.0)
```
fg-mindee, line 37: you won't want the same angle for the entire batch, better to do this as follows with a `max_angle`
```python
angles = 2 * max_angle * (np.random.rand(len(targets)) - 1)
```
and then apply each angle to each image set of targets

fg-mindee (changes requested):
Thanks! I added some suggestions to get better results :+1:

references/obj_detection/train_pytorch.py
SiddhantBahuguna: Thanks, I forgot to change this back to the way it was.

references/obj_detection/train_pytorch.py
SiddhantBahuguna: Thanks, resolved :)

references/obj_detection/train_pytorch.py
SiddhantBahuguna: Thanks, I have changed it, and included it in argument parser with a default bool value set to True.
 `fit_one_epoch(model, train_loader, optimizer, scheduler, mb, use_aug=args.use_augmentations)`

references/obj_detection/train_pytorch.py
SiddhantBahuguna: For sure :) 
`is_transform,` which is now changed to `use_aug` does the following:
It accepts a bool response, and if `use_aug = True`, augmentations will be employed and subsequently, images will be accepted as kwargs in  `convert_to_abs_coords(targets, img_shape, use_aug=False, **kwargs)`. If `use_aug= False`, we will not have to add images as kwargs. And the training will continue normally as it would without the augmentations.

references/obj_detection/train_pytorch.py
SiddhantBahuguna: Thanks, alright. I have added the function named as data_augmentations. However, since I think, before converting to tensor the values of the "targets" in covert_to_abs_coords, would it not be better if instead we call the functions data_augmentation inside it, and then at the end we can convert it all to tensor as required?

fg-mindee:
For reference, this was closed in favour of #654

fg-mindee:
And fyi @SiddhantBahuguna, don't forget to delete the branch once the PR is merged or closed ;)

SiddhantBahuguna:
Yup, for sure :) Thanks

PR #649, opened 2021-11-24
fix: Added automatic worker resolution to remaining training scripts

fg-mindee:
- adds automatic worker number resolution to training scripts where it wasn't enabled
- added missing arg print to object detection script

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #654, opened 2021-11-29
feat: Added random noise augmentation to object detection

SiddhantBahuguna:
With regards to an issue cf. #730 **Extends the list of supported data augmentations**,
> It adds Gaussian noise as an augmentation
```

from doctr.transforms import GaussianNoise
import torch
transfo = GaussianNoise(0., .25) 
output = transfo(torch.rand((3, 224, 224))) # input image as torch.Tensor
```
gn1

> It adds a unittest for the same class

Your feedback is highly welcomed :)


fg-mindee:
The modifications seem to be highly similar to #648, is there a reason to have them both?
I suggest to first check my feedback on the other PR, fix flake8 here and we'll go from there :ok_hand:

SiddhantBahuguna:
> The modifications seem to be highly similar to #648, is there a reason to have them both? I suggest to first check my feedback on the other PR, fix flake8 here and we'll go from there ok_hand



> Yes, this PR introduces random horizontal and vertical flip :)

fg-mindee:
> Yes, this PR introduces random horizontal and vertical flip :)

Alright, we if you open several PRs in parallel, in order to avoid conflict and ease the review, it's better if they are independent. Here you have the same modifications appearing in both PRs :sweat_smile:

SiddhantBahuguna:
> > Yes, this PR introduces random horizontal and vertical flip :)
> 
> Alright, we if you open several PRs in parallel, in order to avoid conflict and ease the review, it's better if they are independent. Here you have the same modifications appearing in both PRs sweat_smile

Yup, I am making changes in these ones as well :) Thanks

SiddhantBahuguna:
I am getting a ValueError when I change the batch_size from 8 to any other size. 
`ValueError: could not broadcast input array from shape (9,5) into shape (9,4).`

Traceback below:
https://github.com/mindee/doctr/blob/1c71c3c76fb82a8a6a3cc788711886ebe2f78e2d/references/obj_detection/train_pytorch.py#L63
Could you please take a look:) Thanks a lot!

fg-mindee:
> I am getting a ValueError when I change the batch_size from 8 to any other size. `ValueError: could not broadcast input array from shape (9,5) into shape (9,4).` _9 being the number of artefacts and 4 being the box coordinates for each batch_ 

That means the dimensions are most likely interpreted in the wrong order :upside_down_face:  

Traceback below:
> 
> https://github.com/mindee/doctr/blob/1c71c3c76fb82a8a6a3cc788711886ebe2f78e2d/references/obj_detection/train_pytorch.py#L63
> 

? that's the one causing the error but not the traceback :) (traceback = the complete error message + tracing you get when you run your code)


Also flake8 isn't happy :sweat_smile:

fg-mindee:
Mind merging main into this please? #691 has changed a few things :+1:

fg-mindee:
Would you mind resolving the conflicts so that we can merge the PR? :pray:

SiddhantBahuguna:
> Would you mind resolving the conflicts so that we can merge the PR? pray

Yup, I have fixed the problem that was causing us to choose arbitrary batch_sizes. Doing some minor code style fixes ta the moment

.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Default ignored files
+/shelf/
+/workspace.xml
```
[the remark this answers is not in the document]
fg-mindee: I added those in #727, please remove the file, then merge main into your branch :pray:

fg-mindee (changes requested):
Thanks for the PR! As discussed, the PR is too massive to correctly review it. I suggest we do several PRs to progressively integrate:
- uniform photometric transformations (noise, blur, etc.)
- non-uniform photometric transformations (shadows, etc.)
- geometric transformations (flips, rotations, crops, etc.)

Ideally, if we could have those as `doctr.transforms` (working on PIL or Torch/TF tensors) that would be perfect! I'll open an issue to track this :ok_hand:

.idea/.gitignore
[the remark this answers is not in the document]
SiddhantBahuguna: Alright. Thanks !

references/obj_detection/constants.py
```diff
@@ -0,0 +1,150 @@
+import random
```
fg-mindee, line 1: I'm not sure that file has its place in this PR

fg-mindee (changes requested):
The PR is massive, let's narrow it down to photometric augmentations as suggested previously. For instance, this PR could be only about adding a Noise augmentation. Everything else needs to be taken out, it's way too big for me to perform a careful review

On that note, if we use the dataloader, we do the transformation on pytorch tensor. Considering the operation you are doing in numpy, you can perfectly implement them in PyTorch or TensorFlow. In your case, you want to use it for a PyTorch script, so : 
- implement the `RandomGaussianNoise` transformation  over here https://github.com/mindee/doctr/blob/main/doctr/transforms/modules/pytorch.py
- adds some unittest
- adds the entry in the documentation
- adds an illustration of the augmentation in the PR description
- and finally add the transform in the Compose passed to the dataset :+1:

doctr/transforms/modules/pytorch.py
```diff
@@ -11,16 +11,16 @@
...
-        self,
-        size: Tuple[int, int],
-        interpolation=F.InterpolationMode.BILINEAR,
-        preserve_aspect_ratio: bool = False,
-        symmetric_pad: bool = False,
+            self,
+            size: Tuple[int, int],
+            interpolation=F.InterpolationMode.BILINEAR,
+            preserve_aspect_ratio: bool = False,
+            symmetric_pad: bool = False,
```
fg-mindee, line 23: please don't change things that aren't related to this PR :pray: 
(also this makes the indentation wrong)

doctr/transforms/modules/pytorch.py
```diff
@@ -53,3 +53,15 @@ def __repr__(self) -> str:
...
+class RandomGaussianNoise():
+    def __init__(self, mean=0.5, std=1.5):
+        self.std = std
+        self.mean = mean
+
+    def __call__(self, tensor):
+        return tensor + torch.randn(tensor.size()) * self.std + self.mean
+
+    def __repr__(self) -> str:
+        return self.__class__.__name__ + f"mean = {self.mean}, std = {self.std}"
```
fg-mindee, line 67: since we use modern python, don't use `+` between string, use f-string all the way :+1: 

I suggest inheriting from `doctr.utils.repr.NestedObject` so that you only have to modify the `def extra_repr(self) -> str:`

doctr/transforms/modules/pytorch.py
```diff
@@ -53,3 +53,15 @@ def __repr__(self) -> str:
...
         return f"{self.__class__.__name__}({_repr})"
+
+
+class RandomGaussianNoise():
+    def __init__(self, mean=0.5, std=1.5):
+        self.std = std
+        self.mean = mean
+
+    def __call__(self, tensor):
+        return tensor + torch.randn(tensor.size()) * self.std + self.mean
```
fg-mindee, line 64: This won't work: if the input tensor is an image, there are two cases:
- the image is in float, in range [0, 1]
- the image is in uint8 in range [0, 255]

in both cases, we want the resulting image to stay in that range.

Additionally, depends how you see the noise addition but:
- take a single pixel in RGB, we don't want a shift, we want variance around its initial value
so this would be (meaning that self.mean should be 0 by default)
```python
if x.dtype == torch.uint8:
    return (x + 255 * (self.mean + self.std * torch.rand(x.shape, device=x.device))).round().clamp(0, 255).to(dtype=torch.uint8)
else:
    return (x + self.mean + self.std * torch.rand_like(x)).clamp(0, 1)
```

fg-mindee: This is only because we perform operations on images and that we want to preserve their value range integrity

references/obj_detection/train_pytorch.py
```diff
@@ -104,33 +105,28 @@ def record_lr(
...
         "boxes": torch.from_numpy(t['boxes']).to(dtype=torch.float32),
         "labels": torch.tensor(t['labels']).to(dtype=torch.long)}
         for t in targets
     ]
-
     return targets
 
 
-def fit_one_epoch(model, train_loader, optimizer, scheduler, mb, amp=False):
+def fit_one_epoch(model, train_loader, optimizer, scheduler, mb, amp):
```
fg-mindee, line 119: Better to keep this default value

references/obj_detection/utils.py
```diff
@@ -3,6 +3,7 @@
 # This program is licensed under the Apache License version 2.
 # See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0.txt> for full license details.
 
+
```
fg-mindee, line 6: let's remove non necessary changes

fg-mindee (changes requested):
Thanks, added some comments!

fg-mindee:
Also, you'll need to:
- add unittest for this transformation
- add an entry in the documentation

doctr/transforms/modules/pytorch.py
SiddhantBahuguna: Yeah, actually I had added a class for adding Gaussian noise in the same script. Due to automatic collapsing of the import I couldnt figure out that the refactoring had changed the spacing :)  Will take care of it for the next time.

fg-mindee (approved):
Looking good, thanks!

PR #655, opened 2021-11-29
ci: Fixed silent isort failure

fg-mindee:
This PR tries to solve why isort doesn't throw an error while there is a file on `main` with the wrong import order. This finally manages to get some isort errors when needed (cf. https://github.com/mindee/doctr/runs/4356017960?check_suite_focus=true)

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #657, opened 2021-11-29
fix: rotation arg in training scripts

charlesmindee:
This PR fixes the rotation arg in training scripts:
- it is now passed in the model to instantiate the postprocessor correctly
- it is not passed in the dataset constructor, otherwise it will load our annotations as rotated boxes, but we have straight boxes and we rotate the images afterwards as a transformation in the get_itm fn.
Any feedback is welcome!

### DIFF
--- a/references/detection/train_pytorch.py
+++ b/references/detection/train_pytorch.py
@@ -112,7 +112,6 @@ def main(args):
         img_folder=os.path.join(args.val_path, 'images'),
         label_path=os.path.join(args.val_path, 'labels.json'),
         sample_transforms=T.Resize((args.input_size, args.input_size)),
-        rotated_bbox=args.rotation
     )
     val_loader = DataLoader(
         val_set,
@@ -131,7 +130,7 @@ def main(args):
     batch_transforms = Normalize(mean=(0.798, 0.785, 0.772), std=(0.264, 0.2749, 0.287))
 
     # Load doctr model
-    model = detection.__dict__[args.arch](pretrained=args.pretrained)
+    model = detection.__dict__[args.arch](pretrained=args.pretrained, assume_straight_pages=not args.rotation)
 
     # Resume weights
     if isinstance(args.resume, str):
@@ -175,7 +174,6 @@ def main(args):
             T.RandomApply(T.ColorInversion(), .1),
             ColorJitter(brightness=0.3, contrast=0.3, saturation=0.3, hue=0.02),
         ]),
-        rotated_bbox=args.rotation
     )
 
     train_loader = DataLoader(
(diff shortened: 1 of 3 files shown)


references/detection/train_pytorch.py
```diff
@@ -112,7 +112,6 @@ def main(args):
         img_folder=os.path.join(args.val_path, 'images'),
         label_path=os.path.join(args.val_path, 'labels.json'),
         sample_transforms=T.Resize((args.input_size, args.input_size)),
-        rotated_bbox=args.rotation
```
fg-mindee, line 115: Mmmh I have the feeling it's not doing the same thing.
Passing it here will output rotated bbox (source data being a polygon), but with your modification, only the model will be allowed to produce bboxes. Shouldn't we keep this?

references/detection/train_pytorch.py
```diff
@@ -175,7 +174,6 @@ def main(args):
             T.RandomApply(T.ColorInversion(), .1),
             ColorJitter(brightness=0.3, contrast=0.3, saturation=0.3, hue=0.02),
         ]),
-        rotated_bbox=args.rotation
```
fg-mindee, line 178: same here

references/detection/train_tensorflow.py
```diff
@@ -82,7 +82,6 @@ def main(args):
         img_folder=os.path.join(args.val_path, 'images'),
         label_path=os.path.join(args.val_path, 'labels.json'),
         sample_transforms=T.Resize((args.input_size, args.input_size)),
-        rotated_bbox=args.rotation
```
fg-mindee, line 85: same here

references/detection/train_tensorflow.py
```diff
@@ -128,7 +128,6 @@ def main(args):
             T.RandomContrast(.3),
             T.RandomBrightness(.3),
         ]),
-        rotated_bbox=args.rotation
```
fg-mindee, line 131: same here

fg-mindee (commented):
Thanks for the PR!

I agree with the model modification, but I'm wondering whether this is a good idea to remove the flag from the dataset

references/detection/train_pytorch.py
charlesmindee: If we pass this arg, the dataset will try to fit a rotated box from the targets, which will throw an error when we use sample_transform for the rotation transformation which takes only straight boxes as inputs.

fg-mindee: Oh I see, but then the solution should be to make special cases for rotated bbox in the transformations in the long-term! For now, I agree, let's keep things simple :ok_hand:

fg-mindee (approved):
Thanks!

fg-mindee (approved):
Thanks!

PR #659, opened 2021-11-29
feat: Added rbox_iou function with a memory-savy option

fg-mindee:
- adds `rbox_iou` function to compute IoU between sets of rotated bounding boxes
- replace iou computation in task metrics with `rbox_iou` while adding a `use_broadcasting` boolean flag
- adds missing constructor arg description in docstrings
- fixes typos in multiple docstrings

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #660, opened 2021-11-30
[refactor] SROIE dataset

felixdittrich92:
This PR refactors the SROIE implementation and enables also rotated boxes


doctr/datasets/sroie.py
```diff
@@ -49,35 +48,36 @@ def __init__(
...
-                        p1_x, p1_y, p2_x, p2_y, p3_x, p3_y, p4_x, p4_y = map(int, row[:8])
-                        left, right = min(p1_x, p2_x, p3_x, p4_x), max(p1_x, p2_x, p3_x, p4_x)
-                        top, bot = min(p1_y, p2_y, p3_y, p4_y), max(p1_y, p2_y, p3_y, p4_y)
-                        if len(label) > 0:
-                            _targets.append((label, [left, top, right, bot]))
-
-            text_targets, box_targets = zip(*_targets)
-
-            self.data.append((img_path, dict(boxes=np.asarray(box_targets, dtype=np_dtype), labels=text_targets)))
+                _lines = [[x.strip() for x in line.split(',')] for line in f.readlines()]
```
fg-mindee, line 63: This is the only line that needs to be indented

fg-mindee: Could you check that your proposition is actually as fast as with `csv` (it's standard library so I'm not sure)

fg-mindee: I think you're missing a `if len(line) > 0` to reproduce the same behaviour

doctr/datasets/sroie.py
```diff
@@ -49,35 +48,36 @@ def __init__(
...
+                labels = [",".join(line[8:]) for line in _lines if len(line) >= 9]
+                # reorder coordinates (8 -> (4,2)) and filter empty lines
+                coords = np.array([np.array(list(map(int, line[:8]))).reshape((4, 2))
+                                  for line in _lines if len(line) >= 9], dtype=np_dtype)
```
fg-mindee, line 67:
```python
coords = np.stack([np.array(list(map(int, line[:8])), dtype=np_dtype).reshape((4, 2))
                                  for line in _lines if len(line) >= 9], axis=0)
```

fg-mindee (commented):
Thanks for the refactor, I added some comments :+1:

doctr/datasets/sroie.py
felixdittrich92: @fg-mindee 
you are right csv is a bit faster :)
i have performed this check by `len(line)>=9 ` but now changed

fg-mindee (approved):
Looks good thanks!

PR #661, opened 2021-11-30
fix: Updates Makefile to match CI

fg-mindee:
This PR simply follows up on #655 to update the Makefile for contributors to use isort properly.

cc @felixdittrich92 

Any feedback is welcome!


charlesmindee (approved):
Thanks!

Issue #658, opened 2021-11-29
Make the demo torch-compatible


charlesmindee:
### 🚀 The feature

Run the demo with pytorch

### Motivation, pitch

This would be great to be able to run the demo with pytorch

felixdittrich92:
#1008 
temporary solution - should be changed to ONNX same for API if we have a valuable integration

PR #663, opened 2021-11-30
[WIP] make demo torch compatible

felixdittrich92:
@charlesmindee 
Issue:
#658


demo/requirements.txt
```diff
@@ -1,2 +1,2 @@
--e git+https://github.com/mindee/doctr.git#egg=python-doctr[tf]
+-e git+https://github.com/mindee/doctr.git#egg=python-doctr[all]
```
fg-mindee, line 1: So that part is tricky: all means that it will install both PyTorch & TF. and I disagree, a user should only need one of them

demo/app.py
```diff
@@ -60,8 +66,15 @@ def main():
...
-    reco_arch = st.sidebar.selectbox("Text recognition model", RECO_ARCHS)
+    framework = st.sidebar.selectbox("Framework", FRAMEWORK)
+    # use framework specific model selection
+    if framework == "PyTorch":
+        os.environ["USE_TORCH"] = "1"
+    else:
+        os.environ["USE_TF"] = "1"
+    (archs_det, archs_reco) = (PT_DET_ARCHS, PT_RECO_ARCHS) if framework == "PyTorch" else (TF_DET_ARCHS, TF_RECO_ARCHS)
+    det_arch = st.sidebar.selectbox("Text detection model", archs_det)
+    reco_arch = st.sidebar.selectbox("Text recognition model", archs_reco)
```
fg-mindee, line 77: I disagree, the backend should be selected via env (i.e. using `USE_TORCH='1' streamlit run demo/app.py` for instance). At runtime otherwise, since we don't have a hold on streamlit internal behaviour, we wouldn't want a user to use one, then try to use the other backend, which might fill up its RAM if the device isn't very beefy

demo/app.py
```diff
@@ -80,9 +93,9 @@ def main():
                 # Forward the image to the model
                 processed_batches = predictor.det_predictor.pre_processor([doc[page_idx]])
                 out = predictor.det_predictor.model(processed_batches[0], return_model_output=True)
-                seg_map = out["out_map"]
-                seg_map = tf.squeeze(seg_map[0, ...], axis=[2])
-                seg_map = cv2.resize(seg_map.numpy(), (doc[page_idx].shape[1], doc[page_idx].shape[0]),
+                seg_map = out["out_map"].numpy()
+                seg_map = np.squeeze(seg_map[0, ...], axis=2)
```
fg-mindee, line 97: the squeeze depends on the framework unfortunately

fg-mindee (changes requested):
Thanks for checking this, but the reason I haven't dealt with this PR is a bit complex:
- changes to the requirements (having everything) is not something we can afford, we don't want to mess up the environment of the user
- I really don't think switching backend at runtime is a good idea

So the problem is: I see perfectly how to make it framework independent in app.py (similar to your proposal), but for the requirements, it has to be conditional, which is complicated with a bare `requirements.txt`. Let's pause on this one (for the future, I suggest only starting to work on issues marked with the "help wanted" label, or at least ask on the issue if you can start to work on it :sweat_smile: )

felixdittrich92:
@fg-mindee 
I will close this ftm.
I have not really another idea without installing both if we want to use it in one demo. :confused: 
Maybe create 2 demos framework specific !? 
But lets discuss this before further iteration :smiley:

PR #666, opened 2021-12-01
feat: Added missing output classes param in DBNet

fg-mindee:
This PR adds `num_classes` as constructor param of DBNet in TF, which was the only text detection model missing this arg.

Any feedback is welcome!


charlesmindee (dismissed):
Thanks!

charlesmindee (approved):
Thanks!

PR #670, opened 2021-12-01
fix: Fixed LinkNet target & loss computation

fg-mindee:
- fixed `edge_mask` building
- fixed `seg_target` for PyTorch (was built as channels last)
- removes focal_loss option in LinkNet loss computation

Any feedback is welcome!


charlesmindee (approved):
Thanks, but why do you remove the focal loss ?

fg-mindee:
> Thanks, but why do you remove the focal loss ?

It was introducing an asymetry in model API, and we added it because we were having bad performances with LinkNet initially. So I figured, it's better to start again simple, make it work and then if we decide that all segmentation models need it, we'll reintroduce it :+1:

charlesmindee (approved):
Thanks!

PR #671, opened 2021-12-01
style: Renamed all compute_target to build_target

fg-mindee:
Since targets are built, and losses are computed, I figured we should update this :+1: 

Any feedback is welcome!


charlesmindee (approved):
Thanks

PR #673, opened 2021-12-02
fix: pymupdf version

charlesmindee:
This PR allows the PyMuPDF version to be 1.18.15 for in-house compatibility issues.

Any feedback is welcome!

### DIFF
--- a/requirements-pt.txt
+++ b/requirements-pt.txt
@@ -1,7 +1,7 @@
 numpy>=1.16.0
 scipy>=1.4.0
 opencv-python>=3.4.5.20
-PyMuPDF>=1.16.0,<=1.18.15
+PyMuPDF>=1.18.13,<=1.18.15
 pyclipper>=1.2.0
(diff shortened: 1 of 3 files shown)


fg-mindee (commented):
Before changing this, we need to ensure that version 1.18.11 is compatible. Because here enlarging the constraint will always install (in the CI) the latest, meaning 1.18.15. Can you confirm that you can run all unittests in `1.18.11`? :pray:

charlesmindee:
It is indeed not working with 1.18.11, it works since the 1.18.13

fg-mindee (dismissed):
Perfect thanks! Two things:
- let's do the same type of check when we want to update some dependency specifiers
- if you have a clear idea of the justification of the upper & lower bounds of the specifier, could you add an inline comment in the `setup.py`? (that will be easier for next time)

fg-mindee (approved):
Perfect thanks!

PR #677, opened 2021-12-07
feat: add line resolution for rotated boxes

charlesmindee:
This PR improves the line resolution heuristic: when we build a document with rotated boxes, lines are resolved the right way (with a rotation of the page).

Any feedback is welcome!

### DIFF
--- a/doctr/models/builder.py
+++ b/doctr/models/builder.py
@@ -10,7 +10,7 @@ import numpy as np
 from scipy.cluster.hierarchy import fclusterdata
 
 from doctr.io.elements import Block, Document, Line, Page, Word
-from doctr.utils.geometry import rbbox_to_polygon, resolve_enclosing_bbox, resolve_enclosing_rbbox
+from doctr.utils.geometry import rbbox_to_polygon, resolve_enclosing_bbox, resolve_enclosing_rbbox, rotate_boxes
 from doctr.utils.repr import NestedObject
 
 __all__ = ['DocumentBuilder']
@@ -32,7 +32,7 @@ class DocumentBuilder(NestedObject):
         resolve_lines: bool = True,
         resolve_blocks: bool = True,
         paragraph_break: float = 0.035,
-        export_as_straight_boxes: bool = False
+        export_as_straight_boxes: bool = False,
     ) -> None:
 
         self.resolve_lines = resolve_lines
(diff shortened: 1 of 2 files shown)


doctr/models/builder.py
```diff
@@ -10,7 +10,7 @@
 from scipy.cluster.hierarchy import fclusterdata
 
 from doctr.io.elements import Block, Document, Line, Page, Word
-from doctr.utils.geometry import rbbox_to_polygon, resolve_enclosing_bbox, resolve_enclosing_rbbox
+from doctr.utils.geometry import rbbox_to_polygon, resolve_enclosing_bbox, resolve_enclosing_rbbox, rotate_boxes
```
fg-mindee, line 13: Watch out, the behaviour of `rotate_boxes` has changed in #488. Make sure you merged "main" into your branch to avoid surprises later on :+1:

fg-mindee (commented):
Thanks! I added some comment, I think we need to ensure this is compatible with #488 and to fix the failing CI job :+1:

doctr/models/builder.py
charlesmindee: yes, this function was created in #488

doctr/models/builder.py
```diff
@@ -51,6 +51,12 @@ def _sort_boxes(boxes: np.ndarray) -> np.ndarray:
             indices of ordered boxes of shape (N,)
         """
         if boxes.shape[1] == 5:
+            boxes = rotate_boxes(
+                loc_preds=np.concatenate((boxes, np.zeros((boxes.shape[0], 1))), -1),
+                angle=-np.median(boxes[:, 4]),
+                orig_shape=(1024, 1024),
+                min_angle=5.,
+            )
```
fg-mindee, line 59: That's good for line aggregation, but don't we need to rotate the line back afterwards then?

fg-mindee (commented):
Small question in comment

doctr/models/builder.py
charlesmindee: I don't think so, because we only rotate the boxes to get the sorted idxs of "straight" boxes, but then we use those idxs to reconstruct the document with the original boxes (not rotated, the rotated ones are just used locally in resolve_lines)

fg-mindee (approved):
Thanks for the clarification!

PR #678, opened 2021-12-07
fix: rotate_boxes angle

charlesmindee:
This PR fixes a bug in the angle computation of rotate_boxes (in utils/geometry.py)

Any feedback is welcome!

### DIFF
--- a/doctr/utils/geometry.py
+++ b/doctr/utils/geometry.py
@@ -223,8 +223,10 @@ def rotate_boxes(
     rotated_centers = image_center + np.matmul(centers - image_center, rotation_mat)
     x_center, y_center = rotated_centers[:, 0] / orig_shape[1], rotated_centers[:, 1] / orig_shape[0]
     # Compute rotated boxes
-    rotated_boxes = np.stack((x_center, y_center, _boxes[:, 2], _boxes[:, 3], angle * np.ones_like(_boxes[:, 0]),
-                              _boxes[:, 5]), axis=1)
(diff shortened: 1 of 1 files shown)


doctr/utils/geometry.py
```diff
@@ -223,8 +223,10 @@ def rotate_boxes(
     rotated_centers = image_center + np.matmul(centers - image_center, rotation_mat)
     x_center, y_center = rotated_centers[:, 0] / orig_shape[1], rotated_centers[:, 1] / orig_shape[0]
     # Compute rotated boxes
-    rotated_boxes = np.stack((x_center, y_center, _boxes[:, 2], _boxes[:, 3], angle * np.ones_like(_boxes[:, 0]),
-                              _boxes[:, 5]), axis=1)
+    rotated_boxes = np.stack(
+        (x_center, y_center, _boxes[:, 2], _boxes[:, 3],
+         angle * np.ones_like(_boxes[:, 0]) + _boxes[:, 4], _boxes[:, 5]), axis=1
```
fg-mindee, line 228: Good catch, I'd only suggest the following
`angle * np.ones_like(_boxes[:, 0]) + _boxes[:, 4]`
---> `_boxes[:, 4] + angle`

fg-mindee (commented):
Thanks for the fix! Small change to simplify the computation of the angle column :+1:

fg-mindee (approved):
Thanks!

PR #681, opened 2021-12-07
fix: target_ar split crops

### LINKED ISSUE #592: Text recognition: Split wide crops parameter highly decrease performances
### 🚀 The feature

Fine tune the wide crops splitting method

### Motivation, pitch

Hey!

It's not a bug per se, because my code is running well. But I retrained a text recognition model, and the performances on my test set were very low compared to my validation set (both from the same dataset).

I digged a bit into the code, and noticed thes lines:
```
remapped = False
if self.split_wide_crops:
    new_crops, crop_map, remapped = split_crops(crops, self.critical_ar, self.target_ar, self.dil_factor)
    if remapped:
        crops = new_crops
```

I commented them, and the performances were back to what's expected.

I understand the motivation on this feature, but i noticed that it creates a lot of characters repetition in the output strings.

Example:
"AZZ812ZZF21" -> "AZZ81212ZZF21"

It's behaving as the cut was adding some padding, and some characters are recognized twice, leading to a repetition when reconstructing the final output.

I can send you a few sample data on Mindee's slack community if you need it.

### Alternatives

_No response_

### Additional context

_No response_

charlesmindee:
This PR set the default `target_ar` to 6 in split_crops, for better performances.
Any feedback is welcome!
closes #592


fg-mindee (approved):
Who doesn't love free perf improvements like those? :grin:

jonathanMindee:
Haha agree @fg-mindee
Thanks @charlesmindee

Note for the future: I struggled a bit to understand the problem was coming from this part. Because I first trained my reco model and got a X% exact match on the validation set. Then when I tried to compute the same metrics on the validation set using an inference script, I got a Y% exact match. Trying to understand why the X was different from the Y led me to this.

fg-mindee:
@jonathanMindee in the original issue #592, you commented the split/merge mechanism which led you to a perf increase.
This was the case in the specific situation of some hyperparameters that trigger the split/merge mechanism.
On my end, I had run some tests with looser constraints (the one of this PR) which got better performances, in a similar case as the one you encountered

> Trying to understand why the X was different from the Y led me to this.

About this part, if by "inference script" you mean https://github.com/mindee/doctr/blob/main/scripts/analyze.py or https://github.com/mindee/doctr/blob/main/scripts/evaluate.py, there are a few reasons that can produce minor diff:
- the evaluation loop of the training script is highly optimized for speed, everything is batched
- the `scripts/evaluate.py` script isn't leveraging batch inference for now (initially so that we don't have to loop several times to get all the metrics). It's on my TODO list to revamp this soon

But those should be minor diff, if that's not the case, we most likely have to fix something in one of those scripts.

I hope this clarifies things, don't hesitate if you have other questions :)

PR #682, opened 2021-12-07
feat: Added option to use AMP with TF scripts

### LINKED ISSUE #263: [references] Add FP16 support for training

fg-mindee:
This PR adds support of AMP for TensorFlow training scripts following this documentation: https://www.tensorflow.org/guide/mixed_precision.

Checking within the script, all modules and operations are indeed running the "mixed_float16" policy, but for some reasons, I haven't noticed any RAM saving on GPU :man_shrugging: 

Please note that some TF models still have issues with AMP (cf. https://github.com/tensorflow/tensorflow/issues/53345) but this PR enables it for training generally speaking.

Closes #263

Any feedback is welcome!

### DIFF
--- a/references/classification/train_tensorflow.py
+++ b/references/classification/train_tensorflow.py
@@ -16,6 +16,7 @@ import numpy as np
 import tensorflow as tf
 import wandb
 from fastprogress.fastprogress import master_bar, progress_bar
+from tensorflow.keras import mixed_precision
 
 gpu_devices = tf.config.experimental.list_physical_devices('GPU')
 if any(gpu_devices):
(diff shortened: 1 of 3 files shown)


charlesmindee (approved):
Thanks!

PR #688, opened 2021-12-09
refactor: Removed postprocessor from high-level init

fg-mindee:
- removes post processors from high-level init
- removes unused kwargs in linknet TF call
- removes legacy unittest of linknet with the focal loss

Any feedback is welcome!


charlesmindee (dismissed):
Thanks for the PR !

charlesmindee (approved):
Thanks for the edits !

Issue #591, opened 2021-11-08
Text recognition pretrained models: Use only pretrained backbone to allow fine tuning on different vocabularies


jonathanMindee:
### 🚀 The feature

The idea would be to be able to fine-tune a text recognition model on a different vocabulary than the pretrained model. Today it's not possible, as the whole pretrained model is frozen.

### Motivation, pitch

I have a use case where my character classification head would require only 34 classes. My dataset doesn't contain that many data, the pretrained feature is awesome to help my model converge. The problem is that today, the pretrained models were trained on 124 characters and it forces my model to be finetuned on the same vocabulary, although the problem is way easier as it's only on 34 characters.

Why not use a pretrained model only for the backbone, and let the classification head be free of constraints, so that we could benefit from the pretraining advantages without any constraints on the vocabulary?

fg-mindee:
The PR #687 will add a `pretrained_backbone` option to the last architecture that didn't have it in our model zoo :+1:

PR #689, opened 2021-12-09
fix: Fixed param override of backbone

fg-mindee:
This PR fixes the initialization when the feature extractor already has defined param values. Before this PR, those values were overridden with a param initialization, and this PR prevents this (for `pretrained_backbone=True` to work correctly).

cf. #591 

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #690, opened 2021-12-09
fix: Added missing AMP flags in training scripts

fg-mindee:
This PR simply adds the missing AMP flags to enable it on all training & evaluation loops for PyTorch.

Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR!

PR #692, opened 2021-12-10
fix: Fixed deprecation warnings from numpy & PyMuPDF

fg-mindee:
This PR fixes the multiple deprecation warnings from PyMuPDF & numpy by using the recommended replacement.

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #693, opened 2021-12-10
fix: Added a 0-sized crop safeguard in split_crops

fg-mindee:
- filters out chunked crops that are 0-sized in `split_crops` (cf. the error in https://github.com/mindee/doctr/runs/4475225188?check_suite_focus=true)
- adds the function docstring

Any feedback is welcome!


charlesmindee (approved):
Thanks for the fix!

PR #694, opened 2021-12-10
feat: Added possibility to change the cache dir of datasets

fg-mindee:
This PR adds the flexibility to override the cache folder used by `VisionDataset`, which makes it possible to have a flexible monkeypatch strategy for unittests.

cc @felixdittrich92


charlesmindee (approved):
Thanks for the PR!

PR #695, opened 2021-12-10
fix: Added safeguard for extreme aspect ratio in Resize

fg-mindee:
This PR fixes the second part of the unittest failures: now that we're positive no crop is 0-sized, in rare cases, preserving the aspect ratio can lead to a 0-sized intermediate image size. 

This should not be encountered for trained models but this PR adds a clipping to 1 pixel in case this happens!

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #697, opened 2021-12-10
fix: Fixed W&B logger in object detection training script

fg-mindee:
This PR fixes the W&B logger in the training script for object detection, and renamed "iou" metric into "mean_iou".

Any feedback is welcome!


charlesmindee (approved):
Thanks for that, I spotted it yesterday evening but I didn't take the time to open a PR

PR #698, opened 2021-12-10
fix: Fixed build_target for detection models with rotated targets

fg-mindee:
- fixes the `build_target` polygon conversion for LinkNet & DBNet when rotated targets are used
- adds a test case for rotated targets
- reduces the input size in unittest from 1024 to 512 to speed up the test & ease the RAM

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #700, opened 2021-12-10
fix: Fixed geometry utils for polygon <--> rbox conversions

fg-mindee:
- fixes the cv2 conventions of angle orientation for polygon <--> rbox conversions
- updates unittests

The following snippet:
```python
import numpy as np
from doctr.utils.geometry import rbbox_to_polygon

def plot_points(points):
    img = np.zeros((200, 150))
    for point in points.round().astype(int):
        img[point[1], point[0]] = 1
    plt.imshow(img); plt.show()


# Horizontal word rotated counter-clockwise by 45°
plot_points(rbbox_to_polygon((100, 100, 40, 10, 45.)))
```
was producing
before_rot

and now produces
now_rot

Additionally this snippet:
```python
import numpy as np
from doctr.utils.geometry import rbbox_to_polygon, fit_rbbox

angles = np.arange(-90, 91)
rboxes = [fit_rbbox(rbbox_to_polygon((100, 100, 40, 10, angle)))for angle in angles]
estimated_angles = [rbox[-1] for rbox in rboxes]
estimated_width = [rbox[2] for rbox in rboxes]

plt.plot(angles, estimated_angles, label="angles")
plt.plot(angles, estimated_width, label="width")
plt.xlabel("Anti-clockwise rotation")
plt.legend(loc="upper left")
plt.show()
```
used to yield:

before

and now yields (similar to before, we can't do more because of cv2):
now

Any feedback is welcome!

### DIFF
--- a/doctr/utils/geometry.py
+++ b/doctr/utils/geometry.py
@@ -22,12 +22,15 @@ def bbox_to_polygon(bbox: BoundingBox) -> Polygon4P:
 
 def rbbox_to_polygon(rbbox: RotatedBbox) -> Polygon4P:
     (x, y, w, h, alpha) = rbbox
-    return cv2.boxPoints(((float(x), float(y)), (float(w), float(h)), float(alpha)))
+    return cv2.boxPoints(((float(x), float(y)), (float(w), float(h)), -float(alpha)))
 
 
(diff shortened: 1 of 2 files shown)


charlesmindee (approved):
Thanks!

PR #702, opened 2021-12-13
[refactor] unify the boxes in relative coordinates

felixdittrich92:
This PR unified the boxes for each dataset which holds boxes

motivation:
we need a clean output format to work further with these

Any feedback and suggestions are very welcome :hugs:


doctr/datasets/cord.py
```diff
@@ -84,7 +86,7 @@ def __init__(
 
             self.data.append((
                 img_path,
-                dict(boxes=np.asarray(box_targets, dtype=int).clip(min=0), labels=text_targets)
+                dict(boxes=np.asarray(box_targets, dtype=np_dtype), labels=text_targets)
```
fg-mindee, line 89: careful here, we still need the clipping

doctr/datasets/datasets/base.py
```diff
@@ -49,6 +51,17 @@ def __getitem__(
             # typing issue cf. https://github.com/python/mypy/issues/5485
             img = self.sample_transforms(img)  # type: ignore[call-arg]
 
+        if isinstance(target, dict) and 'boxes' in target.keys() and \
+                not np.all((target['boxes'] <= 1) & (target['boxes'] >= 0)):
```
fg-mindee, line 55: It would be better to have a dynamic flag: 
```python
if self.convert_to_relative:
```
that would be set in the constructor

tests/pytorch/test_datasets_pt.py
```diff
@@ -61,6 +61,9 @@ def test_dataset(dataset_name, train, input_size, size, rotate):
     assert img.shape == (3, *input_size)
     assert img.dtype == torch.float32
     assert isinstance(target, dict)
+    assert np.all((target['boxes'] <= 1) & (target['boxes'] >= 0))
+    assert isinstance(target['labels'], list) or isinstance(
```
fg-mindee, line 65: Lazy can be cleaner sometimes :)
```python
assert isinstance(target['labels'], (list, tuple, np.ndarray))
```

tests/tensorflow/test_datasets_tf.py
```diff
@@ -51,6 +51,9 @@ def test_dataset(dataset_name, train, input_size, size, rotate):
     assert img.shape == (*input_size, 3)
     assert img.dtype == tf.float32
     assert isinstance(target, dict)
+    assert np.all((target['boxes'] <= 1) & (target['boxes'] >= 0))
+    assert isinstance(target['labels'], list) or isinstance(
+        target['labels'], tuple) or isinstance(target['labels'], np.ndarray)
```
fg-mindee, line 56: same here

fg-mindee (commented):
Thanks for the PR! I added a few comments but the general discussion is about changing the target of text detection :thinking: 

I'm starting to think that if we go this way, we should switch to object detection targets (having labels = 0 = "text" class). Anyway, there are conflicts with the main branch :sweat_smile:

felixdittrich92:
@fg-mindee 
Lets discuss this on monday via Slack in a short way :)
Fine with your requested changes :+1:  .. but we need a check in base.py for coordinates :hugs:

doctr/datasets/datasets/base.py
```diff
@@ -49,6 +53,18 @@ def __getitem__(
             # typing issue cf. https://github.com/python/mypy/issues/5485
             img = self.sample_transforms(img)  # type: ignore[call-arg]
 
+        if isinstance(target, dict) and 'boxes' in target.keys() and self.convert_to_relative:
```
fg-mindee, line 56: No need to check for something else than `self.convert_to_relative` considering we're unifying the target format

doctr/datasets/datasets/base.py
```diff
@@ -49,6 +53,18 @@ def __getitem__(
...
+            # Convert to relative coordinates if not already
+            if not np.all((target['boxes'][:, :4] <= 1) & (target['boxes'][:, :4] >= 0)):
```
fg-mindee, line 58: Same here, this can removed, we'll send to correct flag to specify whether the data need conversion in getitem directly.

doctr/datasets/datasets/base.py
```diff
@@ -20,12 +22,14 @@ class _AbstractDataset:
     def __init__(
         self,
         root: Union[str, Path],
+        convert_to_relative: bool = True,
```
fg-mindee, line 25: We need to add kwargs in `_VisionDataset` constructor, that are then passed to this constructor

doctr/datasets/cord.py
```diff
@@ -54,6 +54,8 @@ def __init__(
         self.data: List[Tuple[str, Dict[str, Any]]] = []
         self.train = train
         self.sample_transforms = sample_transforms
+        np_dtype = np.float32
```
fg-mindee, line 57: On the constructor call line 50, we need to specify the value for `convert_to_relative`

doctr/datasets/funsd.py
```diff
@@ -53,6 +53,8 @@ def __init__(
         # # List images
         tmp_root = os.path.join(self.root, subfolder, 'images')
         self.data: List[Tuple[str, Dict[str, Any]]] = []
+        np_dtype = np.float32
```
fg-mindee, line 56: Same in the constructor of this one

tests/pytorch/test_datasets_pt.py
```diff
@@ -63,14 +63,14 @@ def test_detection_dataset(mock_image_folder, mock_detection_label):
     assert img.dtype == torch.float32
     assert img.shape[-2:] == input_size
     # Bounding boxes
-    assert isinstance(target, np.ndarray) and target.dtype == np.float32
-    assert np.all(np.logical_and(target[:, :4] >= 0, target[:, :4] <= 1))
-    assert target.shape[1] == 4
+    assert isinstance(target['boxes'], np.ndarray) and target['boxes'].dtype == np.float32
```
fg-mindee, line 66: Before this line, we need a `assert isinstance(target, dict)`

tests/tensorflow/test_datasets_tf.py
```diff
@@ -51,14 +51,14 @@ def test_detection_dataset(mock_image_folder, mock_detection_label):
     assert img.shape[:2] == input_size
     assert img.dtype == tf.float32
     # Bounding boxes
-    assert isinstance(target, np.ndarray) and target.dtype == np.float32
-    assert np.all(np.logical_and(target[:, :4] >= 0, target[:, :4] <= 1))
-    assert target.shape[1] == 4
+    assert isinstance(target['boxes'], np.ndarray) and target['boxes'].dtype == np.float32
```
fg-mindee, line 54: same here

fg-mindee (changes requested):
Some extra suggestions in the comments :)

fg-mindee (commented):
Thanks Felix! cf my comment, let's pause this PR while I figure out how to handle this

tests/conftest.py
```diff
@@ -330,7 +330,7 @@ def mock_synthtext_dataset(tmpdir_factory, mock_image_stream):
     annotation_file = synthtext_root.join('gt.mat')
     labels = {
         "imnames": [[["8/ballet_106_0.jpg"], ["8/ballet_106_1.jpg"], ["8/ballet_106_2.jpg"]]],
-        "wordBB": [np.random.randint(1000, size=(2, 4, 3)) for _ in range(3)],
+        "wordBB": [np.random.uniform(low=100., high=600., size=(2, 4, 2)) for _ in range(3)],
```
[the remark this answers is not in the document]
felixdittrich92: @fg-mindee 
Yes i have checked the set again which uses float values around ~100.0 up to ~600.0 and i have faiced that the shape is wrong after testing also rotated boxes if have found this typo last dim 3 -> 2

doctr/datasets/datasets/base.py
```diff
@@ -49,6 +51,16 @@ def __getitem__(
...
+        if self.convert_to_relative:
+            h, w = self._get_img_shape(img)
+
+            # Boxes
+            boxes = target['boxes'].copy()
+            boxes[..., [0, 2]] /= w
+            boxes[..., [1, 3]] /= h
+            boxes = boxes.clip(0, 1)
+            target['boxes'] = boxes
+
```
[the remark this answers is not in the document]
felixdittrich92: Ok but i would really suggest to tackle this before the next release .. to harmonize the dataset and also coresponding test part .. overall the structure (datasets) feels a bit caotical so at starting point the time to fix it 👍 😃

fg-mindee:
Closed in favour of #775

Issue #478, opened 2021-09-16
[references] Add a LR Finder in the training scripts


fg-mindee:
For specific architectures, the learning rate that we generally use seems to yield various performances. We need a more consistent method to select an appropriate value for it.

Adding an option to select the learning rate would be quite valuable!
- [x] PyTorch support (#703)
- [ ] TensorFlow support (#747)

PR #703, opened 2021-12-13
feat: Added LR finder option in PyTorch training scripts

fg-mindee:
Following up on #478, this PR adds a LR finder feature in all PyTorch training scripts

Here is how it renders for artefact detection:
lr_find

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #707, opened 2021-12-13
feat: Added line reading for source PDFs

fg-mindee:
- adds `get_lines` methods on PDF objects to aggregate words into lines on source PDFs
- adds unittest
- adds entry in the documentation

Any feedback is welcome!


charlesmindee (dismissed):
Thanks!

Issue #680, opened 2021-12-07
[datasets] Mocking small versions of datasets for unittests


fg-mindee:
As docTR grows, the number of supported datasets will increase. We cannot afford to add several minutes to the CI tests for every dataset that we add. So I suggest the following:
- adding pytest fixture in `tests/conftest.py` that will create the data files in a temporary folder and return the path to it
- use this for dataset unittests instead of downloading the subsamples or full dataset

The sole inconvenience I can see is the time to implement, but the advantages are that we won't need internet to run those unittests anymore, the CI will be considerably faster and any developer will be able to read the structure of the dataset file in the unittest.

If we move forward with this, we'll have to do PRs for the following datasets:
- [ ] CORD #722
- [ ] DocArtefacts #719 
- [ ] FUNSD #722
- [ ] IC03 #722
- [x] IC13 #662
- [ ] IIIT5K #722
- [ ] SROIE #722
- [x] SVHN #634
- [ ] SVT #722
- [ ] SynthText #722
- [x] DetectionDataset
- [x] RecognitionDataset
- [x] OCRDataset

felixdittrich92:
@fg-mindee 
SVHN can marked as done :hook: :sweat_smile:

felixdittrich92:
@fg-mindee 
Can be closed if you are fine with the PR :smile:

PR #719, opened 2021-12-15
Mock Doc-Artefacts dataset

felixdittrich92:
Make CI a bit more happy :)
Mock Doc-Artefact dataset

Issue:
#680 

NOTE: PR fixed also a issue where `download_from_url` returns a wrong path if we do not pass a filename
so i have added a filename to each dataset which didn't already have one

Without passing a filename:
`Downloading /tmp/pytest-of-felix/pytest-30/datasets1/artefact_detection to /tmp/pytest-of-felix/pytest-30/datasets1/artefact_detection/artefact_detection/artefact_detection`


felixdittrich92:
@fg-mindee 
Closed it is integrated in #722 PR

PR #720, opened 2021-12-15
fix: box computing when assume straight pages is false

charlesmindee:
This PR fixes the following:

- angle computation in visualization.py
- angle computation in fit_rbbox
- Polygon expansion for rbbox: before this PR, we computed the area & perimeter of the polygon for the Vatti clipping algorithm from the raw polygon, now we compute it from the rotated rectangle which is more accurate and allow us to match the boxes when we have assume_straight_pages = False and assume_straight_pages = True.

before:
ex5

now:
 
ex3

### DIFF
--- a/doctr/utils/geometry.py
+++ b/doctr/utils/geometry.py
@@ -27,7 +27,7 @@ def rbbox_to_polygon(rbbox: RotatedBbox) -> Polygon4P:
 
 def fit_rbbox(pts: np.ndarray) -> RotatedBbox:
     ((x, y), (w, h), alpha) = cv2.minAreaRect(pts)
-    return x, y, h, w, 90 - alpha
+    return x, y, h, w, 90 + alpha
 
 
 def polygon_to_bbox(polygon: Polygon4P) -> BoundingBox:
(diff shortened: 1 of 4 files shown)


doctr/utils/geometry.py
```diff
@@ -27,7 +27,7 @@ def rbbox_to_polygon(rbbox: RotatedBbox) -> Polygon4P:
 
 def fit_rbbox(pts: np.ndarray) -> RotatedBbox:
     ((x, y), (w, h), alpha) = cv2.minAreaRect(pts)
-    return x, y, h, w, 90 - alpha
+    return x, y, h, w, 90 + alpha
```
fg-mindee, line 30: Mmmh, is this a good idea?

Because the following snippet:
```python
import numpy as np
from doctr.utils.geometry import rbbox_to_polygon, fit_rbbox

angles = np.arange(-90, 91)
rboxes = [fit_rbbox(rbbox_to_polygon((100, 100, 40, 10, angle)))for angle in angles]
estimated_angles = [rbox[-1] for rbox in rboxes]
estimated_width = [rbox[2] for rbox in rboxes]

plt.plot(angles, estimated_angles, label="angles")
plt.plot(angles, estimated_width, label="width")
plt.xlabel("Anti-clockwise rotation")
plt.legend(loc="upper left")
plt.show()

```

now yields
troubles

fg-mindee (commented):
Thanks a lot! Only a small question about the box fitting function in comment

fg-mindee (approved):
Let's keep in mind that the angle is off by 90°, but the rest is fine!

PR #721, opened 2021-12-16
feat: add classification module for crop orientation 

charlesmindee:
This PR adds a classification module with a `orientation_classifier` which will be used to classify the orientation of crops (0, 90, 180, 270 degrees ccw).
We also provide a `rectify_crops` function in utils which will be used to rotate the crops accordingly.

Any feedback is welcome!


fg-mindee (commented):
Oh my bad, there is a misunderstanding: either we rename backbone into classification (which would make a lot of sense), or we add your modifications to "backbones", but no need to have both I think!

charlesmindee:
We can merge both but we still need a predictor + model zoo for classification architectures, and we also need to have a cfg proper to this classification task (because the actual cfg of the mobilenet for instance is for a character classification, it is not the same input size, checkpoints, ...)

doctr/models/classification/mobilenet/pytorch.py
```diff
@@ -22,6 +23,7 @@
         'std': (0.299, 0.296, 0.301),
         'input_shape': (3, 32, 32),
         'vocab': VOCABS['legacy_french'],
+        'num_classes': len(VOCABS['legacy_french']),
```
fg-mindee, line 26: What do you think about having only one entry "classes" for all classification models which would be 
```python
'classes': list(VOCABS['legacy_french']),
```
in this case

doctr/models/classification/mobilenet/pytorch.py
```diff
@@ -45,8 +49,16 @@
...
         'rect_stride': ['features.2.block.1.0', 'features.4.block.1.0', 'features.9.block.1.0'],
         'vocab': VOCABS['french'],
+        'num_classes': len(VOCABS['french']),
         'url': None,
     },
+    'mobilenet_v3_small_orientation': {
+        'mean': (0.694, 0.695, 0.693),
+        'std': (0.299, 0.296, 0.301),
+        'input_shape': (3, 128, 128),
+        'num_classes': 4,
```
fg-mindee, line 59: with my previous suggestion, here the classes would be 0, 90, 180, 270

doctr/models/classification/predictor/pytorch.py
```diff
@@ -0,0 +1,55 @@
...
+        processed_batches = self.pre_processor(crops)
+        predicted_batches = [
+            self.model(batch)
+            for batch in processed_batches
+        ]
+
+        # Postprocess predictions
+        predicted_batches = [out_batch.argmax(dim=1).cpu().detach().numpy() for out_batch in predicted_batches]
+
+        return [int(pred) for batch in predicted_batches for pred in batch]
```
fg-mindee, line 55: Mmmh, should we return the class index (1) or the class value (90)?

doctr/models/classification/predictor/pytorch.py
```diff
@@ -0,0 +1,55 @@
...
+
+from typing import List, Union
+
+import numpy as np
+import torch
+from torch import nn
+
+from doctr.models.preprocessor import PreProcessor
+
+__all__ = ['OrientationClassifier']
```
fg-mindee, line 14: Also, I'm wondering whether we should keep "Predictor" in all our high level objects to make it clear :thinking: 
What do you think?

tests/pytorch/test_models_backbones_pt.py
```diff
@@ -1,7 +1,7 @@
 import pytest
```
fg-mindee, line 1: this file needs to be renamed and merged with the new one you created

tests/pytorch/test_models_classification_pt.py
```diff
@@ -0,0 +1,50 @@
...
+    assert out.shape == (8, 4)
+
+
+@pytest.mark.parametrize(
+    "arch_name",
+    [
+        "mobilenet_v3_small_orientation",
+    ],
+)
+def test_recognition_zoo(arch_name):
```
fg-mindee, line 34: test_classification_zoo

tests/tensorflow/test_models_backbones_tf.py
```diff
@@ -1,7 +1,7 @@
 import pytest
```
fg-mindee, line 1: same here, to be renamed & merged

fg-mindee (changes requested):
Thanks for the PR! I suggested a few changes!

doctr/models/classification/mobilenet/pytorch.py
charlesmindee: If so,you want the orientation classifier classes to be: `['0', '90', '180', '270']` ?

doctr/models/classification/mobilenet/pytorch.py
charlesmindee: yes ok

doctr/models/classification/predictor/pytorch.py
charlesmindee: I think it is better to return the class index, because I made the classes so that np.rot90(tensor, class) and np.roll(points, class) rotates the points the right way to reorder points of the crop (cf the next PR):

```
def rectify_crops(
    crops: List[np.ndarray],
    orientations: List[int],
) -> List[np.ndarray]:
    """Rotate each crop of the list according to the predicted orientation:
    0: already straight, no rotation
    1: 90 ccw, rotate 3 times ccw
    2: 180, rotate 2 times ccw
    3: 270 ccw, rotate 1 time ccw
    """
    # Inverse predictions (if angle of +90 is detected, rotate by -90)
    orientations = [4 - pred for pred in orientations if pred != 0]
    return [np.rot90(crop, orientation) for orientation, crop in zip(orientations, crops)]


def rectify_loc_preds(
    page_loc_preds: np.ndarray,
    orientations: List[int],
) -> np.ndarray:
    """Orient the quadrangle (Polygon4P) according to the predicted orientation,
    so that the points are in this order: top L, top R, bot R, bot L if the crop is readable
    """
    return np.stack(
        [np.roll(
            page_loc_pred,
            orientation,
            axis=0) for orientation, page_loc_pred in zip(orientations, page_loc_preds)],
        axis=0
    )
```

charlesmindee: (I mean we don't have to do a number of rotation --> angle conversion to redo afterwards an angle --> number of rotation conversion)

doctr/models/classification/predictor/pytorch.py
charlesmindee: what do you think of `orientation_predictor` ?

charlesmindee: or maybe `crop_orientation_predictor` to be clearer

fg-mindee (changes requested):
Thanks! I added a few comments, nothing major. Would you mind adding a corresponding section in the documentation please?

fg-mindee (changes requested):
Looks good, only missing the documentation :ok_hand:

fg-mindee (approved):
Looks good, thanks!

PR #724, opened 2021-12-16
refactor: Refactored detection post-processing

fg-mindee:
- speeds up and clean the binarization of the probability map and the opening transform
- makes it flexible enough to handle multi-channel outputs
- reflects changes in DB & LinkNet to stick with single-channel output

After comparing latency with the option of switching to PyTorch & TF implementations of the opening transform, I reverted back the changes for one reason: on CPU & GPU, it was all faster except for TensorFlow on CPU that takes ~4-6 times longer to compute the opening transform (max_pool2d seems to be really slow on TF for CPU :man_shrugging: )

Any feedback is welcome!

### DIFF
--- a/doctr/datasets/cord.py
+++ b/doctr/datasets/cord.py
@@ -84,7 +84,7 @@ class CORD(VisionDataset):
 
             self.data.append((
                 img_path,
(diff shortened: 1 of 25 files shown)


charlesmindee (dismissed):
Thanks!

PR #725, opened 2021-12-16
test: Fixed TF loss unittest

fg-mindee:
This PR fixes the tolerance threshold for the loss difference between straight & rotated bboxes

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #726, opened 2021-12-16
fix: Fixed edge cases of DB loss in PyTorch

fg-mindee:
This PR fixes the loss computation when the seg_mask is entirely on False.

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #728, opened 2021-12-18
feat: Added inference+post processing script for artefact detection

SiddhantBahuguna:
This PR introduces the following :

-> It adds a bare evaluation script without post nms and post processing. 
Any feedback is welcome :)


scripts/detect_artefacts.py
```diff
@@ -0,0 +1,105 @@
...
+
+import cv2
+import matplotlib.pyplot as plt
+from matplotlib.pyplot import figure
+import torch
+
+from doctr.models import obj_detection
+
+
+def inference(artefact, root_dir, test_list, visualize):
```
fg-mindee, line 22: the function is not doing inference --> "plot_predictions'

fg-mindee, line 22: and "visualize" shouldn't be an arg, but always enabled

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,105 @@
...
+    elif torch.cuda.is_available():
+        args.device = 0
+    else:
+        logging.warning("No accessible GPU, target device set to CPU.")
+    if torch.cuda.is_available():
+        torch.cuda.set_device(args.device)
+        model = model.cuda()
+    model.eval()
+    inf_dic = inference(model, args.root_dir, os.listdir(args.root_dir), args.visualize)
+    print(inf_dic)
```
fg-mindee, line 89: This is only to plot predictions, no need for the dict

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,105 @@
...
+            raise ValueError("Invalid device index")
+    # Silent default switch to GPU if available
+    elif torch.cuda.is_available():
+        args.device = 0
+    else:
+        logging.warning("No accessible GPU, target device set to CPU.")
+    if torch.cuda.is_available():
+        torch.cuda.set_device(args.device)
+        model = model.cuda()
+    model.eval()
```
fg-mindee, line 87: the model can be set to eval mode on line 72

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,105 @@
...
+    inf_dic = inference(model, args.root_dir, os.listdir(args.root_dir), args.visualize)
```
fg-mindee, line 88: Let's do the inference in the main, and only send the output to "plot_predictions"

fg-mindee (changes requested):
Added a few comments!

scripts/detect_artefacts.py
SiddhantBahuguna: Thanks, alright .

scripts/detect_artefacts.py
SiddhantBahuguna: Alright.

scripts/detect_artefacts.py
SiddhantBahuguna: Done! Thanks

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,86 @@
...
+    figure(figsize=(9, 7), dpi=100)
+    plt.imshow(image)
+    plt.show()
+
+
+def main(args):
+    print(args)
+
+    model = obj_detection.__dict__[args.arch](pretrained=True, num_classes=5)
+    model.eval()
```
fg-mindee, line 41: both lines can be merged

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,86 @@
...
+        text_w, text_h = text_size
+        cv2.rectangle(image, (int(val_2[0]), int(val_2[1])), (int(val_2[0]) + text_w, int(val_2[1]) - text_h),
+                      cm[str(int(tg_labels[ind_2]))][0], -1)
+        cv2.putText(image, cl_map[str(int(tg_labels[ind_2]))], (int(val_2[0]), int(val_2[1])),
+                    cv2.FONT_HERSHEY_SIMPLEX, 2, (255, 255, 255), 3)
+    figure(figsize=(9, 7), dpi=100)
+    plt.imshow(image)
+    plt.show()
+
+
```
fg-mindee, line 36: let's avoid computing gradients by adding a decorator
```python
@torch.no_grad()
```

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,86 @@
...
+    elif torch.cuda.is_available():
+        args.device = 0
+    else:
+        logging.warning("No accessible GPU, target device set to CPU.")
+    if torch.cuda.is_available():
+        torch.cuda.set_device(args.device)
+        model = model.cuda()
+
+    cm = {'1': [(0, 0, 150)], '2': [(0, 0, 0)], '3': [(0, 150, 0)], '4': [(150, 0, 0)]}
+    cl_map = {'1': "QR_Code", "2": "Bar_Code", "3": "Logo", "4": "Photo"}
```
fg-mindee, line 58: The target & predicted labels are integers, why use strings of digits while we can use the actual digit?

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,86 @@
...
+    # Silent default switch to GPU if available
+    elif torch.cuda.is_available():
+        args.device = 0
+    else:
+        logging.warning("No accessible GPU, target device set to CPU.")
+    if torch.cuda.is_available():
+        torch.cuda.set_device(args.device)
+        model = model.cuda()
+
+    cm = {'1': [(0, 0, 150)], '2': [(0, 0, 0)], '3': [(0, 150, 0)], '4': [(150, 0, 0)]}
```
fg-mindee, line 57: you take the first index of each of those values, no need to wrap them in a list, it's alright a tuple

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,86 @@
...
+        tg_boxes = pred[0]['boxes'].detach().cpu().numpy()
+        tg_boxes = [list(i) for i in tg_boxes]
+        plot_predictions(imm, tg_boxes, tg_labels, cl_map, cm)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Artefact Detection Training",
+                                     formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('arch', type=str, help='text-detection model to train')
+    parser.add_argument('root_dir', type=str, help='path to image folder')
```
fg-mindee, line 78: Let's use the path to a single image to make this simpler :pray:

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,86 @@
...
+    cm = {'1': [(0, 0, 150)], '2': [(0, 0, 0)], '3': [(0, 150, 0)], '4': [(150, 0, 0)]}
+    cl_map = {'1': "QR_Code", "2": "Bar_Code", "3": "Logo", "4": "Photo"}
+    for val in os.listdir(args.root_dir):
+        im_read = cv2.imread(os.path.join(args.root_dir, val))
+        imm = cv2.cvtColor(im_read, cv2.COLOR_BGR2RGB)
+        imm2 = imm / 255
+        if torch.cuda.is_available():
+            imm2 = [torch.tensor(imm2, dtype=torch.float32).permute(2, 0, 1).cuda()]
+        else:
+            imm2 = [torch.tensor(imm2, dtype=torch.float32).permute(2, 0, 1)]
```
fg-mindee, line 66: We need to reuse what's already been done
```python
from doctr.io.image import read_img_as_tensor

img = read_img_as_tensor(args.img_path).unsqueeze(0)

if torch.cuda.is_available():
    img = img.cuda()

```

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,86 @@
...
+        pred = model(imm2)
+        tg_labels = pred[0]['labels'].detach().cpu().numpy()
+        tg_boxes = pred[0]['boxes'].detach().cpu().numpy()
+        tg_boxes = [list(i) for i in tg_boxes]
```
fg-mindee, line 70: I think you don't need to cast those to list

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,86 @@
...
+        plot_predictions(imm, tg_boxes, tg_labels, cl_map, cm)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Artefact Detection Training",
+                                     formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('arch', type=str, help='text-detection model to train')
```
fg-mindee, line 77: "artefact detection model to use"

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,86 @@
...
+import matplotlib.pyplot as plt
+from matplotlib.pyplot import figure
+import torch
+
+from doctr.models import obj_detection
+
+
+def plot_predictions(image, tg_boxes, tg_labels, cl_map, cm):
+    for ind_2, val_2 in enumerate(tg_boxes):
+        cv2.rectangle(image, (int(val_2[0]), int(val_2[1])), (int(val_2[2]), int(val_2[3])),
```
fg-mindee, line 24: please add some comments specifying which section plot the enclosing rectangle and which one plots the text

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,86 @@
...
+import cv2
+import matplotlib.pyplot as plt
+from matplotlib.pyplot import figure
+import torch
+
+from doctr.models import obj_detection
+
+
+def plot_predictions(image, tg_boxes, tg_labels, cl_map, cm):
+    for ind_2, val_2 in enumerate(tg_boxes):
```
fg-mindee, line 23: instead of casting each value to int, use the numpy array :
```python
boxes = tg_boxes.round().astype(int)
```

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,86 @@
...
+        cv2.rectangle(image, (int(val_2[0]), int(val_2[1])), (int(val_2[2]), int(val_2[3])),
+                      cm[str(int(tg_labels[ind_2]))][0], 2)
+        text_size, _ = cv2.getTextSize(cl_map[str(int(tg_labels[ind_2]))], cv2.FONT_HERSHEY_SIMPLEX, 2, 2)
+        text_w, text_h = text_size
+        cv2.rectangle(image, (int(val_2[0]), int(val_2[1])), (int(val_2[0]) + text_w, int(val_2[1]) - text_h),
+                      cm[str(int(tg_labels[ind_2]))][0], -1)
+        cv2.putText(image, cl_map[str(int(tg_labels[ind_2]))], (int(val_2[0]), int(val_2[1])),
+                    cv2.FONT_HERSHEY_SIMPLEX, 2, (255, 255, 255), 3)
+    figure(figsize=(9, 7), dpi=100)
+    plt.imshow(image)
```
fg-mindee, line 33: missing some:
```python
plt.axis('off')
```

fg-mindee (changes requested):

scripts/detect_artefacts.py
SiddhantBahuguna: Yup, I remember doing it for post-nms purposes.. For this I have removed it. Thanks!

scripts/detect_artefacts.py
SiddhantBahuguna: Thanks!

scripts/detect_artefacts.py
SiddhantBahuguna: I agree. Thanks for the correction :)

scripts/detect_artefacts.py
fg-mindee: This hasn't been changed

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,88 @@
...
+import matplotlib.pyplot as plt
+from matplotlib.pyplot import figure
+import torch
+
+from doctr.models import obj_detection
+from doctr.io.image import read_img_as_tensor
+
+
+def plot_predictions(image, tg_boxes, tg_labels, cl_map, cm):
+    for ind_2, val_2 in enumerate(tg_boxes):
```
fg-mindee, line 24:
```python
for box, label in zip(tg_boxes, tg_labels):
```

fg-mindee: and then replace all "val_2...." and "tg_labels[..." afterwards

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,88 @@
...
+    elif torch.cuda.is_available():
+        args.device = 0
+    else:
+        logging.warning("No accessible GPU, target device set to CPU.")
+    if torch.cuda.is_available():
+        torch.cuda.set_device(args.device)
+        model = model.cuda()
+
+    cm = {1: (0, 0, 150), 2: (0, 0, 0), 3: (0, 150, 0), 4: (150, 0, 0)}
+    cl_map = {1: "QR_Code", 2: "Bar_Code", 3: "Logo", 4: "Photo"}
```
fg-mindee, line 63: Since those are constants, I suggest setting them after imports at the top of the script: (and we don't actually need dict)
```python
CLASSES = ["__background__", "QR Code", "Barcode", "Logo", "Photo"]
CM = [(255, 255, 255), (0, 0, 150), (0, 0, 0), (0, 150, 0), (150, 0, 0)]
```

then replace this accordingly (no need to pass it to "plot_predictions", they can be used inside directly

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,88 @@
...
+                      cm[tg_labels[ind_2]], 2)
+        text_size, _ = cv2.getTextSize(cl_map[int(tg_labels[ind_2])], cv2.FONT_HERSHEY_SIMPLEX, 2, 2)
+        text_w, text_h = text_size
+        # Filled rectangle above bounding box
+        cv2.rectangle(image, (val_2[0], val_2[1]), (val_2[0] + text_w, val_2[1] - text_h),
+                      cm[tg_labels[ind_2]], -1)
+        # Text bearing the name of the artefact detected
+        cv2.putText(image, cl_map[int(tg_labels[ind_2])], (int(val_2[0]), int(val_2[1])),
+                    cv2.FONT_HERSHEY_SIMPLEX, 2, (255, 255, 255), 3)
+    figure(figsize=(9, 7), dpi=100)
```
fg-mindee, line 36: not sure but do we really need to specify the figure here?

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,88 @@
...
+import os
+
+os.environ['USE_TORCH'] = '1'
+
+import argparse
+import logging
+
+import cv2
+import matplotlib.pyplot as plt
+from matplotlib.pyplot import figure
```
fg-mindee, line 16: we might not need this

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,88 @@
...
+# This program is licensed under the Apache License version 2.
+# See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0.txt> for full license details.
+
+
+import os
+
+os.environ['USE_TORCH'] = '1'
+
+import argparse
+import logging
```
fg-mindee, line 12: perhaps run flake8 + isort on this, because we aren't using that import here

scripts/detect_artefacts.py
fg-mindee: (talking about the first argument as well)

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,88 @@
...
+        logging.warning("No accessible GPU, target device set to CPU.")
+    if torch.cuda.is_available():
+        torch.cuda.set_device(args.device)
+        model = model.cuda()
+
+    cm = {1: (0, 0, 150), 2: (0, 0, 0), 3: (0, 150, 0), 4: (150, 0, 0)}
+    cl_map = {1: "QR_Code", 2: "Bar_Code", 3: "Logo", 4: "Photo"}
+    img = read_img_as_tensor(args.img_path).unsqueeze(0)
+    if torch.cuda.is_available():
+        img = img.cuda()
```
fg-mindee, line 66: if we move the img reading before the GPU check, we put the img tensor on GPU at the same time as the model :+1:

scripts/detect_artefacts.py
```diff
@@ -0,0 +1,88 @@
...
+    pred = model(img)
+    tg_labels = pred[0]['labels'].detach().cpu().numpy()
+    tg_labels = tg_labels.round().astype(int)
+    tg_boxes = pred[0]['boxes'].detach().cpu().numpy()
+    tg_boxes = tg_boxes.round().astype(int)
+    img = img.cpu().permute(0, 2, 3, 1).numpy()[0].copy()
+    plot_predictions(img, tg_boxes, tg_labels, cl_map, cm)
```
fg-mindee, line 73: I just realized but why "tg_" prefix? Those aren't target, they are predictions.
Perhaps let's keep something simpler "labels" and "boxes" :+1:

fg-mindee (changes requested):
Still some things to handle

scripts/detect_artefacts.py
SiddhantBahuguna: That was to increase the size of image plot because the default one may be a bit less. I am removing it for the moment though :)

scripts/detect_artefacts.py
SiddhantBahuguna: Agreed :)

scripts/detect_artefacts.py
SiddhantBahuguna: Nice!

scripts/detect_artefacts.py
SiddhantBahuguna: alright :)

fg-mindee (approved):
Looks good!

PR #734, opened 2021-12-20
fix: Fixed computation of Mean IoU

fg-mindee:
- updates the expected mean IoU value in unittests
- fixes the computation of `tot_iou` in metrics which is used for Mean IoU: it was taking the max IoU across all preds for a given GT, while it's supposed to take the max IoU across all GTs for a given pred.

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #735, opened 2021-12-21
test: Switch to a fully mocked PDF for unittests

fg-mindee:
- seeing that CI takes sometimes a long time to download the initial PDF, this PR mocks it completely to avoid data transfer
- updates impacted unittests

This speeds up the unittests of models and avoids CI failure when the URL is unreachable.

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #741, opened 2021-12-23
refactor: Unified return_preds flags across all tasks

fg-mindee:
This PR introduces the following modifications:
- ensures all prediction call flags are named the same way ("return_boxes" --> "return_preds")
- reflected changes on unittests, and demo

Any feedback is welcome!

PR #742, opened 2021-12-23
fix: Fixed detection training script

fg-mindee:
This PR fixes an issue introduced by #741 with the change of flags in the model call.

Any feedback is welcome!


charlesmindee (approved):
Thanks for the fix

PR #743, opened 2021-12-23
feat: add rotated ckpts for pytorch DBNet + fix line resolution for rotated pages

charlesmindee:
This PR adds rotated ckpts in the db_resnet50 cfg in torch, and fixes the document builder:
- fixes the angle computation to straighten boxes when resolving lines
- fixes the order of points in the rotated polygons (rectify_loc_preds function)
- fixes the line resolution

to be merged after #731 
Any feedback is welcome!


doctr/models/builder.py
```diff
@@ -57,27 +60,25 @@ def _sort_boxes(boxes: np.ndarray) -> np.ndarray:
...
                 min_angle=5.,
             )
-            # Points are in this order: top left, top right, bot right, bot left
-            return (boxes[:, 0, 0] + 2 * boxes[:, 2, 1] / np.median(
-                np.linalg.norm(boxes[:, 2, :] - boxes[:, 1, :])
-            )).argsort()
-        return (boxes[:, 0] + 2 * boxes[:, 3] / np.median(boxes[:, 3] - boxes[:, 1])).argsort()
+            xmin, xmax = np.min(boxes[:, :, 0], axis=1), np.max(boxes[:, :, 0], axis=1)
+            ymin, ymax = np.min(boxes[:, :, 1], axis=1), np.max(boxes[:, :, 1], axis=1)
+            boxes = np.stack([xmin, ymin, xmax, ymax], -1)
```
fg-mindee, line 65: Enjoy vectorization:

```python
boxes = np.concatenate((boxes.min(1), boxes.max(1)), -1)
```

doctr/models/builder.py
```diff
@@ -57,27 +60,25 @@ def _sort_boxes(boxes: np.ndarray) -> np.ndarray:
...
+        return (boxes[:, 0] + 2 * boxes[:, 3] / np.median(boxes[:, 3] - boxes[:, 1])).argsort(), boxes
 
     def _resolve_sub_lines(self, boxes: np.ndarray, words: List[int]) -> List[List[int]]:
         """Split a line in sub_lines
 
         Args:
-            boxes: bounding boxes of shape (N, 4) or (N, 4, 2) in case of rotated bbox
+            boxes: bounding boxes of shape (N, 4)
             words: list of indexes for the words of the line
```
fg-mindee, line 73: perhaps we should rename this to `word_idcs` to make it clear

doctr/models/builder.py
```diff
@@ -57,27 +60,25 @@ def _sort_boxes(boxes: np.ndarray) -> np.ndarray:
...
         Returns:
             A list of (sub-)lines computed from the original line (words)
         """
         lines = []
         # Sort words horizontally
-        words = [words[j] for j in np.argsort(
-            [boxes[i, 0, 0] if len(boxes.shape) == 3 else boxes[i, 0] for i in words]
-        ).tolist()]
+        words = [words[j] for j in np.argsort([boxes[i, 0] for i in words]).tolist()]
+
```
fg-mindee, line 81: Boxes is a numpy ndarray
```python
word_idcs = [word_idcs[idx] for idx in boxes[word_idcs, 0].argsort().tolist()]
```

doctr/models/builder.py
```diff
@@ -245,15 +235,15 @@ def _build_blocks(self, boxes: np.ndarray, word_preds: List[Tuple[str, float]])
         # Decide whether we try to form lines
         _boxes = boxes
         if self.resolve_lines:
-            lines = self._resolve_lines(_boxes[:, :4])
+            lines = self._resolve_lines(_boxes if len(_boxes.shape) == 3 else _boxes[:, :4])
```
fg-mindee, line 238:
```python
if _boxes.ndim == 3
```

doctr/models/builder.py
```diff
@@ -245,15 +235,15 @@ def _build_blocks(self, boxes: np.ndarray, word_preds: List[Tuple[str, float]])
...
             # Decide whether we try to form blocks
             if self.resolve_blocks and len(lines) > 1:
-                _blocks = self._resolve_blocks(_boxes[:, :4], lines)
+                _blocks = self._resolve_blocks(_boxes if len(_boxes.shape) == 3 else _boxes[:, :4], lines)
```
fg-mindee, line 241: same

doctr/models/builder.py
```diff
@@ -245,15 +235,15 @@ def _build_blocks(self, boxes: np.ndarray, word_preds: List[Tuple[str, float]])
...
             else:
                 _blocks = [lines]
         else:
             # Sort bounding boxes, one line for all boxes, one block for the line
-            lines = [self._sort_boxes(_boxes[:, :4])]
+            lines = [self._sort_boxes(_boxes if len(_boxes.shape) == 3 else _boxes[:, :4])[0]]
```
fg-mindee, line 246: same

doctr/models/builder.py
```diff
@@ -306,14 +295,11 @@ def __call__(
...
+                for p_boxes in boxes:
                     # Iterate over boxes of the pages
-                    for box in page_boxes:
-                        xmin, xmax = np.min(box[:, 0]), np.max(box[:, 0])
-                        ymin, ymax = np.min(box[:, 1]), np.max(box[:, 1])
-                        straight_boxes_page.append([xmin, ymin, xmax, ymax])
-                    straight_boxes.append(np.asarray(straight_boxes_page))
+                    xmin, xmax = np.min(p_boxes[:, :, 0], axis=1), np.max(p_boxes[:, :, 0], axis=1)
+                    ymin, ymax = np.min(p_boxes[:, :, 1], axis=1), np.max(p_boxes[:, :, 1], axis=1)
+                    straight_boxes.append(np.stack([xmin, ymin, xmax, ymax], -1))
```
fg-mindee, line 302:
```python
straight_boxes.append(np.concatenate((p_boxes.min(1), p_boxes.max(1)), 1))
```

doctr/models/detection/differentiable_binarization/pytorch.py
```diff
@@ -28,7 +28,8 @@
         'input_shape': (3, 1024, 1024),
         'mean': (0.798, 0.785, 0.772),
         'std': (0.264, 0.2749, 0.287),
-        'url': 'https://github.com/mindee/doctr/releases/download/v0.3.1/db_resnet50-ac60cadc.pt',
+        'url_straight_p': 'https://github.com/mindee/doctr/releases/download/v0.3.1/db_resnet50-ac60cadc.pt',
+        'url_rot_p': 'https://github.com/mindee/doctr/releases/download/v0.4.1/db_resnet50-1138863a.pt',
```
fg-mindee, line 32: If the checkpoint has superior perf, I suggest simply replacing the URL.

If it isn't, I would suggest to create a new entry in the config dict to keep things harmonized :+1: 

`db_resnet50_rotation`

doctr/utils/visualization.py
```diff
@@ -218,6 +220,7 @@ def visualize_page(
                         alpha=0.5,
                         color=(0, 0, 1),
                     )
+                pass
```
fg-mindee, line 223: why do we need this?

fg-mindee (changes requested):
Thanks a lot! Minor improvement suggestions in the review :+1:

doctr/models/_utils.py
```diff
@@ -214,9 +214,9 @@ def rectify_loc_preds(
...
     """
     return np.stack(
-        [
-            page_loc_pred if orientation == 0 else np.roll(page_loc_pred, orientation, axis=0)
-            for orientation, page_loc_pred in zip(orientations, page_loc_preds)
-        ],
+        [np.roll(
+            page_loc_pred,
+            orientation - 1,
+            axis=0) for orientation, page_loc_pred in zip(orientations, page_loc_preds)],
```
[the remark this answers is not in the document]
charlesmindee: yes it was shifted by 1 (crops were 90° rotated, I took the screenshot of the rotated samples of the previous PR with this branch)

doctr/models/detection/differentiable_binarization/pytorch.py
charlesmindee: I think it makes the high level commands harder to use if we have 2 separate models, because users will have to specify the model in the predictor instead of just passing "assume_straight_pages=False", what do you think ?

doctr/models/_utils.py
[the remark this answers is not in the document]
fg-mindee: But I'm not understanding why: if orientation == 0, then it doesn't have to be shifted, correct?
but here if orientation is 1, the np.roll is applied with `orientation - 1 = -1` :thinking:

doctr/models/detection/differentiable_binarization/pytorch.py
fg-mindee: Ideally, the checkpoint should improve everything, and we should enable rotation by default and have a single checkpoint. If that's not the case, there should be one entry in the dict by checkpoint we make available

Otherwise, we will end up with various default dict format across all models (I just unified everything across all tasks :sweat_smile: #763 )

charlesmindee: In that case I think we will start with 2 entries before benchmarking both, because I am not sure the rotated checkpoints is better than the previous one on straight documents which represent 95% of the stream

doctr/models/_utils.py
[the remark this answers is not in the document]
charlesmindee: If orientation = 0, it has to be shifted to match the convention loc_preds = [top_left, top_right, bottom_left, bottom_right]
and `np.roll` is OK with negative shift.
This changes nothing to the way the crops are read, it is just to return the `loc_preds` correctly to the user in the output along with the corresponding words, and to display the boxes correctly with the vizualisation features

doctr/models/builder.py
```diff
@@ -48,7 +48,10 @@ def _sort_boxes(boxes: np.ndarray) -> np.ndarray:
             boxes: bounding boxes of shape (N, 4) or (N, 4, 2) (in case of rotated bbox)
 
         Returns:
-            indices of ordered boxes of shape (N,)
+            tuple: indices of ordered boxes of shape (N,), boxes
+            If straight boxes are passed tpo the function, boxes are unchanged
+            else: boxes returned are straight boxes fitted to the straightened rotated boxes
+            so that we fit the lines afterwards to the straigthened page
```
fg-mindee, line 54: for the docstring to be properly rendered, the next lines have to be extra-indented

doctr/models/detection/differentiable_binarization/pytorch.py
```diff
@@ -363,3 +372,25 @@ def db_mobilenet_v3_large(pretrained: bool = False, **kwargs: Any) -> DBNet:
...
+        >>> out = model(input_tensor)
+
+    Args:
+        pretrained (bool): If True, returns a model pre-trained on our text detection dataset
+
+    Returns:
+        text detection architecture
+    """
+
+    return _dbnet('db_resnet50_rotation', pretrained, **kwargs)
```
fg-mindee, line 396:
```python
return _dbnet(
        'db_resnet50_rotation',
        pretrained,
        resnet50,
        ['layer1', 'layer2', 'layer3', 'layer4'],
        None,
        **kwargs,
    )
```

doctr/models/detection/differentiable_binarization/pytorch.py
```diff
@@ -39,6 +39,15 @@
         'std': (0.264, 0.2749, 0.287),
         'url': 'https://github.com/mindee/doctr/releases/download/v0.3.1/db_mobilenet_v3_large-fd62154b.pt',
     },
+    'db_resnet50_rotation': {
+        'backbone': resnet50,
+        'backbone_submodule': None,
+        'fpn_layers': ['layer1', 'layer2', 'layer3', 'layer4'],
```
fg-mindee, line 45: You can remove all those 3 dict entries (cf. the refactoring and the other dict entries)

doctr/transforms/modules/tensorflow.py
```diff
@@ -75,6 +75,7 @@ def extra_repr(self) -> str:
 
     def __call__(self, img: tf.Tensor) -> tf.Tensor:
         input_dtype = img.dtype
+        print(self.output_size)
```
fg-mindee, line 78: to be removed

doctr/models/_utils.py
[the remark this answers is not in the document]
fg-mindee: I think we should shift the points right from the start when we use `cv2.boxPoints` and adapt this to make it clearer and not mixing things

fg-mindee (changes requested):
Thanks, only minor changes  left! Still some CI jobs failing though

fg-mindee (approved):
Thank a lot for all this work!

PR #744, opened 2021-12-23
test: Increased flexibility of loss test

fg-mindee:
Some of our unittests involve random inputs, and the check of similarity between the values of the loss on straight boxes vs rotated ones has failed quite a lot. This PR increases the threshold to avoid that problem.

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #745, opened 2021-12-23
fix: Fixed the bin_thresh of LinkNet

fg-mindee:
This PR simply updates the binary threshold of LinkNet which was the same as in the DBNet. But with the difference in the LinkNet loss, 0.15 is way too low. This modification gives a boost on text detection evaluation.

Any feedback is welcome!


charlesmindee (approved):
Thanks!

Issue #699, opened 2021-12-10
[scripts] Adding scripts to easily evaluate the latency of models


fg-mindee:
As discussed, we need to provide means to accurately evaluate the latency of models in a given environment. This comes with several challenges and questions:
- 1 script per framework?
- 1 script per task? (text detection, object detection, text recognition)
- should we benchmark the predictor or the DL model only? (for OCR we don't have any end-to-end DL model for instance, so that )

I would suggest focusing on mid-level tasks (excluding OCR for now since we don't have an end-to-end archi) and implementing:
- [ ] Text Detection benchmark script (#746)
- [ ] Text Recognition benchmark script (#746)
- [ ] Object Detection benchmark script (#746)
 
What do you think @charlesmindee?

PR #746, opened 2021-12-23
feat: Added latency evaluation scripts for all tasks

fg-mindee:
As per #699, some latency benchmarking was required. This PR introduces scripts in PyTorch & TensorFlow to benchmark latency on CPU & GPU for text recognition and text detection.

The results were added in a csv file in the same folder to keep track of latency changes for each model.

Finally, a CI job was added to ensure those scripts can run properly.

Any feedback is welcome!

### DIFF
--- a/.github/workflows/references.yml
+++ b/.github/workflows/references.yml
@@ -7,7 +7,7 @@ on:
     branches: main
 
 jobs:
-  train-classification:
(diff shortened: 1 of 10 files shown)


charlesmindee (approved):
Thanks!

PR #747, opened 2021-12-23
feat: Added LR Finder for TensorFlow scripts

### LINKED ISSUE #478: [references] Add a LR Finder in the training scripts

fg-mindee:
- fixes output node sizing for mobile nets in TF & PyTorch (encountered problems with the script in char classification)
- adds LR finder option for all supported tasks in TensorFlow

Closes #478

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #750, opened 2021-12-24
refactor: Introduces img + target transforms in Datasets

fg-mindee:
Following up on #352, this PR introduces the following modifications:
- renames "sample_transforms" into "img_transforms" to be more explicit (hesitated with "input_transforms" but we only cover Vision, so I don't think it's necessary)
- adds "sample_transforms" which are performed after "img_transforms" on both the image and the target
- reflected changes in datasets, unittests and training scripts
- added `SampleCompose` (to compose sample_transforms) and `ImgTransformation` (to wrap an img_transform as a sample_transform) utils
- added dedicated unittests

With this PR, the way to pass thransforms becomes the following:
- if you only have transformations that don't modify the target, pass them as `img_transforms`
- if you have both, there are two options:
    - passing the image transformations that have to be performed before the first one to modify the target & pass the rest as `sample_transforms`:
    - passing all transformations as `sample_transforms` and when there is one that is only implemented for image, wrap it with `ImageTransform`

Any feedback is welcome!

### DIFF
--- a/doctr/datasets/classification/base.py
+++ b/doctr/datasets/classification/base.py
@@ -45,10 +45,10 @@ class _CharacterGenerator(AbstractDataset):
         vocab: str,
         num_samples: int,
         cache_samples: bool = False,
-        sample_transforms: Optional[Callable[[Any], Any]] = None,
+        img_transforms: Optional[Callable[[Any], Any]] = None,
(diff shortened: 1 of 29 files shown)


charlesmindee (approved):
Thanks for the PR

PR #753, opened 2021-12-24
fix: Fixed mask computation of DBNet

fg-mindee:
This PR simply fixes the segmentation mask building: I believe we have never entered that loop, because considering its shape, it's impossible that it would have worked.

Any feedback is welcome!


charlesmindee (approved):
Thanks for that fix!

PR #756, opened 2021-12-26
refactor: Removed contiguous param since it's included in torch>=1.7

fg-mindee:
- removes the Pytorch to force params in a contiguous chunk to speed up optimization (now it's included in the core library)
- updates the CI & documentation accordingly
- removes unneeded loader operations in training scripts

Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR

PR #757, opened 2021-12-26
feat: Added latency evaluation & benchmark for image classification

fg-mindee:
Following up on #746, this PR adds latency evaluation for image classification (to evaluate the influence of the backbone/feature extractor on latency).

Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR !

PR #760, opened 2021-12-26
feat: Added WordGenerator dataset

### LINKED ISSUE #262: [datasets] Add a synthetic recognition dataset
The library would greatly benefit from synthetic data. It could be very helpful for character classification, and text recognition.

Here are suggestions about the augmentations operating on discrete spectrum:
- choice of character in the vocab
- choice of font family
- choice of font style (bold, italic, etc)
_The last two would not be relevant for handwritten characters_

And those operating on a continuous spectrum: all common image classifications augmentations (rotation, color modifications, etc.)

fg-mindee:
- implements a `WordGenerator` in the same spirit as the `CharGenerator` using a `min_chars` and `max_chars` to specify the word length.
- updates unittests and documentation

To illustrate the PR, the following snippet:
```python
from doctr.datasets.generator.base import synthesize_text_img

plt.imshow(synthesize_text_img('a', font_size=32)); plt.show()
plt.imshow(synthesize_text_img('abcdefgh', font_size=32)); plt.show()
```
produces:
single_letter
full_word

And the same in `font_size=64`:
single_letter64
word64

Finally, the dataset itself:
```python
from doctr.datasets import WordGenerator, VOCABS
from math import floor

# Font will be picked randomly in the list
ds = WordGenerator(VOCABS['french'], 1, 10, num_samples=8, font_family=["FreeMono.ttf", "FreeSans.ttf", "FreeSerif.ttf"])
# Visualize
_, axes = plt.subplots(4, 2)
for idx in range(8):
    img, target = ds[idx]
    row_idx = int(idx / 2)
    col_idx = idx % 2
    axes[row_idx, col_idx].imshow(img.numpy())
    axes[row_idx, col_idx].set_title(target)
for ax in axes.ravel():
   ax.axis('off')
plt.show()
```
renders as follows:
wordgen

(Easily turned into something more appealing with transformations like ColorInversion, etc.)

_Please note that some characters such as bitcoin symbol are not properly rendered silently, this will have to be investigated_

Closes #262


charlesmindee (approved):
Thanks for the PR, indeed we need to find a way to filter out characters that cannot be rendered, an option could be to cast them with unidecode

PR #761, opened 2021-12-26
feat: Added dedicated evaluation scripts for text detection

fg-mindee:
- adds scripts to evaluate text detection models on supported datasets
- adds the perf report as a .csv

I'll have to open another PR for text recognition because the crop needs to be made live efficiently. Those PRs aim at speeding up evaluation runs for models once they are trained

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #762, opened 2021-12-26
feat: Updated bin_thresh for LinkNet

fg-mindee:
We had set the same bin_thresh value as for the DBNet, which has a very different loss. This PR fixes that by setting a more appropriate threshold, which yields better evaluation performances!

Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR

fg-mindee:
Actually, I had pushed that branch because it was chilling on my local environment. But I just realized those changes were already made in #745 and for some reasons Github is making it look like this PR changes something :thinking: 
Closing this

PR #763, opened 2021-12-27
feat: Refactored & retrained all classification models

fg-mindee:
- refactors and retrained all classification models (apart from resnet31 & magc_resnet31 in TF which are rather slow)
- cleans up the factory functions of text detection & text recognition (ensures to be able to use pretrained backbones)
- updates documentation
- updates unittests

This PR was used to train downstream tasks leveraging the default `pretrained_backbone=True` which increases stability and convergence.

Any feedback is welcome!


charlesmindee (approved):
Thanks!

PR #765, opened 2021-12-28
feat: add rotation option to both detection training scripts

charlesmindee:
This PR adds the ability to train with rotated samples, by simply passing --rotation as an argument of the CLI detection training scripts.
The default rotation transformation performed on your training set is +-90°, and no rotation on the validation set, but this can be easily modified.

Here is how it renders for a batch of training samples:
samples

Any feedback is welcome!


doctr/transforms/functional/tensorflow.py
```diff
@@ -52,6 +52,8 @@ def rotate(
     if expand:
         exp_shape = compute_expanded_shape(img.shape[:-1], angle)
         h_pad, w_pad = int(math.ceil(exp_shape[0] - img.shape[0])), int(math.ceil(exp_shape[1] - img.shape[1]))
+        if min(h_pad, w_pad) < 0:
+            h_pad, w_pad = int(math.ceil(exp_shape[1] - img.shape[0])), int(math.ceil(exp_shape[0] - img.shape[1]))
```
fg-mindee, line 56: mmmh ? but if one of those is negative, it means that `compute_expanded_shape` has to be fixed. Over here, that shouldn't happen no ?

doctr/transforms/functional/tensorflow.py
```diff
@@ -60,17 +62,16 @@ def rotate(
 
     # Get absolute coords
     _boxes = deepcopy(boxes)
-    if boxes.dtype != int:
+    if np.max(_boxes) < 2:
```
fg-mindee, line 65: same here, I don't understand the max part?

references/detection/train_pytorch.py
```diff
@@ -237,11 +241,14 @@ def main(args):
...
         img_transforms=Compose([
-            T.Resize((args.input_size, args.input_size)),
             # Augmentations
             T.RandomApply(T.ColorInversion(), .1),
             ColorJitter(brightness=0.3, contrast=0.3, saturation=0.3, hue=0.02),
         ]),
+        sample_transforms=T.SampleCompose([
+            T.RandomRotate(90, expand=True),
+            T.ImageTransform(T.Resize((args.input_size, args.input_size))),
+        ]) if args.rotation else T.ImageTransform(T.Resize((args.input_size, args.input_size))),
```
fg-mindee, line 251: I suggest `else None`

references/detection/train_pytorch.py
```diff
@@ -237,11 +241,14 @@ def main(args):
         img_folder=os.path.join(args.train_path, 'images'),
         label_path=os.path.join(args.train_path, 'labels.json'),
         img_transforms=Compose([
-            T.Resize((args.input_size, args.input_size)),
             # Augmentations
             T.RandomApply(T.ColorInversion(), .1),
             ColorJitter(brightness=0.3, contrast=0.3, saturation=0.3, hue=0.02),
         ]),
```
fg-mindee, line 247: if the initial behaviour is still correct for straight boxes, I suggest making this conditional as well (only changing it if args.rotation)

If the initial behaviour was wrong (that rotation should be performed after photometric augmentations), then your design is correct

references/detection/train_pytorch.py
```diff
@@ -180,6 +180,10 @@ def main(args):
         img_folder=os.path.join(args.val_path, 'images'),
         label_path=os.path.join(args.val_path, 'labels.json'),
         img_transforms=T.Resize((args.input_size, args.input_size)),
+        sample_transforms=T.SampleCompose([
+            T.RandomRotate(0, expand=True),
+            T.ImageTransform(T.Resize((args.input_size, args.input_size))),
+        ]) if args.rotation else T.ImageTransform(T.Resize((args.input_size, args.input_size))),
```
fg-mindee, line 186: if args.rotation is False, we'll do double rotation here

references/detection/train_tensorflow.py
```diff
@@ -140,6 +140,10 @@ def main(args):
         img_folder=os.path.join(args.val_path, 'images'),
         label_path=os.path.join(args.val_path, 'labels.json'),
         img_transforms=T.Resize((args.input_size, args.input_size)),
+        sample_transforms=T.SampleCompose([
+            T.RandomRotate(0, expand=True),
+            T.ImageTransform(T.Resize((args.input_size, args.input_size))),
+        ]) if args.rotation else T.ImageTransform(T.Resize((args.input_size, args.input_size))),
```
fg-mindee, line 146: same here

fg-mindee (commented):
Thanks for the PR! 
I raised a concern about the optimal order of transforms, cf my comments :+1:

doctr/datasets/detection.py
```diff
@@ -61,16 +61,25 @@ def __getitem__(
...
         if self.sample_transforms is not None:
+            # Here we may modify coordinates, each transformation must accept/return relative coordinates
+            # Otherwise, if we use the resize operation afterwards it will not only modify images but coordinates
             img, target = self.sample_transforms(img, target)
 
         # Boxes
         target = target.copy()
-        target[..., 0] /= w
-        target[..., 1] /= h
+        if np.max(target) > 2:
```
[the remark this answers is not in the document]
charlesmindee: it is to test if we have relative coords, because when we go trough the rotation transform we end up with relative coords

doctr/transforms/functional/pytorch.py
```diff
@@ -51,17 +51,16 @@ def rotate(
 
     # Get absolute coords
     _boxes = deepcopy(boxes)
-    if boxes.dtype != int:
+    if np.max(_boxes) < 2:
```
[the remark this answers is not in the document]
charlesmindee: to check if we have relative coords, more reliable than dtype = int because we can have absolute coords in float

doctr/transforms/functional/tensorflow.py
charlesmindee: same, check for relative coords

doctr/transforms/functional/tensorflow.py
charlesmindee: If you rotate more than 45 degrees a rectangular document, you may end up with a expanded width smaller than the previous width (because basically you switch the axes), in that case one of the 2 pad should be negative and I think we just have to switch axes for exp_shape what do you think ? In pytorch it is hidden in the PIL transform

references/detection/train_pytorch.py
charlesmindee: No it is right in that case, so what should we do ? do a Compose .... if rotation else Compose .... ?

charlesmindee: Or you want to do double resize when rotation is on ?

doctr/datasets/detection.py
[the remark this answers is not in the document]
fg-mindee: Oooooh, I see. But you don't test > 1, is the test > 2 done out of security with this margin?

references/detection/train_pytorch.py
fg-mindee: I think we can only answer this by experimenting. But until then, by default, let's use the same as before (resize in img transforms when there is no rotation)

We certainly don't want a double resize :sweat_smile:

doctr/transforms/functional/tensorflow.py
fg-mindee: Oh I see, yes the padding should be applied to the opposite axis then I agree!

doctr/datasets/detection.py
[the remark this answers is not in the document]
charlesmindee: I can put 1, I put 2 for security because you will never have a 2 pixel document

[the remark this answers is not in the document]
fg-mindee: But even with 2 pixels doc, the relative coords cannot (shouldn't) go beyond 1? Or am I missing something?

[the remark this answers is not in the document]
charlesmindee: yeah, sure, even if we didn't clip somewhere and we have something slightly upon 1
Do you want me to put 1 ?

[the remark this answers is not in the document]
fg-mindee: Mmmh tricky question:
I'd say we should put > 1, and ensure everywhere that this should be clipped when there is a modification of the relative coords

references/detection/train_pytorch.py
```diff
@@ -179,7 +179,11 @@ def main(args):
     val_set = DetectionDataset(
         img_folder=os.path.join(args.val_path, 'images'),
         label_path=os.path.join(args.val_path, 'labels.json'),
-        img_transforms=T.Resize((args.input_size, args.input_size)),
+        img_transforms=T.Resize((args.input_size, args.input_size)) if not args.rotation else None,
+        sample_transforms=T.SampleCompose([
+            T.RandomRotate(0, expand=True),
```
fg-mindee, line 184: Just to clarify: here we do this convert the boxes to polygons, correct?
If so, shouldn't we remove this and simply pass `use_polygons=True` to the dataset constructor?

references/detection/train_pytorch.py
```diff
@@ -236,12 +240,18 @@ def main(args):
...
+            + [
+                # Augmentations
+                T.RandomApply(T.ColorInversion(), .1),
+                ColorJitter(brightness=0.3, contrast=0.3, saturation=0.3, hue=0.02),
+            ]
+        ),
+        sample_transforms=T.SampleCompose([
+            T.RandomRotate(90, expand=True),
+            T.ImageTransform(T.Resize((args.input_size, args.input_size))),
+        ]) if args.rotation else None
```
fg-mindee, line 254: Would you mind sharing a screenshot in the PR of the output when you add `--show-samples` while using rotation? To make sure the transformation is modifying the text correctly

references/detection/train_tensorflow.py
```diff
@@ -139,7 +139,11 @@ def main(args):
     val_set = DetectionDataset(
         img_folder=os.path.join(args.val_path, 'images'),
         label_path=os.path.join(args.val_path, 'labels.json'),
-        img_transforms=T.Resize((args.input_size, args.input_size)),
+        img_transforms=T.Resize((args.input_size, args.input_size)) if not args.rotation else None,
+        sample_transforms=T.SampleCompose([
+            T.RandomRotate(0, expand=True),
```
fg-mindee, line 144: same question here

references/detection/train_tensorflow.py
```diff
@@ -183,15 +187,21 @@ def main(args):
...
+                T.RandomJpegQuality(60),
+                T.RandomSaturation(.3),
+                T.RandomContrast(.3),
+                T.RandomBrightness(.3),
+            ]
+        ),
+        sample_transforms=T.SampleCompose([
+            T.RandomRotate(90, expand=True),
+            T.ImageTransform(T.Resize((args.input_size, args.input_size))),
+        ]) if args.rotation else None
```
fg-mindee, line 204: No need to add a second visualization, but try to run it locally to make sure that show samples produces also good results in TF

fg-mindee (commented):
Added a few comments!

references/detection/train_pytorch.py
charlesmindee: that is True

references/detection/train_pytorch.py
charlesmindee: Yes I checked all the configurations it is correct, I can attach samples!

charlesmindee:
rotation

references/detection/train_pytorch.py
```diff
@@ -179,11 +179,8 @@ def main(args):
     val_set = DetectionDataset(
         img_folder=os.path.join(args.val_path, 'images'),
         label_path=os.path.join(args.val_path, 'labels.json'),
-        img_transforms=T.Resize((args.input_size, args.input_size)) if not args.rotation else None,
-        sample_transforms=T.SampleCompose([
-            T.RandomRotate(0, expand=True),
-            T.ImageTransform(T.Resize((args.input_size, args.input_size))),
-        ]) if args.rotation else None
+        img_transforms=T.Resize((args.input_size, args.input_size)),
+        use_polygons=True if args.rotation else False
```
fg-mindee, line 183: Be lazy, it's simpler :)
```python
use_polygons=args.rotation,
```

references/detection/train_tensorflow.py
```diff
@@ -139,11 +139,8 @@ def main(args):
     val_set = DetectionDataset(
         img_folder=os.path.join(args.val_path, 'images'),
         label_path=os.path.join(args.val_path, 'labels.json'),
-        img_transforms=T.Resize((args.input_size, args.input_size)) if not args.rotation else None,
-        sample_transforms=T.SampleCompose([
-            T.RandomRotate(0, expand=True),
-            T.ImageTransform(T.Resize((args.input_size, args.input_size))),
-        ]) if args.rotation else None
+        img_transforms=T.Resize((args.input_size, args.input_size)),
+        use_polygons=True if args.rotation else False
```
fg-mindee, line 143: same here

references/detection/train_pytorch.py
fg-mindee: You also need to pass `use_polygons` & adapt the transforms accordingly to this `train_set` as well I think

references/detection/train_tensorflow.py
fg-mindee: same here

fg-mindee (commented):
Almost there!

fg-mindee:
> rotation

Isn't that gorgeous :smile: 
Great work @charlesmindee :+1:

charlesmindee:
I think for the train set it is OK, because if we don't rotate we don't want to use polygons and if we rotate we don't need polygons in the constructor because polygons are created in get_item when rotating boxes

fg-mindee (approved):
Looks good to me!

PR #766, opened 2021-12-28
feat: add preserve aspect ratio to predictor and vizualisation utils

charlesmindee:
This PR allows to use the `preserve_aspect_ratio` flag in the predictor:

```
model = ocr_predictor(pretrained=True, assume_straight_pages=False, preserve_aspect_ratio=True)
doc = DocumentFile.from_images(path)
result = model(doc)
result.show(doc, preserve_aspect_ratio=True)

```

Yields:
ar

Instead of:
withourar

Any feedback is welcome!

### DIFF
--- a/doctr/io/elements.py
+++ b/doctr/io/elements.py
@@ -246,15 +246,16 @@ class Page(Element):
         return f"dimensions={self.dimensions}"
 
     def show(
-        self, page: np.ndarray, interactive: bool = True, **kwargs
+        self, page: np.ndarray, interactive: bool = True, preserve_aspect_ratio: bool = False, **kwargs
     ) -> None:
         """Overlay the result on a given image
 
         Args:
             page: image encoded as a numpy array in uint8
             interactive: whether the display should be interactive
+            preserve_aspect_ratio: pass True if you passed True to the predictor
         """
-        visualize_page(self.export(), page, interactive=interactive)
+        visualize_page(self.export(), page, interactive=interactive, preserve_aspect_ratio=preserve_aspect_ratio)
         plt.show(**kwargs)
 
     def synthesize(self, **kwargs) -> np.ndarray:
--- a/doctr/models/zoo.py
+++ b/doctr/models/zoo.py
@@ -17,6 +17,7 @@ def _predictor(
     reco_arch: str,
     pretrained: bool,
     assume_straight_pages: bool = True,
+    preserve_aspect_ratio: bool = False,
     det_bs: int = 2,
     reco_bs: int = 128,
     **kwargs,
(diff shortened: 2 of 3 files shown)


doctr/io/elements.py
```diff
@@ -246,15 +246,16 @@ def extra_repr(self) -> str:
         return f"dimensions={self.dimensions}"
 
     def show(
-        self, page: np.ndarray, interactive: bool = True, **kwargs
+        self, page: np.ndarray, interactive: bool = True, preserve_aspect_ratio: bool = False, **kwargs
```
fg-mindee, line 249: This can be passed as kwargs

doctr/io/elements.py
```diff
@@ -246,15 +246,16 @@ def extra_repr(self) -> str:
...
     ) -> None:
         """Overlay the result on a given image
 
         Args:
             page: image encoded as a numpy array in uint8
             interactive: whether the display should be interactive
+            preserve_aspect_ratio: pass True if you passed True to the predictor
         """
-        visualize_page(self.export(), page, interactive=interactive)
+        visualize_page(self.export(), page, interactive=interactive, preserve_aspect_ratio=preserve_aspect_ratio)
```
fg-mindee, line 258: same here

doctr/models/zoo.py
```diff
@@ -47,6 +49,7 @@ def ocr_predictor(
     pretrained: bool = False,
     assume_straight_pages: bool = True,
     export_as_straight_boxes: bool = False,
+    preserve_aspect_ratio: bool = False,
```
fg-mindee, line 52: Either we add this one as kwargs or we update the docstring :)

fg-mindee (changes requested):
Thanks, a few improvement suggestions in the review!

doctr/models/zoo.py
charlesmindee: I think we can't really add it as a kwarg because kwargs go to the OCRPredictor and we need to instantiate the detection predictor with that

doctr/io/elements.py
charlesmindee: I think we cant' because kwargs here go to plt.show() and not to visualize, what do you think ?

fg-mindee: Oh my bad, you're right!

doctr/models/zoo.py
fg-mindee: Yes but it will be caught by the signature of OCRPredictor. But in this case, I think we should keep this and only update the docstring by adding the arg description

fg-mindee (approved):
Thanks again!

PR #767, opened 2021-12-28
test: Fixed TensorFlow predictor unittest

fg-mindee:
This PR reduces the number of crops processed in unittests by using a pretrained detection model. It changed the number of crops from several thousands to a maximum of a few dozens.

It also prevents degraded localization candidates, and will hopefully fix the ongoing issue with current PR unittests in TensorFlow.

Any feedback is welcome!


charlesmindee (approved):
Thanks for the PR!

PR #769, opened 2021-12-29
feat: Updated timing to more accurate one

fg-mindee:
This PR simply replaces `time.time()` with `time.perf_counter()` which is much more recommended for latency benchmarks.

Any feedback is welcome!


charlesmindee (approved):
Thanks for that !

PR #775, opened 2021-12-30
refactor: Enforced relative coordinates for all dataset geometries

fg-mindee:
Following up on #750, this PR introduces the following modifications:
- adds a pre_transforms mechanism to perform conversion before any transformation
- adds a abs coord to rel coords function (for both boxes & polygons)
- uses both of these to switch all geometry targets to relative coords
- adds corresponding unittests
- fixed FUNSD when `use_polygons=True`
- fixed `OCRDataset` straight boxes
- added option `use_polygons` in `OCRDataset`

Overrules #702 (cc @felixdittrich92)

Any feedback is welcome!

### DIFF
--- a/doctr/datasets/synthtext.py
+++ b/doctr/datasets/synthtext.py
@@ -7,7 +7,7 @@ import os
 from typing import Any, Dict, List, Tuple
 
 import numpy as np
-from scipy import io as scio
+from scipy import io as sio
 from tqdm import tqdm
 
 from .datasets import VisionDataset
@@ -71,7 +71,7 @@ class SynthText(VisionDataset):
             if not os.path.exists(os.path.join(tmp_root, img_path[0])):
                 raise FileNotFoundError(f"unable to locate {os.path.join(tmp_root, img_path[0])}")
 
-            labels = ''.join(txt).split()
+            labels = [elt for word in txt.tolist() for elt in word.split()]
             word_boxes = word_boxes.transpose(2, 1, 0) if word_boxes.ndim == 3 else np.expand_dims(word_boxes, axis=0)
 
             if not use_polygons:
--- a/doctr/datasets/synthtext.py
+++ b/doctr/datasets/synthtext.py
@@ -70,7 +70,7 @@ class SynthText(VisionDataset):
                 raise FileNotFoundError(f"unable to locate {os.path.join(tmp_root, img_path[0])}")
 
             labels = ''.join(txt).split()
-            word_boxes = word_boxes.transpose(2, 1, 0)
+            word_boxes = word_boxes.transpose(2, 1, 0) if word_boxes.ndim == 3 else np.expand_dims(word_boxes, axis=0)
 
             if not use_polygons:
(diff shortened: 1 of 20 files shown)


doctr/datasets/synthtext.py
```diff
@@ -61,8 +69,8 @@ def __init__(
             if not os.path.exists(os.path.join(tmp_root, img_path[0])):
                 raise FileNotFoundError(f"unable to locate {os.path.join(tmp_root, img_path[0])}")
 
-            labels = [elt for word in txt.tolist() for elt in word.split()]
-            word_boxes = word_boxes.transpose(2, 1, 0) if word_boxes.ndim == 3 else np.expand_dims(word_boxes, axis=0)
+            labels = ''.join(txt).split()
+            word_boxes = word_boxes.transpose(2, 1, 0)
```
felixdittrich92, line 73: @fg-mindee 
I think you cannot change this lines please try to load the dataset after downloading :)

tests/conftest.py
```diff
@@ -344,13 +344,12 @@ def mock_cord_dataset(tmpdir_factory, mock_image_stream):
...
     root = tmpdir_factory.mktemp('datasets')
     synthtext_root = root.mkdir('SynthText')
-    # sub_root = synthtext_root.mkdir('SynthText')
     image_folder = synthtext_root.mkdir("8")
     annotation_file = synthtext_root.join('gt.mat')
     labels = {
         "imnames": [[["8/ballet_106_0.jpg"], ["8/ballet_106_1.jpg"], ["8/ballet_106_2.jpg"]]],
-        "wordBB": [np.random.randint(1000, size=(2, 4, 3)) for _ in range(3)],
-        "txt": np.array([['I      ', 'am\na      ', 'Jedi   ', '!'] for _ in range(3)])
+        "wordBB": [[np.random.randint(1000, size=(2, 4, 5)) for _ in range(3)]],
```
felixdittrich92, line 351: @fg-mindee 
same after reading it is mostly of shape (2, 4, 2)
(But please test on original dataset i have currently not the option 😅 )

felixdittrich92:
@fg-mindee 
Thanks :)
One minor: Can you please fix the missing and/or wrong coments in some datasets (if use polygons) 👍 
They have still # x_center, y_center, width, height, alpha or nothing :)
And unify the np_dtype usage ?

fg-mindee:
> @fg-mindee Thanks :) One minor: Can you please fix the missing and/or wrong coments in some datasets (if use polygons) +1 They have still # x_center, y_center, width, height, alpha or nothing :) And unify the np_dtype usage ?

Yup will do!
About the full set, it really takes ages to download... I have tried with the initial subset you had created and with the mock version. I'll try to find another subset because the download duration keeps on increasing :man_facepalming:

tests/conftest.py
fg-mindee: ? But this is to build the mock dataset, so this is before the constructor parses the data. So it's in format (2, 4, N) where N is the number of words in the image.

In this case, the text includes 5 words, not 3

charlesmindee (approved):
Thanks for the PR!

PR #777, opened 2021-12-31
fix: target DB

charlesmindee:
This PR fixes the `compute_target` of the DB
Any feedback is welcome!


fg-mindee (approved):
Good catch :+1:

PR #779, opened 2021-12-31
feat: Added RandomHorizontalFLip in TF

SiddhantBahuguna:
This PR adds the following changes to the existing PR:
> It adds RandomHorizontalFlip implementation in tensorflow. It transforms both the image and the target(with bboxes in relative coordinates)

          from doctr.transforms import RandomHorizontalFlip
           >>> import tensorflow as tf
           >>> transfo = RandomHorizontalFlip(p=0.5)
           >>> image = tf.random.uniform(shape=[64, 64, 3], minval=0, maxval=1)
           >>> target = {
            "boxes": np.array([[0.1, 0.1, 0.4, 0.5] ], dtype= np.float32),
            "labels": np.ones(1, dtype= np.int64)
            }
           >>> out = transfo(image, target)

> It adds a unitest for the same

> It adds RandomHorizontalFlip in the documentation


doctr/transforms/modules/tensorflow.py
```diff
@@ -375,3 +376,44 @@ def __call__(self, x: tf.Tensor) -> tf.Tensor:
...
+           >>> image = tf.random.uniform(shape=[64, 64, 3], minval=0, maxval=1)
+           >>> target = {
+            "boxes": np.array([[0.1, 0.1, 0.4, 0.5] ], dtype= np.float32),
+            "labels": np.ones(1, dtype= np.int64)
+            }
+           >>> out = transfo(image, target)
+
+       Args:
+           p : probability of Horizontal Flip"""
+    def __init__(self, p):
```
fg-mindee, line 397: don't forget to add type hints here

doctr/transforms/modules/tensorflow.py
```diff
@@ -375,3 +376,44 @@ def __call__(self, x: tf.Tensor) -> tf.Tensor:
...
+        super().__init__()
+        self.p = p
+
+    def __call__(
+            self,
+            img: Union[tf.Tensor, np.ndarray],
```
fg-mindee, line 403: ? are we positive that this works with `np.ndarray`?

doctr/transforms/modules/tensorflow.py
```diff
@@ -375,3 +376,44 @@ def __call__(self, x: tf.Tensor) -> tf.Tensor:
...
+            target: Dict[str, Any]
+    ) -> Tuple[Union[tf.Tensor, np.ndarray], Dict[str, Any]]:
+        """
+        Args:
+            img: Image to be flipped.
+            target: Dictionary with boxes (in relative coordinates of shape (N, 4)) and labels as keys
+        Returns:
+            Tuple of numpy nd-array or Tensor and target
+        """
+        if tf.random.uniform([1], 0, 1) < self.p:
```
fg-mindee, line 413: perhaps switch to numpy

fg-mindee (changes requested):
Thanks! Still need to add the entry in the documentation, but I added a few comments :ok_hand:

doctr/transforms/modules/tensorflow.py
SiddhantBahuguna: Thanks!

doctr/transforms/modules/tensorflow.py
SiddhantBahuguna: Yes, it works with numpy array as well :)

SiddhantBahuguna: Yes, I checked it works with `np.ndarray` as well.

doctr/transforms/modules/tensorflow.py
SiddhantBahuguna: Thanks!

doctr/transforms/modules/tensorflow.py
fg-mindee: And the return type is preserved? (when sending a `np.ndarray`, does it return a `np.ndarray`?)

SiddhantBahuguna: Thanks for the correction. No, the type is not preserved. It returns a tf.tensor. Thanks!

fg-mindee (approved):
Looks good! The failed test is not related and will be fixed by #781 (to be merged before this one)

PR #782, opened 2022-01-03
[datasets] unify np_dtype and fix comments

felixdittrich92:
This PR:

- unify `np_dtype `usage instead of `np.float32`
- fix wrong comments or add missing 
@fg-mindee 
- fix wrong shape in SynthText if ndim != 3


doctr/datasets/cord.py
```diff
@@ -65,14 +66,15 @@ def __init__(
                             x = word["quad"]["x1"], word["quad"]["x2"], word["quad"]["x3"], word["quad"]["x4"]
                             y = word["quad"]["y1"], word["quad"]["y2"], word["quad"]["y3"], word["quad"]["y4"]
                             if use_polygons:
+                                # x1, y1, x2, y2, x3, y3, x4, y4
```
fg-mindee, line 69: I would argue we have to specify this more: (x1, y1) points to which corner, and in what direction is the next one

fg-mindee (commented):
Thanks for the PR!
I agree, however I think the comment about polygons should be more specific. Perhaps something like:
```python
# (x, y) coordinates of top left, top right, bottom right, bottom left corners
```

doctr/datasets/cord.py
felixdittrich92: agree :+1:

doctr/datasets/synthtext.py
```diff
@@ -72,9 +72,12 @@ def __init__(
                 raise FileNotFoundError(f"unable to locate {os.path.join(tmp_root, img_path[0])}")
 
             labels = [elt for word in txt.tolist() for elt in word.split()]
-            word_boxes = word_boxes.transpose(2, 1, 0) if word_boxes.ndim == 3 else np.expand_dims(word_boxes, axis=0)
+            # (x, y) coordinates of top left, top right, bottom right, bottom left corners
+            word_boxes = word_boxes.transpose(2, 1, 0) if word_boxes.ndim == 3 else np.expand_dims(
+                word_boxes.transpose(1, 0), axis=0)
```
fg-mindee, line 77: This isn't related to the PR description: are you adamant that this is correct?

fg-mindee (commented):
Looks good to me! Only a question about a transpose that you changed :ok_hand:

doctr/datasets/synthtext.py
felixdittrich92: @fg-mindee 
Yes i know and yes i have tested it on the original dataset
Before:
(2, 4) -> (1, 2, 4)
After:
(4, 2) -> (1, 4, 2)

fg-mindee (approved):
Thanks Felix!

PR #786, opened 2022-01-06
fix: Clip in rotation transform + eval_straight mode for training

charlesmindee:
This PR fixes 2 things:

- Added a clipping in the transformation rotation so that polygons always stay between 0 and 1
- Added an `eval_straight` option to the training script to evaluate the metrics with straight bounding boxes instead of polygons: we were doing a straight evaluation (no rotation with the eval dataset) but with polygons, which is far more heavy and long to compute, and even crash on most of the computers (masks for polygons in the metric). This option performs the same computation but with straight bboxes, so it is fast and reliable to compare to the metrics we have with no rotation since it is computed the same way.

Any feedback is welcome!


references/detection/train_tensorflow.py
```diff
@@ -113,7 +113,17 @@ def evaluate(model, val_loader, batch_transforms, val_metric):
...
+                # Convert pred to boxes [xmin, ymin, xmax, ymax]  N, 4, 2 --> N, 4
+                boxes_pred = np.stack(
+                    (
+                        np.min(boxes_pred[:, :, 0], 1),
+                        np.min(boxes_pred[:, :, 1], 1),
+                        np.max(boxes_pred[:, :, 0], 1),
+                        np.max(boxes_pred[:, :, 1], 1),
+                    ),
+                    -1
+                )
```
fg-mindee, line 126:
```python
boxes_preds = np.concatenate((boxes_pred.min(axis=1), boxes_pred.max(axis=1)), axis=-1)
```

to use vectorization a bit better :+1:

fg-mindee (changes requested):
Thanks!

There is a file added by mistake I think (doctr/transforms.zip), would you mind removing it from the PR please?

references/detection/train_tensorflow.py
```diff
@@ -328,6 +332,8 @@ def parse_args():
                         help='Load pretrained parameters before starting the training')
     parser.add_argument('--rotation', dest='rotation', action='store_true',
                         help='train with rotated documents')
+    parser.add_argument('--eval_straight', dest='eval_straight', action='store_true',
```
fg-mindee, line 335:
```python
parser.add_argument('--eval-straight', action='store_true',
```
will produce the exact same behaviour :+1:

fg-mindee (commented):
Looks good! 
Sorry, I didn't notice the first time but could you add the same option in the PyTorch script? (so that we keep a good balance)

fg-mindee (approved):
Thanks!

PR #787, opened 2022-01-06
fix: show sample when batch size is 1

charlesmindee:
this PR fixes the show_sample function when the batch size is 1

Any feedback is welcome!


fg-mindee (approved):
Good catch!

PR #793, opened 2022-01-10
feat: add straight-eval arg in evaluate script

charlesmindee:
This PR adds a straight-eval  arg in the evaluate script to allow using the metrics with straight bounding boxes while using a detection predictor working with polygons (assume straight pages can be False).
I also fixed a bug in the predictor to cast to numpy before calling the extraction function (when we use a dataset we have tf/pt tensors otherwise)

Any feedback is welcome!

### DIFF
--- a/doctr/models/predictor/tensorflow.py
+++ b/doctr/models/predictor/tensorflow.py
@@ -74,7 +74,11 @@ class OCRPredictor(NestedObject, _OCRPredictor):
 
         # Localize text elements
         loc_preds = self.det_predictor(pages, **kwargs)
+
         # Crop images
+        if not isinstance(pages[0], np.ndarray):
+            pages = [page.numpy() for page in pages]
+
         crops, loc_preds = self._prepare_crops(
             pages, loc_preds, channels_last=True, assume_straight_pages=self.assume_straight_pages
         )
--- a/doctr/models/predictor/tensorflow.py
+++ b/doctr/models/predictor/tensorflow.py
@@ -76,9 +76,6 @@ class OCRPredictor(NestedObject, _OCRPredictor):
         loc_preds = self.det_predictor(pages, **kwargs)
 
         # Crop images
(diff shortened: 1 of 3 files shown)


fg-mindee:
The CI job for the evaluate script is failing :/

doctr/models/predictor/tensorflow.py
```diff
@@ -74,7 +74,11 @@ def __call__(
 
         # Localize text elements
         loc_preds = self.det_predictor(pages, **kwargs)
+
         # Crop images
+        if not isinstance(pages[0], np.ndarray):
+            pages = [page.numpy() for page in pages]
+
```
fg-mindee, line 81: Isn't that a detour? We convert it to numpy, but it will be converted back to TF/Pytorch tensor afterwards?

fg-mindee (commented):
Looks good! I only added a small question as a comment

doctr/models/predictor/tensorflow.py
charlesmindee: the prepare_crops method needs numpy arrays

fg-mindee: We typed it as such, because it's in base but it's not exactly the case: it works with all sorts of tensors

scripts/evaluate.py
```diff
@@ -75,15 +84,16 @@ def main(args):
                 gt_boxes = np.stack([xmin, ymin, xmax, ymax], axis=-1)
 
             # Forward
+            extraction_fn = extract_crops if args.eval_straight else extract_rcrops
```
fg-mindee, line 87: you can set this out of the loop :+1:

fg-mindee (approved):
Looks good!
I think some of the code of the evaluation script could be refactored, but this could be done in another PR! Either way, we'll need to reharmonize the evaluation script for specific tasks (in references) with this one

fg-mindee (approved):
Cheers!

Issue #730, opened 2021-12-20
[transforms] Extends the list of supported data augmentations


fg-mindee:
As discussed in #654, the artefact detection needs to improve its robustness. In order to do so and prevent overfitting, I would suggest gradually extending the list of our supported transformations:
- [x] uniform photometric transformations (noise, blur, etc.) #758 #654 #768 #771
- [x] non-uniform photometric transformations (shadows, etc.) #811 
- [ ] geometric transformations (flips, rotations, crops, etc.) #773 #799

Ideally, a given transform should be implemented in `doctr/transforms/modules` so that with the corresponding backend, we can do:
```python
transfo = ....
pil_img = ... 
augmented_img = transfo(pil_img)
```
for transformations that only change the image.

and:
```python
transfo = ....
pil_img = ... 
target = {...}
augmented_img, augmented_target = transfo(pil_img, target)
```
for transformations that alter the target. In doing so, that will work nicely with our Dataloaders :+1:

felixdittrich92:
@frgfm only @SiddhantBahuguna PR left or do you have other augmentations in mind ? :)

frgfm:
For now, I think the random perspective would be the last one required yeah :+1:

felixdittrich92:
Hi @SiddhantBahuguna any way in near future that you finish your draft PR ? :)

SiddhantBahuguna:
> Hi @SiddhantBahuguna any way in near future that you finish your draft PR ? :) Or should we keep it free to take ?

Greetings @felixdittrich92 , I am really sorry for the unwanted delay. 
I will get it done this week positively :) Sorry again !

felixdittrich92:
Sounds nice :hugs: :+1:

felixdittrich92:
@SiddhantBahuguna Do you think you could finish your PR that we can close this ? :)

felixdittrich92:
@odulcy-mindee I think we can close this ?
I don't see any issues with different perspectives (detection models) your dataset seems to capture this already well.
Or do you think we should add RandomPerspective as last one ftm ?

odulcy-mindee:
Yeah, this issue can be closed, that's fine

PR #799, opened 2022-01-10
[DRAFT] feat: Added RandomPerspective augmentation

SiddhantBahuguna:
This PR introduces the following changes to the existing repo:

>It adds Random Perspective augmentation that transforms the bboxes along with the image.
 
      from doctr.transforms import RandomPerspective
       import torchvision; import numpy as np; import torch
       transformed = RandomPerspective(0.2, 1, interpolation=torchvision.transforms.functional.InterpolationMode("nearest"))
       target = {"boxes": np.array([[0.1, 0.1, 0.4, 0.5] ], dtype= np.float32), "labels": np.ones(1, dtype= np.int64)}
       image, target = transformed(torch.rand((3, 224, 224)), target) 
2x3_artefact

> It adds unittest for the same. The entire page is taken as a bbox to test the transformation of the image as well as the target bounding boxes

General outline

1.   A mask(with zeros) is computed with the same shape as of the original image
2.    Bboxes are drawn over the mask and filled with respective color for each label (this will be helpful to us in label identification after the transform)
3.    Both mask and original image are transformed
4.    Mask of each label is obtained by filtering the color ranges
5.    Contours of each mask are calculated and later using the extreme points, the new coordinates are obtained

Any feedback is welcome!

### DIFF
--- a/doctr/transforms/modules/pytorch.py
+++ b/doctr/transforms/modules/pytorch.py
@@ -6,13 +6,15 @@
 import math
 from typing import Any, Dict, Tuple, Union
 
+import cv2
+import numpy as np
 import torch
 from PIL.Image import Image
 from torch.nn.functional import pad
 from torchvision.transforms import functional as F
 from torchvision.transforms import transforms as T
 
-__all__ = ['Resize', 'GaussianNoise', 'ChannelShuffle', 'RandomHorizontalFlip']
+__all__ = ['Resize', 'GaussianNoise', 'ChannelShuffle', 'RandomHorizontalFlip', 'RandomPerspective']
 
 
 class Resize(T.Resize):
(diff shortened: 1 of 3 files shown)


SiddhantBahuguna:
To be added:
Unittest for the same in the same PR soon

SiddhantBahuguna:
Unittest(Issues & Errors):

`AttributeError: module 'torchvision.transforms.functional' has no attribute '_get_image_num_channels'`
There is some mismatch between the versions of torchvision as aparently _get_image_num_channels is no longer an attribute and I think is changed to get_image_num_channels. Strangely, test ran perfectly fine locally. 

Also, I need to think of a better way to calculate the transformed bounding box coordinates.

SiddhantBahuguna:
> Unittest(Issues & Errors):
> 
> `AttributeError: module 'torchvision.transforms.functional' has no attribute '_get_image_num_channels'` There is some mismatch between the versions of torchvision as aparently _get_image_num_channels is no longer an attribute and I think is changed to get_image_num_channels. Strangely, test ran perfectly fine locally.
> 
> Also, I need to think of a better way to calculate the transformed bounding box coordinates.

I overrode the methods and it fixed the issue. Since, we arent and we havent tested for PIL image in the unittest, codcov couldnt cover all the part of the code.

doctr/transforms/modules/pytorch.py
```diff
@@ -119,3 +121,93 @@ def forward(
...
+
+            #  transformed mask
+            transformed_mask = F.perspective(
+                mask, startpoints, endpoints, self.interpolation, fill
+            ).permute(1, 2, 0).numpy()
+
+            new_img = F.perspective(img, startpoints, endpoints, self.interpolation, fill)
+
+            #  filtering to separate bboxes of each label
+            mask_qr = cv2.inRange(transformed_mask, np.array([220, 220, 220]), np.array([240, 240, 240]))
```
SiddhantBahuguna, line 179: There are some operations in the PR that could be accelerated using multiprocessing such as:

        with Pool() as p:
            masks = p.map(
                    self.filter_color,
                    [
                        (transformed_mask, np.array([220, 220, 220]), np.array([240, 240, 240])),
                        (transformed_mask, np.array([180, 180, 180]), np.array([200, 200, 200])),
                        (transformed_mask, np.array([160, 160, 160]), np.array([180, 180, 180])),
                        (transformed_mask, np.array([240, 240, 240]), np.array([255, 255, 255])),
                    ]
            )
However, I am facing:
`AssertionError: daemonic processes are not allowed to have children`

felixdittrich92: @SiddhantBahuguna 
Threading brings no advantage ?
Else have you tried (py3.8) import concurrent.futures.ProcessPoolExecutor as Pool ?

SiddhantBahuguna: @felixdittrich92 
Thanks for taking time to resolve this :) 
I gave concurrent.futures.ProcessPoolExecutor a try, and the same Assertion Error popped up. May be we require/employ something non-daemonic for the workers process to avoid this.

fg-mindee:
Sorry for the late reply! 
I'll have to review this carefully, by any chance, have you checked https://pytorch.org/vision/stable/transforms.html#torchvision.transforms.RandomPerspective ?

I think it would be a good start to try to inherit as much as possible for there? (which will avoid cv2 deps hopefully)

SiddhantBahuguna:
> Sorry for the late reply! I'll have to review this carefully, by any chance, have you checked https://pytorch.org/vision/stable/transforms.html#torchvision.transforms.RandomPerspective ?
> 
> I think it would be a good start to try to inherit as much as possible for there? (which will avoid cv2 deps hopefully)

Hi, I have actually inherited as much as possible from the same. The mask is what is first converted to a torch.Tensor and then back to numpy array for the exploration of contours. I just realized that we have some operations of drawing bounding boxes in torchvision, and so I will make some changes to it although we will still have some cv2 operations. Will update the current PR first with some first hand improvements. Thanks!

fg-mindee:
Any update @SiddhantBahuguna ? :)

fharper:
@SiddhantBahuguna is this PR still good? If yes, can you review FG's comments, modify in consequences so we can potentially merge it please?

felixdittrich92:
@SiddhantBahuguna any updates ? :)

SiddhantBahuguna:
> @SiddhantBahuguna any updates ? :)



> @SiddhantBahuguna is this PR still good? If yes, can you review FG's comments, modify in consequences so we can potentially merge it please?

Hi Fred, 
This PR makes one assumption that might be specific to our generated dataset. And that is why, I have to make modifications plus have to optimize it further. We can either take this PR down and I will open a new one soon or modify the existing one.. :)

SiddhantBahuguna:
> @SiddhantBahuguna any updates ? :)

Hi Felix, 
Thanks for your message :) 
I will redo few elements of it. It should be done before next week.
Thanks.

fharper:
@SiddhantBahuguna: you tell me what you prefer, but for now, I'll move this PR in a draft state :)

felixdittrich92:
@SiddhantBahuguna same here any update ? 😅

frgfm:
To put this into context, this transformation is one of the most natural one we can find on pictures taken of documents such as recipes. So if we manage to integrate it in our trainings, I'm sure that will help a lot with robustness (also on the artefact detection part actually)

felixdittrich92:
Hi @SiddhantBahuguna would ask again if there are updates ? :)

felixdittrich92:
#730

PR #801, opened 2022-01-13
fix: polygon orientation + line aggregation

charlesmindee:
This PR fixes the polygon orientations, which was bot left, top left, top rigth, bot right instead of top left, top rigth, bot right, bot left.
This fixes the line aggregation.
Any feedback is welcome!

### DIFF
--- a/doctr/utils/geometry.py
+++ b/doctr/utils/geometry.py
@@ -230,12 +230,13 @@ def estimate_page_angle(polys: np.ndarray) -> float:
     """Takes a batch of rotated previously ORIENTED polys (N, 4, 2) (rectified by the classifier) and return the
     estimated angle ccw in degrees
     """
-    x1 = (polys[:, 0, 0] + polys[:, 3, 0]) / 2
-    y1 = (polys[:, 0, 1] + polys[:, 3, 1]) / 2
-    x2 = (polys[:, 1, 0] + polys[:, 2, 0]) / 2
-    y2 = (polys[:, 1, 1] + polys[:, 2, 1]) / 2
+    # Compute mean left points and mean right point with respect to the reading direction (oriented polygon)
+    xleft = polys[:, 0, 0] + polys[:, 3, 0]
+    yleft = polys[:, 0, 1] + polys[:, 3, 1]
+    xright = polys[:, 1, 0] + polys[:, 2, 0]
+    yright = polys[:, 1, 1] + polys[:, 2, 1]
     return np.median(np.arctan(
-        (y1 - y2) / (x2 - x1)  # Y axis from top to bottom!
+        (yleft - yright) / (xright - xleft)  # Y axis from top to bottom!
     )) * 180 / np.pi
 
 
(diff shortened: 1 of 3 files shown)


doctr/utils/geometry.py
```diff
@@ -230,9 +230,12 @@ def estimate_page_angle(polys: np.ndarray) -> float:
...
     estimated angle ccw in degrees
     """
+    x1 = (polys[:, 0, 0] + polys[:, 3, 0]) / 2
+    y1 = (polys[:, 0, 1] + polys[:, 3, 1]) / 2
+    x2 = (polys[:, 1, 0] + polys[:, 2, 0]) / 2
+    y2 = (polys[:, 1, 1] + polys[:, 2, 1]) / 2
     return np.median(np.arctan(
-        (polys[:, 0, 1] - polys[:, 1, 1]) /  # Y axis from top to bottom!
-        (polys[:, 1, 0] - polys[:, 0, 0])
+        (y1 - y2) / (x2 - x1)  # Y axis from top to bottom!
```
fg-mindee, line 238: Since we divide them, the `/2` isn't necessary

I'm not entirely sure, but shouldn't be `np.arctan((polys[:, 3, 1] - polys[:, 0, 1]) / (polys[:, 1, 0] - polys[:, 0, 0]))`?
Introducing the average by summing won't help (on the contrary) in my opinion :thinking:

fg-mindee (commented):
Thanks! Sounds good for the _utils, but I think there is an issue in geometry

doctr/utils/geometry.py
charlesmindee: the average is to find a middle ground when the polygon is not rectangular (or at least when the 2 sides on top and bottom in the reading direction are not parallel), we need to take the mean angle of those 2 vectors instead of choosing arbitrarily one of the sides.
Here, x1 is the "mean x left", x2 the "mean x right" with respect to the reading direction (ordered polygon) and y1 is the "mean y left", y2 the "mean y right" (always if you consider the crop to be in the reading direction), and indeed the / 2 is not necessary

doctr/utils/geometry.py
```diff
@@ -230,9 +230,13 @@ def estimate_page_angle(polys: np.ndarray) -> float:
...
     """
+    # Compute mean left points and mean right point with respect to the reading direction (oriented polygon)
+    xleft = polys[:, 0, 0] + polys[:, 3, 0]
+    yleft = polys[:, 0, 1] + polys[:, 3, 1]
+    xright = polys[:, 1, 0] + polys[:, 2, 0]
+    yright = polys[:, 1, 1] + polys[:, 2, 1]
     return np.median(np.arctan(
-        (polys[:, 0, 1] - polys[:, 1, 1]) /  # Y axis from top to bottom!
-        (polys[:, 1, 0] - polys[:, 0, 0])
+        (yleft - yright) / (xright - xleft)  # Y axis from top to bottom!
```
fg-mindee, line 239: Now this looks quite logical, but I think we're missing a unittest somewhere. This will change the existing behaviour for sure, but for some reasons, the unittests are still passing. Could you check please? :pray:

charlesmindee: I added a test!

charlesmindee: But this doesn't really modify the behaviour of this function, it is just more accurate/robust to skewed cases so I guess it is expected that we don't break unitests here

fg-mindee: I meant: that change will change the output value for a given input if the polygon is not a straight bounding box

fg-mindee (approved):
thanks!

PR #803, opened 2022-01-18
fix: linknet target computation

charlesmindee:
This PR fixes the linknet target computation in the case we have polygons

Any feedback is welcome


fg-mindee (approved):
Good catch, thanks!

PR #805, opened 2022-01-18
feat: add boosted edges for rotated pages

charlesmindee:
This PR adds boosted edges for rotated polygons: we successively draw a dilated polygon and fill the inside of the shrank polygon with 0, so that we drew the edges. We can choose the edges thickness with 2 parameters, which are for the moment set to 1.5 and 0.4 (such as in the DB paper).

Any feedback is welcome!

### DIFF
--- a/doctr/models/detection/differentiable_binarization/base.py
+++ b/doctr/models/detection/differentiable_binarization/base.py
@@ -224,7 +224,7 @@ class _DBNet:
         padding.AddPath(subject, pyclipper.JT_ROUND, pyclipper.ET_CLOSEDPOLYGON)
         padded_polygon = np.array(padding.Execute(distance)[0])
 
-        # Fill the mask with 1 on the new shrinked polygon
+        # Fill the mask with 1 on the new shrank polygon
         cv2.fillPoly(mask, [padded_polygon.astype(np.int32)], 1.0)
 
         # Get min/max to recover polygon after distance computation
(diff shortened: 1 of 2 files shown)


doctr/models/detection/differentiable_binarization/base.py
```diff
@@ -224,7 +224,7 @@ def draw_thresh_map(
         padding.AddPath(subject, pyclipper.JT_ROUND, pyclipper.ET_CLOSEDPOLYGON)
         padded_polygon = np.array(padding.Execute(distance)[0])
 
-        # Fill the mask with 1 on the new shrinked polygon
+        # Fill the mask with 1 on the new shrank polygon
```
fg-mindee, line 227: it's actually "shrunken" in english :)

doctr/models/detection/linknet/base.py
```diff
@@ -137,24 +140,45 @@ def build_target(
...
+                    if len(expanded_points) >= 1:
+                        cv2.fillPoly(edge_mask[idx], [expanded_points.astype(np.int32)[0]], 1)
+
+                        # Shrink with pyclipper to unfill the inside
+                        distance = area * (1 - np.power(0.4, 2)) / length
+                        padding = pyclipper.PyclipperOffset()
+                        padding.AddPath(points, pyclipper.JT_ROUND, pyclipper.ET_CLOSEDPOLYGON)
+                        shrank_polygon = np.array(padding.Execute(distance))
+                        if len(shrank_polygon) >= 1:
+                            cv2.fillPoly(edge_mask[idx], [shrank_polygon.astype(np.int32)[0]], 0)
```
fg-mindee, line 181: I agree that this could be a nice addition, but we have to test that it actually helps I guess?
(The original linknet didn't build the target that way, so let's train something first with this branch)

fg-mindee (commented):
Thanks!

I think this is the type of modification where we need to train the model with this branch to ensure that we don't merge something that would decrease perf :ok_hand:

doctr/models/detection/linknet/base.py
charlesmindee: yes I tested it and I looked at the masks, and it seems to make the convergence easier, and besides before this when using polygons the loss computation was throwing an error because the edge_mask was not instantiated before and the method was still trying returning it, so I need we need to compute this mask anyway since the method returns it

charlesmindee:
Closing this since we will remove the boosted edge feature in the loss and add a dice loss term

PR #808, opened 2022-01-18
feat: Added support of GPU for predictors in PyTorch

fg-mindee:
This PR simply adds a dynamic device selection for predictor inference.
This snippet:
```python
import os

os.environ['USE_TORCH'] = '1'

doc = DocumentFile.from_pdf('/path/to/sample.pdf').as_images()
predictor = ocr_predictor(pretrained=True).cuda()

out = predictor(doc)
```
used to yield:
```
RuntimeError: Input type (torch.FloatTensor) and weight type (torch.cuda.FloatTensor) should be the same or input should be a MKLDNN tensor and weight is a dense tensor
```
because the model is on cuda, while the inputs are never moved to GPU.
And now, the snippet executes perfectly (& much faster than on CPU)

Since the predictions returned by each postprocessor are in numpy (automatically moved to CPU), this doesn't break any previous behaviour :) 

Any feedback is welcome!

### DIFF
--- a/doctr/models/detection/predictor/pytorch.py
+++ b/doctr/models/detection/predictor/pytorch.py
@@ -44,8 +44,9 @@ class DetectionPredictor(nn.Module):
             raise ValueError("incorrect input shape: all pages are expected to be multi-channel 2D images.")
 
         processed_batches = self.pre_processor(pages)
+        _device = next(self.model.parameters()).device
         predicted_batches = [
-            self.model(batch, return_preds=True, **kwargs)['preds']  # type:ignore[operator]
+            self.model(batch.to(device=_device), return_preds=True, **kwargs)['preds']  # type:ignore[operator]
             for batch in processed_batches
         ]
         return [pred for batch in predicted_batches for pred in batch]
--- a/doctr/models/recognition/predictor/pytorch.py
+++ b/doctr/models/recognition/predictor/pytorch.py
@@ -70,8 +70,9 @@ class RecognitionPredictor(nn.Module):
         processed_batches = self.pre_processor(crops)
 
         # Forward it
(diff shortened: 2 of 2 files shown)


charlesmindee (approved):
Thanks for the fix

gganes3:
When using a pretrained model, using map_location="gpu" throws an error
RuntimeError: don't know how to restore data location of torch.FloatStorage (tagged with gpu)
        det_params = torch.load(r"C:\Projects\db_resnet50-ac60cadc.pt", map_location="gpu")
        reco_params = torch.load(r"C:\Projects\crnn_vgg16_bn-9762b0b0.pt", map_location="gpu")
        # det_model.load_state_dict()
        det_model.load_state_dict(det_params)
        reco_model.load_state_dict(reco_params)

        # Ask the preprocessor of each task to resize and normalize similarly to your training
        # cf. https://github.com/mindee/doctr/blob/main/references/detection/train_pytorch.py#L94 & https://github.com/mindee/doctr/blob/main/references/detection/train_pytorch.py#L109
        det_predictor = DetectionPredictor(
            PreProcessor((128, 1024), batch_size=1, mean=(0.798, 0.785, 0.772), std=(0.264, 0.2749, 0.287)), det_model)
        # cf. https://github.com/mindee/doctr/blob/main/references/recognition/train_pytorch.py#L97 & https://github.com/mindee/doctr/blob/main/references/recognition/train_pytorch.py#L111
        reco_predictor = RecognitionPredictor(
            PreProcessor((32, 128), preserve_aspect_ratio=True, batch_size=32, mean=(0.694, 0.695, 0.693),
                         std=(0.299, 0.296, 0.301)), reco_model)


        model = OCRPredictor(det_predictor, reco_predictor).cuda()

PR #809, opened 2022-01-18
refactor: Avoids instantiating orientation predictor when unnecessary

fg-mindee:
- simply saves a bit of memory by avoiding to instantiating the crop orientation predictor when it will never be used.
- refactors predictor constructor (moving as many things as possible to base): that way, kwargs of `DocumentBuilder` can be directly passed to the high-level API `ocr_predictor`
- updates unittests to check that the crop orientation predictor is only there when needed

Any feedback is welcome!


charlesmindee (dismissed):
thanks!

PR #811, opened 2022-01-20
feat: Added shadow augmentation for all backends

fg-mindee:
Following up on #730, this PR introduces the following modifications:
- adds `RandomShadow` augmentation
- adds unittests
- adds documentation

The following snippet:
```python
from doctr.io.image import read_img_as_tensor
from doctr.transforms import RandomShadow

transform = RandomShadow((.2, .8))
img = read_img_as_tensor("path/to/your/image.jpg")
plt.imshow(transform(img).numpy()); plt/axis('off'); plt.show()
```
yields:
shadow_sample

Briefly:
- this first creates a 4-point polygon with 1 side on the top edge of the image
- the polygon is rotated & expanded
- the polygon is filled with 1s
- we apply an intensity gradient (so that the shadow fades as we go further from the source)
- with TF or PyTorch we overlay this after applying gaussian blur (a sharp shadow is not very believable)

Apart from filling the polygons, everything is in pure numpy until we overlay with the tensor image!

Any feedback is welcome!

### DIFF
--- a/doctr/transforms/functional/base.py
+++ b/doctr/transforms/functional/base.py
@@ -62,7 +62,7 @@ def expand_line(line: np.ndarray, target_shape: Tuple[int, int]) -> Tuple[int, i
         return line[0]
     # Get the line equation
     _tmp = line[1] - line[0]
-    _valid = _tmp > 0
+    _direction = _tmp > 0
     alpha = _tmp[1] / _tmp[0]
     beta = line[1, 1] - alpha * line[1, 0]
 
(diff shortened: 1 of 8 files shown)


doctr/transforms/functional/base.py
```diff
@@ -42,3 +45,102 @@ def crop_boxes(
     is_valid = np.logical_and(boxes[:, 1] < boxes[:, 3], boxes[:, 0] < boxes[:, 2])
 
     return boxes[is_valid]
+
+
+def expand_line(line: np.ndarray, target_shape: Tuple[int, int]) -> Tuple[int, int]:
```
charlesmindee, line 50: can you add minimal comments here to describe what the function does please ?

charlesmindee (commented):
Thanks for this tranform, only a small comment

doctr/transforms/functional/base.py
fg-mindee: good idea!

fg-mindee:
Ready for review @charlesmindee @SiddhantBahuguna :grin:

charlesmindee (approved):
Thanks!

Issue #800, opened 2022-01-11
Restore remap_boxes function


Rob192:
### Bug description

Following #731, the function `remap_boxes` and its associated test `test_remap_boxes` that were introduced in #488 were removed. This prevents the correct remapping of the boxes after they were rotated (parameter `straighten_pages` of `OCRPredictor` set to `True`). 

### Code snippet to reproduce the bug

```python
from pathlib import Path

import matplotlib.pyplot as plt

from doctr.io import DocumentFile
from doctr.models.detection.zoo import detection_predictor
from doctr.models.predictor import OCRPredictor
from doctr.models.recognition.zoo import recognition_predictor
from doctr.utils.visualization import visualize_page

file = Path(r"myfile.jpg")

doc = DocumentFile.from_images(file)
det_bs=2
reco_bs=128

det_predictor = detection_predictor('db_resnet50', pretrained=True, batch_size=det_bs, assume_straight_pages=True)
reco_predictor = recognition_predictor('crnn_vgg16_bn', pretrained=True, batch_size=reco_bs)

predictor = OCRPredictor(det_predictor=det_predictor, reco_predictor=reco_predictor, assume_straight_pages=True, straighten_pages=True)

# Plot OCR output
out = predictor(doc)

fig = visualize_page(out.pages[0].export(), doc[0], interactive=False)
plt.show()
```

### Error traceback

No error, however, the boxes that appear are misplaced. 

### Environment

Note: I pulled the last version from the repo but the DocTR version still appears as `0.4.1a0`. 

```
DocTR version: 0.4.1a0
TensorFlow version: 2.5.0
PyTorch version: 1.9.1+cpu (torchvision 0.10.1+cpu)
OpenCV version: 4.5.2
OS: Could not collect
Python version: 3.8.7
Is CUDA available (TensorFlow): No
Is CUDA available (PyTorch): No
CUDA runtime version: 11.0.221
GPU models and configuration: Could not collect
Nvidia driver version: Could not collect
cuDNN version: C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v11.0\bin\cudnn_ops_train64_8.dll
```

fg-mindee:
Hi @Rob192,

Thanks for opening the issue :pray: 
So I suggest doing the following in a PR:
- starting by designing a relevant unittest where we mock a white page with only a few lines (easy)
[...]

Issue #788, opened 2022-01-10
Release tracker - v0.5.1


fg-mindee:
This issue is to be used to track the roadmap of docTR for release v0.5.1.

- Models
    - [ ] Bridge the gap between TensorFlow & PyTorch support of text detection on rotation
- Transforms
    - [x] Harmonize `RandomRotate` between PyTorch & TensorFlow #784 
    - [x] Add missing TensorFlow support for docTR transforms #779

charlesmindee:
Closing this since the released has been done

felixdittrich92:
@charlesmindee 
pin sar/master fix to 0.6.0 ? :)

PR #812, opened 2022-01-25
Restore remap boxes

Rob192:
This PR is to address #800 as part of #788. 

It restores the function `remap_boxes` from #488 and adds a new functionality test to verify the output of a pretrained predictor on a rotated document.

### DIFF
--- a/doctr/utils/geometry.py
+++ b/doctr/utils/geometry.py
@@ -132,7 +132,7 @@ def remap_boxes(
     orig_shape: Tuple[int, int],
     dest_shape: Tuple[int, int]
 ) -> np.ndarray:
-    """ Remaps a batch of rotated locpred (x, y, w, h, alpha, c) expressed for an origin_shape to a destination_shape.
+    """ Remaps a batch of rotated locpred (N, 4, 2) expressed for an origin_shape to a destination_shape.
     This does not impact the absolute shape of the boxes, but allow to calculate the new relative RotatedBbox
     coordinates after a resizing of the image.
     Args:
--- a/doctr/utils/geometry.py
+++ b/doctr/utils/geometry.py
@@ -150,13 +150,6 @@ def remap_boxes(
     orig_height, orig_width = orig_shape
     dest_height, dest_width = dest_shape
     mboxes = loc_preds.copy()
-    # # remaps position of the box center for the destination image shape
-    # mboxes[:, 0] = ((loc_preds[:, 0] * orig_width) + (dest_width - orig_width) / 2) / dest_width
-    # mboxes[:, 1] = ((loc_preds[:, 1] * orig_height) + (dest_height - orig_height) / 2) / dest_height
-    # # remaps box dimension for the destination image shape
-    # mboxes[:, 2] = loc_preds[:, 2] * orig_width / dest_width
-    # mboxes[:, 3] = loc_preds[:, 3] * orig_height / dest_height
-
     mboxes[:, :, 0] = ((loc_preds[:, :, 0] * orig_width) + (dest_width - orig_width) / 2) / dest_width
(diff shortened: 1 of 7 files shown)


doctr/utils/geometry.py
```diff
@@ -127,11 +127,41 @@ def rotate_abs_geoms(
     return rotated_polys
 
 
+def remap_boxes(
+    loc_preds: np.ndarray,
+    orig_shape: Tuple[int, int],
+    dest_shape: Tuple[int, int]
+) -> np.ndarray:
+    """ Remaps a batch of rotated locpred (x, y, w, h, alpha, c) expressed for an origin_shape to a destination_shape.
```
charlesmindee, line 135: Is this a typo in the docstring ? The function should work with the new box format [[x1, y1], [x2, y2], ...]]

doctr/utils/geometry.py
```diff
@@ -127,11 +127,41 @@ def rotate_abs_geoms(
...
+    This does not impact the absolute shape of the boxes, but allow to calculate the new relative RotatedBbox
+    coordinates after a resizing of the image.
+    Args:
+        loc_preds: (N, 6) array of RELATIVE locpred (x, y, w, h, alpha, c)
+        orig_shape: shape of the origin image
+        dest_shape: shape of the destination image
+    Returns:
+        A batch of rotated loc_preds (N, 6): (x, y, w, h, alpha, c) expressed in the destination referencial
```
charlesmindee, line 143: same here, you need to return boxes in the new format

doctr/utils/geometry.py
```diff
@@ -127,11 +127,41 @@ def rotate_abs_geoms(
...
+    """
+
+    if len(dest_shape) != 2:
+        raise ValueError(f"Mask length should be 2, was found at: {len(dest_shape)}")
+    if len(orig_shape) != 2:
+        raise ValueError(f"Image_shape length should be 2, was found at: {len(orig_shape)}")
+    orig_height, orig_width = orig_shape
+    dest_height, dest_width = dest_shape
+    mboxes = loc_preds.copy()
+    mboxes[:, :, 0] = ((loc_preds[:, :, 0] * orig_width) + (dest_width - orig_width) / 2) / dest_width
```
charlesmindee, line 153: I am not sure I understand well the computation here: If we have relative coordinates [[x1, y1], [x2, y2], [x3, y3], ...] we just need to multiply by dest_width & dest_height to get the absolute points coordinates in dest_shape referential, or am I missing something ?

charlesmindee (commented):
Hi @Rob192, thank you for your PR. 
I added a few comments because I am not sure I fully understood the purpose of the remap_boxes function since we moved to a new format of boxes where boxes are described by points and not angles. In any case, can you attach in the issue #800 the image of the document where the boxes are misplaced to help us understanding the bug ? :pray: Thank you!

doctr/utils/geometry.py
Rob192: Hi @charlesmindee, no you need to take into account the padding that was introduced by `rotate_image`. This is the main point of this function `remap_boxes` that is here to accommodate these differences.

doctr/utils/geometry.py
Rob192: indeed, this is a typo !

doctr/utils/geometry.py
Rob192: typo indeed

doctr/utils/geometry.py
Rob192: I added pictures for reference in #800

charlesmindee: Ok, thanks I see the point

charlesmindee (approved):
thanks for this fix!

PR #816, opened 2022-02-09
feat: add dice loss in linknet

charlesmindee:
This PR adds dice loss in the linknet loss (which is now bce_loss + dice_loss) and remove the boosted edges part of the loss since it hasn't led to successful trainings. Any feedback is welcome!

### DIFF
--- a/doctr/models/detection/linknet/pytorch.py
+++ b/doctr/models/detection/linknet/pytorch.py
@@ -184,7 +184,7 @@ class LinkNet(nn.Module, _LinkNet):
         dice_loss = 1 - 2 * inter / (cardinality + 1e-8)
 
         # Only consider contributions overlaping the mask
-        return bce_loss[seg_mask].mean() + dice_loss.mean()
+        return bce_loss[seg_mask].mean() + dice_loss
 
 
 def _linknet(
--- a/doctr/models/detection/linknet/tensorflow.py
+++ b/doctr/models/detection/linknet/tensorflow.py
@@ -164,7 +164,7 @@ class LinkNet(_LinkNet, keras.Model):
         cardinality = tf.math.reduce_sum(prob_map[seg_mask] + seg_target[seg_mask])
         dice_loss = 1 - 2 * inter / (cardinality + 1e-8)
 
-        return tf.math.reduce_mean(bce_loss[seg_mask]) + tf.math.reduce_mean(dice_loss)
+        return tf.math.reduce_mean(bce_loss[seg_mask]) + dice_loss
 
     def call(
         self,


doctr/models/detection/linknet/pytorch.py
```diff
@@ -158,34 +158,33 @@ def compute_loss(
...
-        # Get the cross_entropy for each entry
-        loss = F.binary_cross_entropy_with_logits(out_map, seg_target, reduction='none')
+        # BCE loss
+        bce_loss = F.binary_cross_entropy_with_logits(out_map, seg_target, reduction='none')
+
+        # Dice loss
+        prob_map = torch.nn.functional.sigmoid(out_map)
+        inter = (prob_map[seg_mask] * seg_target[seg_mask]).sum()
+        cardinality = (prob_map[seg_mask] + seg_target[seg_mask]).sum()
+        dice_loss = 1 - 2 * inter / (cardinality + 1e-8)
```
fg-mindee, line 184: we'll have to update this in another PR because it highly advantages classes that occupy the biggest area

doctr/models/detection/linknet/pytorch.py
```diff
@@ -158,34 +158,33 @@ def compute_loss(
...
 
-        # Compute BCE loss with highlighted edges
-        if edge_factor > 0:
-            loss = ((1 + (edge_factor - 1) * edge_mask) * loss)
         # Only consider contributions overlaping the mask
-        return loss[seg_mask].mean()
+        return bce_loss[seg_mask].mean() + dice_loss.mean()
```
fg-mindee, line 187: the dice loss is already a scalar (since you used a bare `.sum()`), so no need for the `.mean()` on that one

doctr/models/detection/linknet/tensorflow.py
```diff
@@ -139,37 +139,32 @@ def compute_loss(
...
-                loss
-            )
+        # Dice loss
+        prob_map = tf.math.sigmoid(out_map)
+        inter = tf.math.reduce_sum(prob_map[seg_mask] * seg_target[seg_mask])
+        cardinality = tf.math.reduce_sum(prob_map[seg_mask] + seg_target[seg_mask])
+        dice_loss = 1 - 2 * inter / (cardinality + 1e-8)
 
-        return tf.reduce_mean(loss[seg_mask])
+        return tf.math.reduce_mean(bce_loss[seg_mask]) + tf.math.reduce_mean(dice_loss)
```
fg-mindee, line 167: same here, the dice loss is already a scalar

fg-mindee (dismissed):
Thanks! Added a few optional suggestions

doctr/models/detection/linknet/pytorch.py
charlesmindee: Yes, I will do it in the next PR !

SiddhantBahuguna: To curb this, may be FL + Dice Loss could be a better option? Because, even if the area was same for the classes, since total area of combined bboxes / total area of image could be really less, it may tend to learn more negatives.

SiddhantBahuguna (approved):

doctr/models/detection/linknet/pytorch.py
fg-mindee: Yup, I meant that the baseline option would be to average the loss over classes (not mixing the contribution of each class to the loss before the last step). But your suggestion @SiddhantBahuguna is worth exploring as a follow-up and could improve results :+1:

charlesmindee: Thanks for your suggestions, I will add focal loss in the next PR !

fg-mindee (approved):
Looks good!

PR #818, opened 2022-02-15
feat: replace PyMuPDF  by pdf2image, for license compatibility

### LINKED ISSUE #113: [conda] Unable to make a conda build
Unfortunately, one of the project dependencies does not have any conda release or any way to make one. I opened an issue on their repo https://github.com/pymupdf/PyMuPDF/issues/938 to track this, but so far I haven't found any way to release the project on anaconda with this dependency.

### LINKED ISSUE #486: Consider removing PyMuPDF for dependency that is not AGPL licensed
Hi folks!  Love the project, but the dependency on PyMuPDF can pose a problem for commercial use because of it's AGPL license (I think use of PyMuPDF may technically require the project itself to be GPL licensed, although I'm no expert here). Would you consider integrating a PR that replaces the PyMuPDF dependency with an alternative?

charlesmindee:
This PR replaces PyMuPDF by pdf2image, which is under MIT license.
This library allows to convert a pdf (from both path and stream) to a list of images.
We loose the text extraction feature for source pdf, but we didn't use it so far so it won't change anything.

Any feedback is welcome!
closes #486 
closes #113

### DIFF
--- a/demo/app.py
+++ b/demo/app.py
@@ -52,7 +52,7 @@ def main():
     uploaded_file = st.sidebar.file_uploader("Upload files", type=['pdf', 'png', 'jpeg', 'jpg'])
     if uploaded_file is not None:
         if uploaded_file.name.endswith('.pdf'):
-            doc = DocumentFile.from_pdf(uploaded_file.read()).as_images()
+            doc = DocumentFile.from_pdf(uploaded_file.read())
         else:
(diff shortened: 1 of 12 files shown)


felixdittrich92:
@charlesmindee 👋 
Do you have compared with pdfminer.six which is also under MIT but in this case without loosing anything !? 😄

PR #822, opened 2022-02-16
feat: add shrinked target in linknet + dilation in postprocessing

charlesmindee:
Similarly to the DB, this PR includes a shrink of the polygons in the target and a dilation in the postprocessing of the linknet, to prevent boxes to be merged on the heatmap because their boundaries are too close. 
After some experiments, this lead to much better results.
Any feedback is welcome!

### DIFF
--- a/doctr/models/detection/linknet/base.py
+++ b/doctr/models/detection/linknet/base.py
@@ -121,7 +121,7 @@ class LinkNetPostProcessor(DetectionPostProcessor):
 
             if score < self.box_thresh:   # remove polygons with a weak objectness
                 continue
-            
+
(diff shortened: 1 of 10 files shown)


fg-mindee (commented):
Perhaps we could do this in a smoother way by sharing the box shrink+dilation ops between the DB & LinkNet?

Because I feel like we're only adding duplicate lines of code here :sweat_smile:

charlesmindee:
I modified the typo, for the code duplicates there is only the polygon_to_box part in common, I suggest refactoring that in another PR

fg-mindee (approved):
Thanks Charles!

PR #823, opened 2022-02-16
feat: add target to resize transform for aspect ratio training (detection task)

charlesmindee:
This PR adds the optional arg `target` in the Resize transform to train detection models preserving the aspect ratio which leads to much consistent results. 
The double resize in the script is mandatory because first we resize and pad to keep the aspect ratio (and we modify the labels as well), then we rotate everything, and finally we need to resize the image so that each image has the same size and can be batched.
Any feedback is welcome!


doctr/transforms/modules/pytorch.py
```diff
@@ -41,11 +47,27 @@ def forward(self, img: torch.Tensor) -> torch.Tensor:
...
+            # In case boxes are provided, resize boxes if needed (for detection task if preserve aspect ratio)
+            if target is not None:
+                if self.preserve_aspect_ratio:
+                    # Get absolute coords
+                    if target.shape[1:] == (4,):
+                        target[:, [0, 2]] *= raw_shape[-1] / self.size[1]
+                        target[:, [1, 3]] *= raw_shape[-2] / self.size[0]
+                    elif target.shape[1:] == (4, 2):
+                        target[..., 0] *= raw_shape[-1] / self.size[1]
+                        target[..., 1] *= raw_shape[-2] / self.size[0]
```
fg-mindee, line 66: If the aspect ratio is preserved, there might be some shifting (padding) to be done on the target as well I'd argue

I might be missing something but it looks like it's missing the padding step

references/detection/train_pytorch.py
```diff
@@ -243,17 +243,18 @@ def main(args):
...
-        sample_transforms=T.SampleCompose([
-            T.RandomRotate(90, expand=True),
-            T.ImageTransform(T.Resize((args.input_size, args.input_size))),
-        ]) if args.rotation else None,
+        sample_transforms=T.SampleCompose(
+            [T.Resize((args.input_size, args.input_size), preserve_aspect_ratio=True)]
+            + ([T.RandomRotate(90, expand=True),
+                T.ImageTransform(T.Resize((args.input_size, args.input_size), preserve_aspect_ratio=True))
+                ] if args.rotation else [])
+        ),
```
fg-mindee, line 257: For the sake of reducing the amount of interpolation being done, perhaps we should start with the expanded rotation first then the resize

fg-mindee (commented):
Thanks for the PR!

Two concerns I have:
- I might be wrong but it looks like the target isn't being padded after the scaling (when preserving aspect ratio)? have you checked what it looks like with `show-samples`? (mind adding a screenshot if so? :pray: )
- I added a comment about the transformation order

Let me know what you think!

references/detection/train_pytorch.py
charlesmindee: Unfortunately with our GPU's size we need to resize before the rotation otherwise the interpolation on very big images make the training crash for big angles

doctr/transforms/modules/pytorch.py
charlesmindee: I checked with `show-samples`, it should be OK. 
What is done here is just a changing of referential, we need to multiply the coordinate of each point by the ratio old_shape / new_shape, the old shape being the shape before padding (for instance, if you pad half of the picture, you want to divide by 2 all x and y coordinates)

charlesmindee: ratio

references/detection/train_pytorch.py
fg-mindee: alright I thought that could be reason :sweat_smile: 
fair enough

doctr/transforms/modules/pytorch.py
fg-mindee: I perfectly agree with you about getting the shape & target before padding. But in the next lines, the image is padded, but the target remains unchanged and I'm concerned about that part (again I might be missing something) :upside_down_face:

charlesmindee: I think it is just a linear transformation since the padding is not centered (this is why the content is not centered). You have to de-correlate the rotation and the resizing, for the right and left samples we first perform a resizing with padding to keep the aspect ratio, and then we rotate the sample, so the outputs are as expected. But as you suggests, we could remove the padding in the first resizing (this resizing is just to be sure we aren't exploding the memory), then we could rotate and finally resize and pad preserving the aspect ratio.

charlesmindee: For the second point, I think I should indeed add the boxes to the transformation

fg-mindee: Oh yeah, I see, now it makes sense!
So I assume that on the first sample, it was padded on the bottom or on the right, but the rotation made it look like it's from the top right?

Either way, I would argue the first and last samples have an issue: they are padded on 2 sides (we could have expanded the content more while preserving the aspect ratio). Is that because of the double resize? If so, we should find a better way, and I may have a suggestion, say we want to resize to (W, H):
- the first resize should resize the minimum side of that image to max(W,H), (just scaling the image so that the shortest side is equal to max(W, H)) without padding anything
- we then do a rotation while adding expand=True
- then we do a resize, this time by preserving aspect ratio (thus padding)

What do you think?

Also: should we try to make it centered though?

charlesmindee: yes

charlesmindee: This should be OK now!
aspect ratio and no symmetric padding:
nosym
aspect ratio and symmetric padding:
sym
no aspect ratio:
noratio

charlesmindee: I tested all the options in both tf & pt, by default I put conservation of the aspect ratio & symmetric padding in the scripts

fg-mindee (commented):
Almost there!

So since we already have a few flags related to this, I think it would be best to avoid having a new flag in the transform:
- if preserve_aspect_ratio is True and the original & target shapes don't have the same aspect ratio, then there will have to be padding
- to scale the image, in other libraries, they usually pass an integer instead of a tuple and desired shape. I'd argue we could have a good use for a similar mechanism


What do you think?

charlesmindee:
I am not sure I understand your second point, can you develop please ? 
For the first point, we can have a target size of 1024, 1024 and a different source aspect ratio, we reshape preserving the aspect ratio to (1024, X) or (X, 1024) but we still don't want to pad to avoid adding unnecessary information on the image before the rotation (for now this is the case in` references/detection/train_tensorflow.py` l 204) ?

doctr/transforms/modules/pytorch.py
```diff
@@ -27,26 +28,78 @@ def __init__(
...
 
-    def forward(self, img: torch.Tensor) -> torch.Tensor:
-        target_ratio = self.size[0] / self.size[1]
-        actual_ratio = img.shape[-2] / img.shape[-1]
-        if not self.preserve_aspect_ratio or (target_ratio == actual_ratio):
+    def forward(
+        self,
+        img: torch.Tensor,
+        target: Optional[np.ndarray] = None,
+    ) -> Union[torch.Tensor, Tuple[torch.Tensor, np.ndarray]]:
```
fg-mindee, line 35: Thinking about it, I think we should have:
- one function to resize the image
- one function to resize the target
- use them in the module implementation
And I'm not sure what would be the best, but probably having `Resize` (image only), and `SampleResize` (image + target), would be relevant. What do you think?

fg-mindee (commented):
Thanks! I left a few other comments

doctr/transforms/modules/pytorch.py
charlesmindee: Yes we can split the function with 1 for the targets and 1 for the image, but I am not sure it is relevant to have a` Resize` and a `SampleResize`, because if you want to keep the aspect ratio of the image while resizing, you will almost always pad (except if you give only 1 target size), and this will modify the targets. The ` Resize`  function won't be able to preserve the aspect ratio of the images, and this would need to differentiate the function we are using in the training scripts and in the preprocessor regarding the option selected by the user (`preserve_aspect_ratio` or not). This can even be dangerous if someone tries to modify the aspect ratio in ` Resize`  without changing the targets I think, do yo agree ?

charlesmindee: And for a differentiation of the target and image function, I think it will complexify the code because we use many attributes of the `Resize` class to resize both the image and the target (`self.symmetric_pad`, `self.preserve_aspect_ratio`, `self.output_size`) which would be passed as arguments and we have also many shared computations for the image and the target that would need to be done twice (or added in the signature of the functions but I think it will make the code less understandable): `offset`,` raw_shape,` and `img.shape` is also used in the target computation, what do you think ?

fg-mindee: About splitting the two transforms, the reason behind this is that the target, depending on the task, will not necessarily be modified. So either we can add a lot of cases (for each target type) in the Resize transform, or we could have a Resize that only transforms the image, and `SampleResize` that inherits from it and transforms the target.

I agree this would be dangerous for something to add this without changing the target, but transforms will only be played with by people willing to train models. So I would argue it's safe to assume they either use our default training script or have a good understanding of what is going on under the hood :man_shrugging: 

However, I fully agree about the duplication of symmetric pad / computation duplication for some aspects. I don't have an ideal suggestion for this

charlesmindee: For the `SampleResize` transform, you want it to inherit from `Resize`, but since they won't have the same signature so I don't really see why we do that ? @fg-mindee

doctr/transforms/modules/pytorch.py
```diff
@@ -29,26 +30,78 @@ def __init__(
...
+
+        if isinstance(self.size, int):
+            target_ratio = img.shape[-2] / img.shape[-1]
+            actual_ratio = img.shape[-2] / img.shape[-1]
+        elif isinstance(self.size, tuple) or isinstance(self.size, list):
+            target_ratio = self.size[0] / self.size[1]
+            actual_ratio = img.shape[-2] / img.shape[-1]
+        else:
+            raise AssertionError("size should be either a tuple or an int")
```
fg-mindee, line 46: Let's move this check in the constructor :)

doctr/transforms/modules/pytorch.py
fg-mindee: We can always refactor this later on anyway! Feel free to implement it the way you prefer (the optional target passing you suggested is probably the best one)

doctr/transforms/modules/pytorch.py
```diff
@@ -29,26 +30,78 @@ def __init__(
...
+
+        if not self.preserve_aspect_ratio or (
+            target_ratio == actual_ratio and (isinstance(self.size, tuple) or isinstance(self.size, list))
+        ):
+            if target is not None:
+                return super().forward(img), target
             return super().forward(img)
```
fg-mindee, line 53: More of a question: shouldn't we return the target in both cases?
Because the transform will be used in `sample_transforms` most likely, creating a change of arg/return signatures could bring about a few issues later on

doctr/transforms/modules/pytorch.py
```diff
@@ -29,26 +30,78 @@ def __init__(
...
         else:
             # Resize
```
fg-mindee, line 55: Could you add extra info in this comment? (computing the size without padding to use regular resize first)

doctr/transforms/modules/pytorch.py
```diff
@@ -29,26 +30,78 @@ def __init__(
...
+    def forward(
+        self,
+        img: torch.Tensor,
+        target: Optional[np.ndarray] = None,
+    ) -> Union[torch.Tensor, Tuple[torch.Tensor, np.ndarray]]:
+
+        if isinstance(self.size, int):
+            target_ratio = img.shape[-2] / img.shape[-1]
+            actual_ratio = img.shape[-2] / img.shape[-1]
+        elif isinstance(self.size, tuple) or isinstance(self.size, list):
```
fg-mindee, line 42:
```python
elif isinstance(self.size, (tuple, list)):
```

doctr/transforms/modules/pytorch.py
```diff
@@ -29,26 +30,78 @@ def __init__(
...
+            if target is not None:
+                return super().forward(img), target
             return super().forward(img)
         else:
             # Resize
-            if actual_ratio > target_ratio:
-                tmp_size = (self.size[0], max(int(self.size[0] / actual_ratio), 1))
-            else:
-                tmp_size = (max(int(self.size[1] * actual_ratio), 1), self.size[1])
+            if isinstance(self.size, tuple) or isinstance(self.size, list):
```
fg-mindee, line 56: if isinstance(self.size, (tuple, list)):

doctr/transforms/modules/pytorch.py
```diff
@@ -29,26 +30,78 @@ def __init__(
...
             # Scale image
             img = F.resize(img, tmp_size, self.interpolation)
-            # Pad (inverted in pytorch)
-            _pad = (0, self.size[1] - img.shape[-1], 0, self.size[0] - img.shape[-2])
-            if self.symmetric_pad:
-                half_pad = (math.ceil(_pad[1] / 2), math.ceil(_pad[3] / 2))
-                _pad = (half_pad[0], _pad[1] - half_pad[0], half_pad[1], _pad[3] - half_pad[1])
-            return pad(img, _pad)
+            raw_shape = img.shape[-2:]
+            if isinstance(self.size, tuple) or isinstance(self.size, list):
```
fg-mindee, line 70: elif isinstance(self.size, (tuple, list)):

doctr/transforms/modules/pytorch.py
```diff
@@ -29,26 +30,78 @@ def __init__(
...
+                    half_pad = (math.ceil(_pad[1] / 2), math.ceil(_pad[3] / 2))
+                    _pad = (half_pad[0], _pad[1] - half_pad[0], half_pad[1], _pad[3] - half_pad[1])
+                img = pad(img, _pad)
+
+            # In case boxes are provided, resize boxes if needed (for detection task if preserve aspect ratio)
+            if target is not None:
+                if self.preserve_aspect_ratio:
+                    # Get absolute coords
+                    if target.shape[1:] == (4,):
+                        if (isinstance(self.size, tuple) or isinstance(self.size, list)) and self.symmetric_pad:
```
fg-mindee, line 83: `if isinstance(self.size, (tuple, list)) and self.symmetric_pad:`

doctr/transforms/modules/pytorch.py
```diff
@@ -29,26 +30,78 @@ def __init__(
...
+                            if np.max(target) <= 1:
+                                offset = half_pad[0] / img.shape[-1], half_pad[1] / img.shape[-2]
+                            target[:, [0, 2]] = offset[0] + target[:, [0, 2]] * raw_shape[-1] / img.shape[-1]
+                            target[:, [1, 3]] = offset[1] + target[:, [1, 3]] * raw_shape[-2] / img.shape[-2]
+                        else:
+                            target[:, [0, 2]] *= raw_shape[-1] / img.shape[-1]
+                            target[:, [1, 3]] *= raw_shape[-2] / img.shape[-2]
+                    elif target.shape[1:] == (4, 2):
+                        if (isinstance(self.size, tuple) or isinstance(self.size, list)) and self.symmetric_pad:
```
fg-mindee, line 92: if isinstance(self.size, (tuple, list)) and self.symmetric_pad:

doctr/transforms/modules/tensorflow.py
```diff
@@ -77,18 +77,63 @@ def extra_repr(self) -> str:
...
+        self,
+        img: tf.Tensor,
+        target: Optional[np.ndarray] = None,
+    ) -> Union[tf.Tensor, Tuple[tf.Tensor, np.ndarray]]:
+
         input_dtype = img.dtype
-        img = tf.image.resize(img, self.output_size, self.method, self.preserve_aspect_ratio)
+        if isinstance(self.output_size, int):
+            wanted_size = (self.output_size, self.output_size)
+        elif isinstance(self.output_size, tuple) or isinstance(self.output_size, list):
```
fg-mindee, line 89:
```python
elif isinstance(self.output_size, (tuple, list)):
```

doctr/transforms/modules/tensorflow.py
```diff
@@ -77,18 +77,63 @@ def extra_repr(self) -> str:
...
+            wanted_size = self.output_size
+        else:
+            raise AssertionError("Output size should be either a list, a tuple or an int")
```
fg-mindee, line 92: Let's move this to the constructor (it's strange to throw an error about constructor arg only when we perform inference)

doctr/transforms/modules/tensorflow.py
```diff
@@ -77,18 +77,63 @@ def extra_repr(self) -> str:
...
         if self.preserve_aspect_ratio:
-            # pad width
-            if not self.symmetric_pad:
-                offset = (0, 0)
-            elif self.output_size[0] == img.shape[0]:
-                offset = (0, int((self.output_size[1] - img.shape[1]) / 2))
-            else:
-                offset = (int((self.output_size[0] - img.shape[0]) / 2), 0)
-            img = tf.image.pad_to_bounding_box(img, *offset, *self.output_size)
+            if (isinstance(self.output_size, tuple) or isinstance(self.output_size, list)):
```
fg-mindee, line 97:
```python
if isinstance(self.output_size, (tuple, list)):
```

doctr/transforms/modules/tensorflow.py
```diff
@@ -77,18 +77,63 @@ def extra_repr(self) -> str:
...
+                img = tf.image.pad_to_bounding_box(img, *offset, *self.output_size)
+
+        # In case boxes are provided, resize boxes if needed (for detection task if preserve aspect ratio)
+        if target is not None:
+            if self.preserve_aspect_ratio:
+                # Get absolute coords
+                if target.shape[1:] == (4,):
+                    if (
+                        isinstance(self.output_size, tuple) or isinstance(self.output_size, list)
+                    ) and self.symmetric_pad:
```
fg-mindee, line 114:
```python
if isinstance(self.output_size, (tuple, list)) and self.symmetric_pad:
```

doctr/transforms/modules/tensorflow.py
```diff
@@ -77,18 +77,63 @@ def extra_repr(self) -> str:
...
+                            offset = offset[0] / img.shape[0], offset[1] / img.shape[1]
+                        target[:, [0, 2]] = offset[1] + target[:, [0, 2]] * raw_shape[1] / img.shape[1]
+                        target[:, [1, 3]] = offset[0] + target[:, [1, 3]] * raw_shape[0] / img.shape[0]
+                    else:
+                        target[:, [0, 2]] *= raw_shape[1] / img.shape[1]
+                        target[:, [1, 3]] *= raw_shape[0] / img.shape[0]
+                elif target.shape[1:] == (4, 2):
+                    if (
+                        isinstance(self.output_size, tuple) or isinstance(self.output_size, list)
+                    ) and self.symmetric_pad:
```
fg-mindee, line 125:
```python
if isinstance(self.output_size, (tuple, list)) and self.symmetric_pad:
```

fg-mindee (changes requested):
Minor adjustments to do!

Also, perhaps in a later PR, but our Resize transform is starting to get complex. We should add an illustration similarly to https://mindee.github.io/doctr/transforms.html#doctr.transforms.RandomRotate

doctr/transforms/modules/pytorch.py
charlesmindee: `img ` is not in the constructor

doctr/transforms/modules/pytorch.py
charlesmindee: do you mean returning `None` when there is no target ? 
For the moment if you pass img, target it returns img, target and if you pass img it returns img

fg-mindee (approved):
Cheers :pray:

PR #824, opened 2022-02-16
feat: replace bce by focal loss in linknet loss

charlesmindee:
Following a suggestion by @fg-mindee and @SiddhantBahuguna, this PR replaces  the BCE loss by the Focal loss in the linknet loss to increase the recall (imbalanced classes)

Any feedback is welcome!


doctr/models/detection/linknet/pytorch.py
```diff
@@ -174,17 +178,26 @@ def compute_loss(
         seg_target, seg_mask = torch.from_numpy(seg_target).to(dtype=out_map.dtype), torch.from_numpy(seg_mask)
         seg_target, seg_mask = seg_target.to(out_map.device), seg_mask.to(out_map.device)
 
-        # BCE loss
+        # Focal loss
         bce_loss = F.binary_cross_entropy_with_logits(out_map, seg_target, reduction='none')
+        if gamma and gamma < 0:
```
fg-mindee, line 183: gamma is not None by default and a float, so I think we can safely keep `if gamma < 0:`

doctr/models/detection/linknet/pytorch.py
```diff
@@ -174,17 +178,26 @@ def compute_loss(
...
+            raise ValueError("Value of gamma should be greater than or equal to zero.")
+        # Convert logits to prob, compute gamma factor
+        pred_prob = torch.sigmoid(out_map)[seg_mask]
```
fg-mindee, line 186: Let's avoid doing the computation twice (cf. `prob_map` of the dice loss, which we could put before so that it can be masked here)

doctr/models/detection/linknet/pytorch.py
```diff
@@ -174,17 +178,26 @@ def compute_loss(
...
+        p_t = (seg_target[seg_mask] * pred_prob) + ((1 - seg_target[seg_mask]) * (1 - pred_prob))
+        # Compute alpha factor
+        alpha_factor = seg_target[seg_mask] * alpha + (1 - seg_target[seg_mask]) * (1 - alpha)
+        # compute the final loss
+        focal_loss = (alpha_factor * (1. - p_t) ** gamma * bce_loss[seg_mask]).mean()
```
fg-mindee, line 191: I think we need to address the masking + reduction problem of the dice loss first: once masked, this reduces the tensor to something in 1D. So if one class has 10 times more masked region than another, this will be a problem

I'd suggest doing the following: changing
```python
seg_target[mask].mean()
```
to
```python
mask = mask.to(dtype=torch.foat32)
# Average on N, H, W
class_loss = (seg_target * mask).sum((0, 2, 3)) / mask.sum((0, 2, 3))
loss = class_loss.mean()
```

or average it on H, W only before the final mean

fg-mindee (commented):
Thanks, I added a comment on the implementation! Have you tried to check if it improves training perf?

doctr/models/detection/linknet/pytorch.py
charlesmindee: I am not sure I understand well here: do you want to remove each ...[mask] occurrence ?

fg-mindee: So the difference is the following:
- `my_tensor[mask]` is 1D tensor with a number of elements = number of `True` in the mask
- `my_tensor * mask.to(dtype=torch.float32)` has the shape shape as `my_tensor`, it only puts zero on elements that are masked out

Now if you perform a reduction operation like mean:
- in the first case, you divide by the number of elements in `my_tensor[mask]` = `mask.sum()`
- in the second one, you dive by the number of elements in `my_tensor`

And this extends to dimension-specific operations, so if we mask it, we lose the separation of the dimensions to get a contiguous tensor in the end. To properly scale the loss, in the first case, this widely increases the contribution of classes with the highest amount of positive mask (makes no difference if there is only a single class). And since we specifically want to help balance the less-frequent classes here, I suggest leveraging that second option with my suggestion above :+1:

fg-mindee: (and we need to do this for both the dice loss and the focal loss)

it will only make a difference in cases of multi-class and where the mask isn't only `True` but that might be safer!

Either way, I think we should run a training with the configuration to make sure this yields a positive change :+1:

doctr/models/detection/linknet/pytorch.py
```diff
@@ -177,22 +177,27 @@ def compute_loss(
...
         seg_target, seg_mask = seg_target.to(out_map.device), seg_mask.to(out_map.device)
-        proba_map = torch.sigmoid(out_map)[seg_mask]
+        seg_mask = seg_mask.to(dtype=torch.float32)
 
-        # Focal loss
         bce_loss = F.binary_cross_entropy_with_logits(out_map, seg_target, reduction='none')
+        # Average on N, H, W
+        bce_loss = (bce_loss * seg_mask).sum((0, 2, 3)) / seg_mask.sum((0, 2, 3))
+        proba_map = (torch.sigmoid(out_map) * seg_mask).sum((0, 2, 3)) / seg_mask.sum((0, 2, 3))
+        seg_target = (seg_target * seg_mask).sum((0, 2, 3)) / seg_mask.sum((0, 2, 3))
```
fg-mindee, line 186: The mask should be applied after the computation of the unreduced loss. You can remove this

doctr/models/detection/linknet/pytorch.py
```diff
@@ -177,22 +177,27 @@ def compute_loss(
...
+        alpha_factor = seg_target * alpha + (1 - seg_target) * (1 - alpha)
         # compute the final loss
-        focal_loss = (alpha_factor * (1. - p_t) ** gamma * bce_loss[seg_mask]).mean()
+        focal_loss = (alpha_factor * (1. - p_t) ** gamma * bce_loss).mean()
 
         # Dice loss
-        inter = (proba_map * seg_target[seg_mask]).sum()
-        cardinality = (proba_map + seg_target[seg_mask]).sum()
+        inter = (proba_map * seg_target).sum()
+        cardinality = (proba_map + seg_target).sum()
```
fg-mindee, line 200:
```python
inter = (seg_mask * probs_map * target).sum((0, 1, 2))
cardinality = (seg_mask * (probs_map + target)).sum((0, 1, 2))
```

doctr/models/detection/linknet/pytorch.py
```diff
@@ -177,22 +177,27 @@ def compute_loss(
...
         dice_loss = 1 - 2 * inter / (cardinality + 1e-8)
```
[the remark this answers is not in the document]
fg-mindee: (and you'll have to average over the class before returning:
```python
return focal_loss.mean() + dice_loss.mean()
```

doctr/models/detection/linknet/pytorch.py
```diff
@@ -177,22 +177,27 @@ def compute_loss(
...
             raise ValueError("Value of gamma should be greater than or equal to zero.")
         # Convert logits to prob, compute gamma factor
-        p_t = (seg_target[seg_mask] * proba_map) + ((1 - seg_target[seg_mask]) * (1 - proba_map))
+        p_t = (seg_target * proba_map) + ((1 - seg_target) * (1 - proba_map))
         # Compute alpha factor
-        alpha_factor = seg_target[seg_mask] * alpha + (1 - seg_target[seg_mask]) * (1 - alpha)
+        alpha_factor = seg_target * alpha + (1 - seg_target) * (1 - alpha)
         # compute the final loss
-        focal_loss = (alpha_factor * (1. - p_t) ** gamma * bce_loss[seg_mask]).mean()
+        focal_loss = (alpha_factor * (1. - p_t) ** gamma * bce_loss).mean()
```
fg-mindee, line 196:
```python
p_t = (1 - proba_map) * seg_target + proba_map * (1 - seg_target)
# Unreduced version
focal_loss = p_t ** gamma * bce_loss
# Class reduced
focal_loss = (seg_mask * focal_loss).sum((0, 1, 2)) / seg_mask.sum((0, 1, 2))
```

doctr/models/detection/linknet/pytorch.py
```diff
@@ -177,22 +177,27 @@ def compute_loss(
 
         seg_target, seg_mask = torch.from_numpy(seg_target).to(dtype=out_map.dtype), torch.from_numpy(seg_mask)
         seg_target, seg_mask = seg_target.to(out_map.device), seg_mask.to(out_map.device)
-        proba_map = torch.sigmoid(out_map)[seg_mask]
+        seg_mask = seg_mask.to(dtype=torch.float32)
 
-        # Focal loss
         bce_loss = F.binary_cross_entropy_with_logits(out_map, seg_target, reduction='none')
```
fg-mindee, line 182: That's where alpha intervenes:
```python
bce_loss = F.binary_cross_entropy_with_logits(out_map, seg_target, reduction='none', pos_weight=alpha * torch.ones(out_map.shape[1], dtype=torch.float32, device=out_map.device)))
```

Alpha is the positive weight term in the loss (cf. https://pytorch.org/docs/stable/generated/torch.nn.BCEWithLogitsLoss.html)

fg-mindee: For the TF implementation though, there isn't such parameter

fg-mindee (changes requested):
A few corrections in the loss computation and we'll be good to go!

doctr/models/detection/linknet/pytorch.py
```diff
@@ -173,18 +178,30 @@ def compute_loss(
...
+            pos_weight=alpha * torch.ones(out_map.shape[1], dtype=torch.float32, device=out_map.device),
+        )
+        proba_map = torch.sigmoid(out_map)
+
+        # Focal loss
+        if gamma < 0:
+            raise ValueError("Value of gamma should be greater than or equal to zero.")
+        p_t = (1 - proba_map) * seg_target + proba_map * (1 - seg_target)
+        # Unreduced version
+        focal_loss = p_t ** gamma * bce_loss
```
fg-mindee, line 194: To stick to the paper:
```python
p_t = proba_map * seg_target + (1 - proba_map) * (1 - seg_target)
# Unreduced version
focal_loss = (1 - p_t) ** gamma * bce_loss
```
(it's strictly equivalent)

doctr/models/detection/linknet/tensorflow.py
```diff
@@ -154,17 +159,30 @@ def compute_loss(
...
+            raise ValueError("Value of gamma should be greater than or equal to zero.")
+        # Convert logits to prob, compute gamma factor
+        p_t = (seg_target * proba_map) + ((1 - seg_target) * (1 - proba_map))
+        modulating_factor = tf.pow((1.0 - p_t), gamma)
+        # Compute alpha factor
+        alpha_factor = seg_target * alpha + (1 - seg_target) * (1 - alpha)
+        # compute the final loss
+        focal_loss = tf.reduce_sum(
+            seg_mask * alpha_factor * modulating_factor * bce_loss, (0, 1, 2)
+        ) / tf.reduce_sum(seg_mask, (0, 1, 2))
```
fg-mindee, line 178: So that we keep the same structure with the PyTorch implementation and stick with paper notations
```python
alpha = seg_target * alpha + (1 - seg_target) * (1 - alpha)
# Unreduced loss
focal_loss = alpha * (1 - p_t) ** gamma * bce_loss
# Class reduced
focal_loss = tf.reduce_sum(seg_mask * focal_loss, (0, 1, 2)) / tf.reduce_sum(seg_mask, (0, 1, 2))
```

doctr/models/detection/linknet/pytorch.py
```diff
@@ -188,18 +193,30 @@ def compute_loss(
...
         seg_target, seg_mask = torch.from_numpy(seg_target).to(dtype=out_map.dtype), torch.from_numpy(seg_mask)
         seg_target, seg_mask = seg_target.to(out_map.device), seg_mask.to(out_map.device)
+        seg_mask = seg_mask.to(dtype=torch.float32)
 
-        # BCE loss
-        bce_loss = F.binary_cross_entropy_with_logits(out_map, seg_target, reduction='none')
+        bce_loss = F.binary_cross_entropy_with_logits(
+            out_map, seg_target, reduction='none',
+            pos_weight=alpha * torch.ones(out_map.shape[1], dtype=torch.float32, device=out_map.device),
+        )
```
fg-mindee, line 201: As discussed offline, my bad, with this design, we only apply alpha to the positive contribution to the loss, while we have to apply (1 -alpha) to the negative part:
```python
bce_loss = bce_loss = F.binary_cross_entropy_with_logits(out_map, seg_target, reduction='none')
```

doctr/models/detection/linknet/pytorch.py
```diff
@@ -188,18 +193,30 @@ def compute_loss(
...
+        proba_map = torch.sigmoid(out_map)
+
+        # Focal loss
+        if gamma < 0:
+            raise ValueError("Value of gamma should be greater than or equal to zero.")
+        p_t = proba_map * seg_target + (1 - proba_map) * (1 - seg_target)
+        # Unreduced version
+        focal_loss = (1 - p_t) ** gamma * bce_loss
```
fg-mindee, line 209:
```python
alpha_t = alpha * seg_target + (1 - alpha) * (1 - seg_target)
# Unreduced version
focal_loss = alpha_t * (1 - p_t) ** gamma * bce_loss
```

doctr/models/detection/linknet/tensorflow.py
```diff
@@ -166,17 +171,28 @@ def compute_loss(
...
+        proba_map = tf.sigmoid(out_map)
+
+        # Focal loss
+        if gamma < 0:
+            raise ValueError("Value of gamma should be greater than or equal to zero.")
+        # Convert logits to prob, compute gamma factor
+        p_t = (seg_target * proba_map) + ((1 - seg_target) * (1 - proba_map))
+        alpha = seg_target * alpha + (1 - seg_target) * (1 - alpha)
+        # Unreduced loss
+        focal_loss = alpha * (1 - p_t) ** gamma * bce_loss
```
fg-mindee, line 186: To avoid modifying alpha inplace:
```python
alpha_t = seg_target * alpha + (1 - seg_target) * (1 - alpha)
# Unreduced loss
focal_loss = alpha_t * (1 - p_t) ** gamma * bce_loss
```

fg-mindee (changes requested):
My bad, a few fixes to do on my previous suggestion

fg-mindee (approved):
Thanks!

PR #826, opened 2022-02-18
[FIX] german vocab

felixdittrich92:
normally there is no usage of a uppercase `ß`.
on the position where would be an uppercase `ß ` is replaced with `SS`
also i have found no font which supports an uppercase `ß`.


fg-mindee (changes requested):
Thanks Felix!
Actually, to make sure, I just checked and this is not a duplication: one of them lower case, the other is upper case.

You might think that if there is no visual diff, then they are the same but actually it entirely depends on the font (if you do the string comparison in python, you can clearly see the difference). 

So my bad, but actually we can close this PR :sweat_smile:

felixdittrich92:
@fg-mindee 
I saw it was added 2017 to the german vocab. (I have never seen before :sweat_smile: )
But normally we wrote SS instead of an uppercase ß. 
Example: street -> straße -> STRASSE 
I think then the problem is more that there are not many fonts which supports this character. 
I have tested some common like Arial, Helvetica, OpenSans

PR #827, opened 2022-02-21
[datasets] Add MJSynth (Synth90K)

felixdittrich92:
This PR integrates the ability to load the MJSynth dataset [MJSynth](https://www.robots.ox.ac.uk/~vgg/data/text/)
This one is the first pure recognition dataset with ~ 9M english word images

It´s used in many other repositories to train a recognition model (in some cases in addition with crops from SynthText (maybe prepare this later)) before this will be evaluated on IC03, SVT, SVHN, and so on.

I have decided to integrate it not with the download link in fact that the http download takes ~48hrs and the BitTorrent download is a lot faster.

sample:
download


doctr/datasets/mjsynth.py
```diff
@@ -0,0 +1,70 @@
...
+            raise FileNotFoundError(
+                f"unable to locate {labels_path if not os.path.exists(labels_path) else img_folder}")
+
+        self.data: List[Tuple[str, Dict[str, str]]] = []
+        self.train = train
+
+        with open(labels_path) as f:
+            img_paths = f.readlines()
+
+        train_samples = int(len(img_paths) * 0.85)
```
fg-mindee, line 60: we've been using 0.9 so far, let's keep it that way for now :+1:

doctr/datasets/mjsynth.py
```diff
@@ -0,0 +1,70 @@
...
+        set_slice = slice(train_samples) if self.train else slice(train_samples, None)
+
+        for path in tqdm(iterable=img_paths[set_slice], desc='Unpacking MJSynth', total=len(img_paths[set_slice])):
+            label = path.split('_')[1]
+            img_path = os.path.join(img_folder, path[2:]).strip()
+
+            self.data.append((img_path, dict(labels=label)))
```
fg-mindee, line 67: Is that a long operation?
If not, we could do a list comprehension

tests/conftest.py
```diff
@@ -596,3 +596,29 @@ def mock_ic03_dataset(tmpdir_factory, mock_image_stream):
...
+        "./mjsynth/images/12_a_34.jpg\n",
+        "./mjsynth/images/12_Jedi_34.jpg\n",
+        "./mjsynth/images/12_!_34.jpg\n",
+    ]
+
+    with open(label_file, "w") as f:
+        for label in labels:
+            f.write(label)
+
+    file = BytesIO(mock_image_stream)
```
fg-mindee, line 619: correct me if I'm wrong, but this variable isn't being used :thinking:

tests/pytorch/test_datasets_pt.py
```diff
@@ -434,3 +434,24 @@ def test_ic03(input_size, num_samples, rotate, mock_ic03_dataset):
...
+    assert repr(ds) == f"MJSynth(train={True})"
+    assert isinstance(image, torch.Tensor)
+    assert image.shape[-2:] == input_size
+    assert image.dtype == torch.float32
+    assert isinstance(target['labels'], str)
+
+    loader = DataLoader(ds, batch_size=2, collate_fn=ds.collate_fn)
+    images, labels = next(iter(loader))
+    assert isinstance(images, torch.Tensor) and images.shape == (2, 3, *input_size)
+    assert isinstance(labels, list) and all(isinstance(elt['labels'], str) for elt in labels)
```
fg-mindee, line 457: any way to use `_validate_dataset` instead? to keep a consistent testing structure

tests/tensorflow/test_datasets_tf.py
```diff
@@ -422,3 +422,25 @@ def test_ic03(input_size, num_samples, rotate, mock_ic03_dataset):
...
+
+    assert isinstance(image, tf.Tensor)
+    assert image.shape[:2] == input_size
+    assert image.dtype == tf.float32
+    assert isinstance(target['labels'], str)
+
+    loader = DataLoader(ds, batch_size=2)
+    images, labels = next(iter(loader))
+    assert isinstance(images, tf.Tensor) and images.shape == (2, *input_size, 3)
+    assert isinstance(labels, list) and all(isinstance(elt['labels'], str) for elt in labels)
```
fg-mindee, line 446: same question here

fg-mindee (commented):
Thanks Felix!

General comments on this:
- I think we should discuss integrating new datasets. There are already quite a lot available through docTR, and we should assess whether it's worth adding one first
- about long downloads, that's an important point (major inconvenience)
- the dataset is synthetic, so we should check whether it's worth integrating considering there is already one that can generate images on-the-fly
- with the number of datasets, we need to find a good way to document / structure datasets in terms of target task: is this an image with OCR label? text recognition ? text detection? etc

doctr/datasets/mjsynth.py
```diff
@@ -0,0 +1,70 @@
...
+class MJSynth(AbstractDataset):
+    """MJSynth dataset from `"Synthetic Data and Artificial Neural Networks for Natural Scene Text Recognition"
+    <https://www.robots.ox.ac.uk/~vgg/data/text/>`_.
+
+    Example::
+        >>> # NOTE: This is a pure recognition dataset without bounding box labels.
+        >>> # NOTE: You need to download the dataset.
+        >>> from doctr.datasets import MJSynth
+        >>> train_set = MJSynth(img_folder="/path/to/mjsynth/mnt/ramdisk/max/90kDICT32px",
+        >>>                  label_folder="/path/to/mjsynth/mnt/ramdisk/max/90kDICT32px/imlist.txt",
```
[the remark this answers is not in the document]
felixdittrich92: :+1:

doctr/datasets/mjsynth.py
felixdittrich92: :+1:

tests/conftest.py
felixdittrich92: used in line 623 to write the image file

tests/pytorch/test_datasets_pt.py
felixdittrich92: :+1:

felixdittrich92:
@fg-mindee 
Any chance to integrate this before we (discuss / refactor) the datasets into ocr/detection/recognition splits/tasks ? 

How i have wrote you the idea behind this is more for research purpose most model benchmarks (recognition) are created on Mjsynth + SynthText merge for train and svt, ic03, ic13, ... for eval .. so if we want to do something like this we need it :sweat_smile: 

And i agree the refactoring into task splits or better to say task specific totally :+1:

felixdittrich92:
Depends on: #891  so for the moment revert to draft :smiley:

charlesmindee (commented):
Thanks!

charlesmindee (dismissed):

felixdittrich92:
@charlesmindee ready to merge :+1:

charlesmindee (approved):
Thanks!

Issue #789, opened 2022-01-10
[models] Ensure all PyTorch models are ONNX exportable


fg-mindee:
Most users of the library are more interested in existing pretrained models to use for inference rather than training. For this reason, it's important to ensure we can easily export those trained models.

- General
    - [x] Add a unittest/CI job to check that all models can be exported with ONNX (#976, #979, #980 )

- Classification
     - [x] VGG (#830)
    - [x] ResNet (#830)
    - [x] MobileNet V3 (#830)
    - [x] MAGC ResNet (#830)

- Text Detection
    - [x] LinkNet #979 
    - [x] DBNet #979 

- Text Recognition
    - [x] CRNN #980 
    - [x] SAR #980 
    - [x] MASTER #980  (#986)
    
    NOTE: master is exportable but in fact of an onnxruntime internal issue (https://github.com/microsoft/onnxruntime/issues/10994) the loading fails

felixdittrich92:
@charlesmindee  classification complete :+1:

frytoli:
I'm relatively green when it comes to ML engineering, but I've been attempting to look at exporting the DBNet models to ONNX. I think we're relatively close already and just need to convert inputs/outputs to a data structure compatible with ONNX (specifically not numpy.ndarray). Here's some sample code and the resulting error. I'm slowly working on solving this on my end, but I wanted to post it here in case this is a really easy problem to solve for someone else more familiar with the project.

```
import torch
from doctr.models.detection import db_resnet50_rotation
from doctr.models.utils import export_classification_model_to_onnx

model = db_resnet50_rotation(pretrained=True)
torch.onnx.export(model, torch.randn(1, 3, 1024, 1024), "model.onnx", input_names=['input'], output_names=['output'], opset_version=13)
```

```
RuntimeError: Only tuples, lists and Variables are supported as JIT inputs/outputs. Dictionaries and strings are also accepted, but their usage is not recommended. Here, received an input of unsupported type: numpy.ndarray
```

felixdittrich92:
Hi @frytoli :wave:,

thanks for working on this :hugs: 

To clearify where the problem is: 
If you run the following with `verbose=True` you get a more detailed output:

```
import torch
from doctr.models.detection import db_resnet50_rotation

model = db_resnet50_rotation(pretrained=True)
dummy_input = torch.rand((1, 3, 1024, 1024), dtype=torch.float32)
torch.onnx.export(
        model,
        dummy_input,
        "test.onnx",
[...]

PR #830, opened 2022-02-22
[onnx] classification models export 

felixdittrich92:
Counterpart of #789 for all classifiaction models

- [x] onnx export integrated in classification training script
- [x] classification models onnx export test

Issue:
#789 


felixdittrich92:
@fg-mindee 
Any update ? Or keep it for review after 0.5.1 release ? 😄

references/classification/train_pytorch.py
```diff
@@ -28,6 +28,27 @@
...
+        dynamic_axes={'input': {0: 'batch_size'}, 'logits': {0: 'batch_size'}},
+        export_params=True, opset_version=13, verbose=False
+    )
+    print(f"Model exported to {name}.onnx")
+    try:
+        from onnxruntime.quantization import quantize_qat
+        quantize_qat(f"{name}.onnx", f'{name}.quant.onnx')
+        print(f"Quantized model saved to {name}.quant.onnx")
+    except ImportError:
+        print("ONNX Runtime is not installed. Skipping quantization.")
```
charlesmindee, line 49: Wince you put onnx in the requirements, I think we don't need the try except and we can simply import it at the beginning of the file

tests/pytorch/test_models_classification_pt.py
```diff
@@ -89,3 +94,45 @@ def test_crop_orientation_model(mock_text_box):
...
+                export_params=True, opset_version=13, verbose=False
+            )
+            # Inference
+            quantize_qat(os.path.join(tmpdir, "model.onnx"), os.path.join(tmpdir, "model.quant.onnx"))
+            ort_session = onnxruntime.InferenceSession(os.path.join(tmpdir, "model.quant.onnx"))
+            ort_outs = ort_session.run(['logits'], {'input': dummy_input.numpy()})
+            assert isinstance(ort_outs, list) and len(ort_outs) == 1
+            assert ort_outs[0].shape == (batch_size, *output_size)
+    except RuntimeError:
+        assert False
```
charlesmindee, line 138: Why do you put a try except in this test ?

charlesmindee (commented):
Thanks for this PR, I think the right place for the export_to_onnx() function might be in the package, in the models/utils/export.py file or something like that, what do you think ?

tests/pytorch/test_models_classification_pt.py
felixdittrich92: Good catch  it's removed :+1:

felixdittrich92:
@charlesmindee ok i have moved it to utils/pytorch ftm :+1:  and also removed both deps we can use torch.onnx and the quantization maybe later add directtly as one line in the training scripts wdyt ?

setup.py
```diff
@@ -65,6 +65,7 @@
     "pytest>=5.3.2",
     "coverage>=4.5.4",
     "hdf5storage>=0.1.18",
+    "onnxruntime>=1.11.0",
```
charlesmindee, line 68: Since we removed the onnxruntime dep from the requirements, we may want to remove from the setup file as well

setup.py
```diff
@@ -137,6 +138,7 @@ def deps_list(*pkgs):
     "coverage",
     "requests",
     "hdf5storage",
+    "onnxruntime",
```
charlesmindee, line 141: same here

charlesmindee (approved):
Thanks for this feature !

Issue #802, opened 2022-01-14
Cannot train pytorch sar_resnet31 and master recognition model


alexcailletMonk:
### Bug description

We tried to train sar_resnet31 and master with `references/recognition/train_pytorch.py`. In both cases, it doesn't train. Train loss, train val are at 0 and exact and partial match too.
Training works with VGG.

### Code snippet to reproduce the bug

```python train_pytorch.py train_data path val_data_path sar_resnet31 --wb -b 64```
```python train_pytorch.py train_data path val_data_path master --wb -b 64```

### Error traceback

no error

### Environment

DocTR version: 0.5.1a0
TensorFlow version: 2.6.2
PyTorch version: 1.9.0+cu111 (torchvision 0.10.0+cu111)
OpenCV version: 4.5.4-dev
OS: Debian GNU/Linux 10 (buster)
Python version: 3.8.12
Is CUDA available (TensorFlow): Yes
Is CUDA available (PyTorch): Yes
CUDA runtime version: 11.4.120
GPU models and configuration: GPU 0: Tesla T4
Nvidia driver version: 470.57.02
cuDNN version: Could not collect

- [x] #920 
- [x] #925
- [x] #941 
- [x] #949

fg-mindee:
Thanks @alexcailletMonk!

I was just able to reproduce the error, I'll investigate! We haven't spent much time on SAR & MASTER lately because of their latency compared to other architectures, but we want to catch up with the others :ok_hand:

fg-mindee:
Hi @alexcailletMonk :wave: 

Quick update on this with ongoing & incoming PRs:
- [ ] #831 
- [ ] Fixing MASTER implementation in PyTorch

felixdittrich92:
@frgfm @charlesmindee After some debugging i have MASTER fixed (with some look on https://github.com/wenwenyu/MASTER-pytorch) so i will cleanup a bit and open a PR that we can have a look at improvements :+1:

felixdittrich92:
- [ ] fix MASTER implementation (PyTorch and Tensorflow)

- [ ] fix SAR implementation (PyTorch and Tensorflow)

- [ ] add pretrained checkpoints for both models in both frameworks

felixdittrich92:
- [ ] fix MASTER implementation (PyTorch and Tensorflow)

- [ ] fix SAR implementation (PyTorch and Tensorflow)

- [ ] add pretrained checkpoints for both models in both frameworks

[...]

PR #831, opened 2022-02-23
fix: Fixed SAR model for training and inference in PyTorch

fg-mindee:
Following up on #802, this PR introduces the following changes:
- fixes a major typo in the loss computation (the sequence masking was inverted)
- cleaned encoder and decoder for a much more understandable version

I tried training the model with the synthetic training set, and a real-world validation set and managed to reach almost 100% of exact match after a few epochs.

Any feedback is welcome!

### DIFF
--- a/doctr/models/recognition/sar/pytorch.py
+++ b/doctr/models/recognition/sar/pytorch.py
@@ -110,7 +110,7 @@ class SARDecoder(nn.Module):
         self,
         features: torch.Tensor,  # (N, C, H, W)
         holistic: torch.Tensor,  # (N, C)
-        gt: Optional[torch.Tensor] = None, # (N, L)
+        gt: Optional[torch.Tensor] = None,  # (N, L)
     ) -> torch.Tensor:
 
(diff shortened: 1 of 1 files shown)


felixdittrich92:
@fg-mindee 
Offtopic:
I have faiced the same in my tests (~99% exact) but after loading this trained model in the whole OCR pipeline there was only wrong results 😅 Do you tested this also ? 😄  (train size was in my test cases 1M / val ~ 200k)

charlesmindee (approved):
Thanks!

fg-mindee:
> @fg-mindee Offtopic: I have faiced the same in my tests (~99% exact) but after loading this trained model in the whole OCR pipeline there was only wrong results sweat_smile Do you tested this also ? smile (train size was in my test cases 1M / val ~ 200k)

I had the same issue until I changed the transformation pipeline during training. On my end, I used a real-world validation dataset, and it worked! If it doesn't on the full OCR pipeline, the problem comes from something else then :)

PR #835, opened 2022-02-25
feat: add aspect ratio for ocr predictor

charlesmindee:
This PR adds the option to preserve aspect ratio of the pages in the whole OCRPredictor pipeline, which leads to much better results on the detection task. 
Any feedback is welcome!

### DIFF
--- a/doctr/models/predictor/base.py
+++ b/doctr/models/predictor/base.py
@@ -110,15 +110,27 @@ class _OCRPredictor:
                 if h > w:
                     # y unchanged, dilate x coord
                     if self.symmetric_pad:
-                        loc_pred[:, :, 0] = np.clip((loc_pred[:, :, 0] - .5) * h / w + .5, 0, 1)
+                        if self.assume_straight_pages:
+                            loc_pred[:, [0, 2]] = np.clip((loc_pred[:, [0, 2]] - .5) * h / w + .5, 0, 1)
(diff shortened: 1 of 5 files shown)


doctr/models/predictor/base.py
```diff
@@ -34,12 +34,16 @@ def __init__(
         self,
         assume_straight_pages: bool = True,
         straighten_pages: bool = False,
+        preserve_aspect_ratio: bool = True,
+        symmetric_pad: bool = True,
```
fg-mindee, line 38: The docstring needs to be updated :)

fg-mindee (changes requested):
Thanks for the PR!
I added a few comments, could you confirm whether this improves perf with the existing checkpoints? (it isn't specified in the PR details, so I prefer to make sure we use everything)

If that improves for existing checkpoints (trained with images that were stretched, that should strongly indicate that we should update our resizing strategy for detection training asap :))

charlesmindee:
Yes @fg-mindee it doesn't hurt performances for now, and I will retrain all detection models preserving the aspect ratio to have a coherent pipeline, because it is way better not to deform writtings on documents for the model convergence!

fg-mindee (dismissed):
Nice!

fg-mindee (approved):
Thanks!

PR #836, opened 2022-02-25
feat: Added loading method for PyTorch artefact detection models from HF Hub

fg-mindee:
Following up on #426, this PR introduces the following modifications:
- refactored faster rcnn
- added a `from_hub` function to load artefact detection architectures in PyTorch
- added unittest
- fixed header unittest

With PyTorch backend, the following snippets are completely equivalent:
```python
# Pretrained
from doctr.models.obj_detection import fasterrcnn_mobilenet_v3_large_fpn
model = fasterrcnn_mobilenet_v3_large_fpn(pretrained=True)
```

```python
# HF Hub
from doctr.models.obj_detection.factory import from_hub
model = from_hub("mindee/fasterrcnn_mobilenet_v3_large_fpn")
```

Any feedback is welcome!

### DIFF
--- a/doctr/models/obj_detection/faster_rcnn/pytorch.py
+++ b/doctr/models/obj_detection/faster_rcnn/pytorch.py
@@ -17,9 +17,7 @@ default_cfgs: Dict[str, Dict[str, Any]] = {
         'input_shape': (3, 1024, 1024),
         'mean': (0.485, 0.456, 0.406),
         'std': (0.229, 0.224, 0.225),
-        'anchor_sizes': [32, 64, 128, 256, 512],
-        'anchor_aspect_ratios': (0.5, 1., 2.),
-        'num_classes': 5,
+        'classes': ["background", "qr_code", "bar_code", "logo", "photo"],
         'url': 'https://github.com/mindee/doctr/releases/download/v0.4.1/fasterrcnn_mobilenet_v3_large_fpn-d5b2490d.pt',
     },
 }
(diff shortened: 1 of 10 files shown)


doctr/models/obj_detection/faster_rcnn/pytorch.py
```diff
@@ -31,11 +29,11 @@ def _fasterrcnn(arch: str, pretrained: bool, **kwargs: Any) -> FasterRCNN:
         "image_mean": default_cfgs[arch]['mean'],
         "image_std": default_cfgs[arch]['std'],
         "box_detections_per_img": 150,
-        "box_score_thresh": 0.15,
+        "box_score_thresh": 0.5,
```
charlesmindee, line 32: Are you positive this change doesn't decrease performances ?

charlesmindee (commented):
Thanks for this amazing feature! just a small comment because you modified a parameter in the refacto

doctr/models/obj_detection/faster_rcnn/pytorch.py
fg-mindee: I experimented a bit with the model as is, and this certainly decreases the recall, but the bare predictions got me concerned on the precision side. @SiddhantBahuguna suggested having a class-specific threshold strategy and that might be the best thing to do, but for now, perhaps it might be better to maximize our precision.

Happy to revert this if you think it's best to favor recall @charlesmindee :ok_hand:

charlesmindee: Let's put it to 0.5 for now, but we should implement quickly a threshold by class if it leads to better results

charlesmindee (approved):
Thanks, @SiddhantBahuguna could you check this threshold by class when you have the time to do so ?

felixdittrich92:
@fg-mindee Can you also add a section into the docs for this feature ? 🤗 Maybe where we can add later links to some models on the hub (this would also fix the different vocab pretrained support i think)

fg-mindee:
> @fg-mindee Can you also add a section into the docs for this feature ? hugs Maybe where we can add later links to some models on the hub (this would also fix the different vocab pretrained support i think)

Hey Felix :wave: 

This is something we have to address but the documentation is built using the TF backend for now (since most high-level feature are identical) and this PR is only about PyTorch (we have no implementation of faster rcnn in TF in docTR for now) :sweat_smile:

charlesmindee (approved):
Thanks!

SiddhantBahuguna:
> Thanks, @SiddhantBahuguna could you check this threshold by class when you have the time to do so ?

Hi @fg-mindee @charlesmindee, actually, since we are using the bare output of the model it lacks post nms strategy. 
After observing the output, along with other filter methods I did use thresholds for logos specifically and it did lead to a better recall. I will open a PR soon for the same :)

PR #842, opened 2022-03-07
fix: Fixed expand_line for horizontal & vertical cases

fg-mindee:
- fixed expand_line for vertical & horizontal lines
- adds clipping when using float in both frameworks
- added unittests for expand_line

_This fixes the randomly failing test we sometimes have on PRs_

Any feedback is welcome!


charlesmindee (approved):
Thanks for this fix!

PR #845, opened 2022-03-08
Switch to new pypdfium2 API

mara004:
Modifications to use the new pypdfium2 API (see https://github.com/mindee/doctr/discussions/844).
I did not dare to try `render_pdf_tobytes()`, as I think that might change the colour format from `RGB` to `BGR`.


doctr/io/pdf.py
```diff
@@ -24,15 +25,17 @@ def read_pdf(file: AbstractFile, scale: float = 2, **kwargs: Any) -> List[np.nda
...
+    if not isinstance(file, (str, bytes)):
         raise TypeError("unsupported object type for argument 'file'")
 
-    if isinstance(file, (str, Path)) and not Path(file).is_file():
+    if isinstance(file, str) and not os.path.isfile(file):
         raise FileNotFoundError(f"unable to access {file}")
 
-    # Read pages with fitz and convert them to numpy ndarrays
-    return [np.asarray(img) for img, _ in pdfium.render_pdf(file, scale=scale)]
+    return [np.asarray(img) for img, _num in pdfium.render_pdf_topil(file, scale=scale, **kwargs)]
```
fg-mindee, line 41: While the 1.0.0 release is in beta, perhaps we could constrain the version specifier and keeping the old behavior?

Additionally, `_num` isn't being used, would you mind reverting back to `_` (it's for the lint checker)

fg-mindee (commented):
Thanks a lot for the PR!

It's much appreciated that upstream projects help with the transition from one version of their project to the next :pray: I added a few comments, let me know what you think!

doctr/io/pdf.py
```diff
@@ -24,15 +25,17 @@ def read_pdf(file: AbstractFile, scale: float = 2, **kwargs: Any) -> List[np.nda
...
+    if isinstance(file, Path):
+        file = str(file)
+    if not isinstance(file, (str, bytes)):
         raise TypeError("unsupported object type for argument 'file'")
 
-    if isinstance(file, (str, Path)) and not Path(file).is_file():
+    if isinstance(file, str) and not os.path.isfile(file):
         raise FileNotFoundError(f"unable to access {file}")
 
-    # Read pages with fitz and convert them to numpy ndarrays
```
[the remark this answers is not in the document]
mara004: You're right, I'll change that

doctr/io/pdf.py
mara004: Okay, will change it too.
(I personally tend not to use `_` for unused variables since it can conflict with the `gettext.gettext()`, which is commonly aliased to an underscore, but you're right that's not relevant here.)

mara004:
I have now released pypdfium2 1.0.0 and applied the review suggestions. Could you please re-run CI?

fg-mindee:
> I have now released pypdfium2 1.0.0 and applied the review suggestions. Could you please re-run CI?

Thanks for the ping :pray: 

I'm all for loose version specifiers as long as they don't put too much constraints to avoid pushing to many env constraints on the user. So here are a few questions:
- are there dependency changes with this new release?
- apart from the import/call of the function to render the page, does the behaviour change in any noticeable way in release `1.0.0`?

If it's only a change of import location, perhaps we could have a dynamic import depending on the version of pypfdium2. But if there are no modifications on version specifiers with this new release, then it's not worth going through this trouble and we can move forward with the PR as is :)

mara004:
> are there dependency changes with this new release?

The dependency on PIL/Pillow has been made optional. This means the core library now does not depend on external components anymore. Otherwise, there are no dependency changes.

> If it's only a change of import location, perhaps we could have a dynamic import depending on the version of pypfdium2.

It essentially is a renaming from `render_pdf()` to `render_pdf_topil()`. Theoretically, I believe it would be possible to use `hasattr()` or check the version specifier to make the code work with older versions of pypdfium2. Probably somewhat like this:
```python3
import pypdfium2 as pdfium

if pdfium._version.V_MAJOR == 0:
    render_pdf_func = pdfium.render_pdf
else:
    render_pdf_func = pdfium.render_pdf_topil

# ...
return [np.asarray(img) for img, _ in render_pdf_func(file, scale=scale, **kwargs)]
```
However, I don't believe it's worth the additional code, considering that pypdfium2 < 1 has not been widely used.

fg-mindee (approved):
Thanks a lot for the PR :pray:  Indeed with the small dependency specifier modifications, it's not worth going through the trouble of a dynamic import!

mara004:
Thanks!

PR #848, opened 2022-03-10
fix: Fixes inplace target modifications for AbstractDatasets

### LINKED ISSUE #840: [datasets] Targets are modified inplace
### Bug description

**Targets** are being changed when iterating over some dataset more than one time.
The reason is storing targets in self.data, and changing them in the `__getitem__` ***in place*** using `pre_transforms`, etc.
```python
# _AbstractDataset
def __getitem__(
        self,
        index: int
    ) -> Tuple[Any, Any]:

        # Read image
        img, target = self._read_sample(index)
        # Pre-transforms (format conversion at run-time etc.)
        if self._pre_transforms is not None:
            img, target = self._pre_transforms(img, target)

        if self.img_transforms is not None:
            # typing issue cf. https://github.com/python/mypy/issues/5485
            img = self.img_transforms(img)  # type: ignore[call-arg]

            img, target = self.sample_transforms(img, target)

        return img, target
```

This can be fixed by copying target in the `_read_sample` 
```python
# AbstractDataset
def _read_sample(self, index: int) -> Tuple[tf.Tensor, Any]:
        img_name, target = self.data[index]
        # Read image
        img = read_img_as_tensor(os.path.join(self.root, img_name), dtype=tf.float32)

        return img, target
```

**OR** returning a copy of the target in all transform methods.
```python
def convert_target_to_relative(img: ImageTensor, target: Dict[str, Any]) -> Tuple[ImageTensor, Dict[str, Any]]:

    target['boxes'] = convert_to_relative_coords(target['boxes'], get_img_shape(img))
    return img, target
```

```python
def process_image(train_example):
  img, target = train_example
  img_numpy = img.numpy() * 255
  for example in target['boxes']:
      print(example)
      unnormalized_example = [int(example[0]*img.shape[1]), int(example[1]*img.shape[0]),
                              int(example[2]*img.shape[1]), int(example[3]*img.shape[0])]
      cv2.rectangle(img=img_numpy,
                    pt1=(unnormalized_example[0], unnormalized_example[1]),
                    pt2=(unnormalized_example[2], unnormalized_example[3]),
                    color=(0, 0, 255), thickness=2)
  return img_numpy  


train_set = SROIE(train=True, download=True)

for i in range(2):
  for j, example in enumerate(train_set):
    if j == 0:    
      print(f"{i} ____")
      img_n = process_image(example)
```

P.S. Sorry for not a pretty code style. This snippet is just for an example :) 

### Error traceback

~changed target box coordinates

### Environment

.

fg-mindee:
- updates `_read_sample` to return a deepcopy of the target to prevent inplace modifications
- adds a dedicated unittest (accessing several times a specific target will return the same value each time now, whether there are inplace transforms or not)

Closes #840

Any feedback is welcome!

### DIFF
--- a/doctr/datasets/datasets/pytorch.py
+++ b/doctr/datasets/datasets/pytorch.py
@@ -4,6 +4,7 @@
 # See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0.txt> for full license details.
 
 import os
+from copy import deepcopy
 from typing import Any, List, Tuple
 
 import torch
@@ -22,7 +23,7 @@ class AbstractDataset(_AbstractDataset):
         # Read image
         img = read_img_as_tensor(os.path.join(self.root, img_name), dtype=torch.float32)
 
-        return img, target
+        return img, deepcopy(target)
 
     @staticmethod
     def collate_fn(samples: List[Tuple[torch.Tensor, Any]]) -> Tuple[torch.Tensor, List[Any]]:
--- a/doctr/datasets/datasets/tensorflow.py
+++ b/doctr/datasets/datasets/tensorflow.py
@@ -4,6 +4,7 @@
 # See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0.txt> for full license details.
 
 import os
+from copy import deepcopy
 from typing import Any, List, Tuple
 
 import tensorflow as tf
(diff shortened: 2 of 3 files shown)


charlesmindee (approved):
Thanks for spotting this !

PR #853, opened 2022-03-13
Info

felixdittrich92:
console print for version and which framework is used


fg-mindee:
Hi @felixdittrich92 :) 

Thanks for the PR! I'm having mixed feelings about this feature:
- perhaps it's best that the output of the environment collection script doesn't change according to env variables :thinking: 
- using `doctr.file_utils`, users can easily know which backend they are running

What do you think?

felixdittrich92:
@fg-mindee 
So normally you don't have both frameworks running in one environment. 😅 
But if that should be the case, I would generally be in favor of giving the user information about which one is being used (the default on tf in particular is difficult for a user to understand without looking at the code). 
However, this is not that easy to implement, since we cannot simply write something to the console or overwrite a possible root logger. 
If you have another idea I would be happy 🤗 
Another thing we could do is document this better ! For example use from doctr.file_utils -> is_xyz_available or setup a basic logger to log informations from doctr

fg-mindee:
Considering that the env collection script may be run in another window with different env var, I think it's not useful to display it there. The script shows whether TF or PyTorch can be used, which usually gives us all the necessary information.

> Another thing we could do is document this better ! For example use from doctr.file_utils -> is_xyz_available or setup a basic logger to log informations from doctr

That might be a good idea actually, to log the enabled backend. For now, without any env var set, file_utils with return True for both PyTorch and TF if both libraries are installed. The priority is only given in `__init__` (cf. https://github.com/mindee/doctr/blob/main/doctr/datasets/generator/__init__.py) but this might complicate things a little.

Perhaps the best option would be to edit the bug report template? (ask user to specify which backend they are running, and how to check that)

felixdittrich92:
@fg-mindee 

ftm i agree that this would be the easiest way i have totally forgot this templates :sweat_smile: :+1:

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -57,6 +57,19 @@ body:
       ```
   validations:
     required: true
+- type: textarea
+  attributes:
+    label: Active backend
```
fg-mindee, line 62: Deep Learning backend

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -57,6 +57,19 @@ body:
...
+    description: |
+      Please run the following snippet inside your code and paste the output below.
```
fg-mindee, line 64: you can remove "inside your code" :)

fg-mindee (commented):
Thanks! Only a few things left to change

PR #851, opened 2022-03-10
[docs] Rework documentation [new style / features / structure]

felixdittrich92:
This PR updates the current documentation 

- [x] new more modern style (furo)
- [x] dark & light theme / framework split docu
- [x] restructure documentation
- move code of cunduct & contributing to docu
- [x] add sample image grid for each dataset

motivation:
Every active modern library also needs decent documentation, which is the showpiece. Therefore we should improve it (even if it's not fun :sweat_smile:) and try to cover everything as best as possible, including the framework difference

further todo: (maybe in another PR)
- same image additions for transformations 
- read the docs grammar stuff and so on

issues:
#729 could be closed with this

Any feedback is very welcome :hugs: 
@fg-mindee @charlesmindee @SiddhantBahuguna  wdyt ? :smile: 

Examples:
light
dark
light2
contrib

I have also checked how it looks as docstring while coding with the tabs and personally i found it pretty nice to have both snippets at once. The only drawback is the much larger docstring.

PR #857, opened 2022-03-16
docs: Adds illustration in the docstrings of doctr.datasets

felixdittrich92:
#851


fg-mindee (approved):
Thanks! It builds & renders well on the previous theme :+1:

PR #861, opened 2022-03-16
Doc models #5

felixdittrich92:
#851


fg-mindee (commented):
Thanks for the PR!

I think there are some mistakes here: in files that are framework-specific, we cannot add tabs.
Two reasons behind this:
- the documentation is currently only built using one framework (TF by default)
- if we have one tab about pytorch, even if the args are called the same way, the typing is not the same for PyTorch or TF. So this might bring some confusion

For now, I think we need to add tabs on features that have framework-inpendent interfaces (i.e. not in a `.../pytorch.py` or `.../tensorflow.py` file). And we'll have to think about the rest. Most likely on the long term:
- we will have to build it for both framework and move generated files somehow (but let's keep that for a later discussion)

felixdittrich92:
@fg-mindee 
I'm really not excited about these requested changes as we lose a good overview of framework independence in the documentation. :sweat:  
On the other hand, I understand your point of view. 
I will change it but we should urgently add the bridging to the 0.6.0 release. (maybe a good point in addition to the export issues - where i think this brings trouble) :sweat_smile:

fg-mindee (approved):
Thanks a lot :pray:

PR #865, opened 2022-03-21
fix: linknet hyperparameters postprocessing + demo for rotation model

charlesmindee:
This PR adjusts the Linknet parameters in the postprocessing to have better performances, and includes the `linknet_resnet18_rotation` model in the streamlit demo. Pages are rendered only for straight models (for now), and we don't write the words above each boxes if the model is a rotation model.
This demo will be improved later, but for the moment it allows anyone to quickly see the performances of our rotation model without modifying the behavior for all other models.

Any feedback is welcome!

This is how it renders for the demo:
image

### DIFF
--- a/demo/app.py
+++ b/demo/app.py
@@ -21,7 +21,7 @@ from doctr.io import DocumentFile
 from doctr.models import ocr_predictor
 from doctr.utils.visualization import visualize_page
 
-DET_ARCHS = ["db_resnet50", "db_mobilenet_v3_large"]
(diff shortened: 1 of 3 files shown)


fg-mindee:
Thanks @charlesmindee :pray: 
What about the page synthesis? Is it still working?

fg-mindee (approved):
Thanks for the PR! I think there are a few things that might need tweaking for the next release, but this can be done in a later PR :+1:

Issue #855, opened 2022-03-15
[datasets/benchmark] Improve research use


felixdittrich92:
### 🚀 The feature

This request can be split into three parts:

- [x] ensure that any already integrated dataset which has the information (boxes & text labels) to be used for the recognition task can also be used for this (crop boxes with corresponding labels)
- [x] add a section in the documentation (like models) for datasets (also split into detection / recognition)
- [x] integrate a script for benchmarking into references/ or directly into the training script for detection and recognition which follows current papers / common used benchmarking splits

detection:
TODO
train:COCO-Text/ ... ?
val: IC03/IC13/...?

recognition:
train: MJSynth/SynthText
val: SVHN/SVT/IIIT5K/IC03/IC13 (+Funsd/Cord)

### Motivation, pitch

It would be great to get a comparison to other implementations or other OCR applications for research purposes. 
This would make the entire library or its implemented models a little more transparent and easier to compare with others. 
As a final point, I have to add that it's just great to see if an implementation reach better benchmarks as other 😅 

### Additional context

Any feedback or suggestion is very welcome 💯

felixdittrich92:
closed with #933

PR #867, opened 2022-03-23
[datasets][PoC] Enable dataset usage for recognition task

felixdittrich92:
This PR is handled as Proof of Concept for further discussions to enable the ability to use existing datasets also for recognition task ( main goal: benchmarks ).
It's easier to show the idea directly in code instead of opening a discussion.

Things to investigate if the concept should be fine:
- Synthtext cropping (if `use_polygons=True`) is to slow (multiprocessing ? maybe in another PR
reminder: [maybe a good reference](https://github.com/bml1g12/benchmarking_multiprocessing_np_array_comms/blob/main/array_benchmark/queue_benchmarking/benchmarks.py))

@fg-mindee 
No worry we can split this later in parts (maybe geometry, torch, tf) for review if you want :sweat_smile: 

Issue: 
#855  First task of this
A good documentation would be part two 
(ATTENTION: reminder for docs: SROIE & SVT does only provide uppercase labels and does not match the 'case-sensitive' in images)

Any feedback is very welcome :+1: 
@charlesmindee @SiddhantBahuguna @fg-mindee 

- [x] Part1: #889 
- [x] Part2: #891


felixdittrich92:
@fg-mindee 
I saw we have already `extract_crops` `extract_rcrops` in `models/_utils.py ` can we maybe move this to geometry ?

felixdittrich92:
@frgfm
Same here any feedback ? 😄

frgfm:
Hey there 🙂

So I had thought about this a few months back. To make sure we are all on the same page, the goal is to:
- take OCR annotated datasets
- make a text recognition dataset out of it 

Correct?

If so, two major options arise:
- doing this dynamically (at it getitem)
- doing this statically in the constructor


If this is for training, I'd argue the second option is the one for a few reasons :
- latency
- memory consumption (opening an image 100 times as big to crop it, in a number of samples that isn't necessarily the batch size comes with heavy consequences)

So perhaps this could be done in a temporary or cache folder 🤷‍♂️

What do you think?

felixdittrich92:
@frgfm 
Yes you are right that's the goal :)
I have had this ways also in mind so currently i have implemented the 'constructor' option (which works really good) with one problem .... SynthText 😓 it is much to slow and needs a lot of memory (really much ~30gb ram) ...but thanks for your opinon now i can iterate on this and find a way to make it better performing 👍 
I was also on track to think about saving for faster reload but it's no option it would waste a lot of the users space
One last thing for the moment are you fine if i move extract_rcrops / extract_crops to geometry otherwise we will have duplicated stuff !? :)

felixdittrich92:
I think it's mostly done i will split it into 2 PRs for easier reviews :+1:

fharper (changes requested):
Can you fix the issues found by Codacy & CodeFactor please?

felixdittrich92:
@fharper that' s only a PoC PR will be split into 2 PRs for review 👍 
Part 1: #889

PR #871, opened 2022-03-25
[Fix] commit tags

felixdittrich92:
update commit tags from: 
https://github.com/mindee/doctr/tags


charlesmindee (approved):
Thanks for the fix!

frgfm:
Hey @felixdittrich92 :) 
Quick question: is this really a fix? Or a refactor? Because I don't remember that there are constraints on the length of the hash prefix to specify it 🤔

frgfm:
@charlesmindee for the next release note, I suggest adding the proper tags here :) 
Perhaps "type: enhancement" & "ext: docs"

felixdittrich92:
@frgfm 👋 ,
I am not sure after merging the postrelease modifications all the tags have changed which was really confusing 😅 I can not really explain why this happens maybe in fact of removing some sphinx stuff !? So i would say in this case it was more a fix

frgfm:
But what do you mean by "the tag changed" ? :) This PR is simply removing the last letter of every single tag, and I'm trying to understand why this was needed 🙃

felixdittrich92:
Yes all tags in https://github.com/mindee/doctr/tags have changed after the documentation update / or the postrelease merge ( i am not sure i have faiced this issue by checking the documentation after the postrelease)

frgfm:
I mean, the tags haven't changed :)
Take the first tag for instance https://github.com/mindee/doctr/releases/tag/v0.5.1, you see "9d03085" but this is only the first character of the full hash (cf. https://github.com/mindee/doctr/commit/9d03085643da8a94f683216cacb08ed8245c7ec1)

So here, whether you use the first 7 or 8 characters, git is flexible enough to accept it. Initially we had 8 chars, and it has been working up until now :)

felixdittrich92:
I think i got it any last character was set wrong (from my side) so the documentation has not build correctly -> i have used only the visible hash to fix this from tags and it worked :sweat_smile: ok than it was a mistake from me

felixdittrich92:
But i think only to take the visible hash is also save and will avoid some mistakes again wdyt ?

frgfm:
> But i think only to take the visible hash is also save and will avoid some mistakes again wdyt ?

Sure, but if you click on https://github.com/mindee/doctr/commit/9d03085643da8a94f683216cacb08ed8245c7ec1, you can see in the URL the full hash. The number of chars is not exactly relevant (i think github requires a minimum, most likely 7)

felixdittrich92:
Now after clarification i would say it was a refactor and not a fix or better a stupid mistake :sweat: :sweat_smile:

frgfm:
Not a mistake don't worry, I was just trying to understand whether github updated the max number of chars for hash resolution or something :)

You can rename the PR "fix" --> "refactor" to better help for PR sorting/labelling for the next release though 👍

PR #874, opened 2022-03-29
[WIP][PyTorch][references] fix pretrained with different vocabs

felixdittrich92:
This PR fixes the mismatching linear layers at each model if we pass a **different vocab as french** and the `pretrained `flag to the training scripts
- recognition: only crnn because sar and master broken
- classification: all

Todo:
- [ ]  tests
- [ ]  magc_resnet31


charlesmindee (approved):
Thank you for the fix!

felixdittrich92:
@charlesmindee TF needs the same but will do in another PR :) i think in TF its enough to modify the checkpoint loading

PR #883, opened 2022-04-04
[classification] Fix cfgs

felixdittrich92:
This PR fixes the following:

- some classification models have had not `cfg` now they have :sweat_smile: 
- also the updating was not correct

Needs to be pushed before: #881  Otherwise the upload from trained classification models to HF hub wont work


charlesmindee (commented):
Hi @felixdittrich92, thank you for this PR!
Before merging, I just have a question:
I don't understand why you need to add the "classes" entry in the model and in the cfg ?

felixdittrich92:
@charlesmindee 
This has no functional impact but if we save the model and it's corresponding cfg in #881  to the hf hub its really confusing to see the french (default) vocab in the classes but model was trained on maybe germn (108 num_classes). 
So it's just there so that the configuration looks correct and there are no misunderstandings

charlesmindee (approved):
Thanks for that!

frgfm:
@charlesmindee missing PR labels here as well :)

PR #892, opened 2022-04-13
[feature][tf/pt] integrate from_hub for all tasks

felixdittrich92:
This PR: 

- follow up on @fg-mindee `from_hub` for all tasks (TF and PT)
- fixes classification archs for push to hub which was not on track

TODO:

- [x] upload dummy models to hub for all working models (exclude other ftm in tests)
- [x] test for each uploaded model
- [x] test whole workflow (push to hub / load from hub)

Documentation will follow in another PR (maybe same after datasets PR merge which needs also documentation)


charlesmindee (approved):
Thanks for the integration!

PR #896, opened 2022-04-23
[documentation/Fix] add documentation for huggingface feature and fix ocr_predictor model loading

felixdittrich92:
This PR:

- adds documentation how to use the huggingface integration (load / push)
- add tables for community models

- fix for passing a model itself directly to the predictor

Issue:
#426 (can be closed after merge)

- [ ] TODO: rework text in documentation

Review:
Code: @charlesmindee 
Documentation:  @fharper 

Any feedback is welcome :)

PR #897, opened 2022-04-26
fix: bug in loading 'db_resnet50_rotation' model from ROT_ARCHS in zoo.py

jasonantonio10:
fix: bug in loading 'db_resnet50_rotation' model from ROT_ARCHS due to ValueError raised when first checking whether the model is in ARCHS in line 36

### DIFF
--- a/doctr/models/detection/zoo.py
+++ b/doctr/models/detection/zoo.py
@@ -33,10 +33,10 @@ def _predictor(
     **kwargs: Any
 ) -> DetectionPredictor:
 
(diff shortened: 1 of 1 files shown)


felixdittrich92:
@jasonantonio10 Thanks a lot for the PR :+1: 
But this fix is ​​already built into #896  :)

charlesmindee:
Thanks! As @felixdittrich92, this will be fixed when merging #896, so closing this to avoid duplicates/divergences

PR #862, opened 2022-03-16
docs: Updated folder hierarchy of doc source and nootbooks to rst file

felixdittrich92:
#851

PR #901, opened 2022-04-27
docs: Fixed typo in installation instructions

frgfm:
Hey there 👋 

I noticed a small typo introduced in #862 for installation instructions. This PR simply fixes that for both tabs.

Any feedback is welcome!

### DIFF
--- a/docs/source/getting_started/installing.rst
+++ b/docs/source/getting_started/installing.rst
@@ -43,13 +43,13 @@ We strive towards reducing framework-specific dependencies to a minimum, but som
 
         .. code:: bash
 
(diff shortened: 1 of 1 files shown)


felixdittrich92:
Huhu @frgfm 
I have integrated this also in #896 😅

frgfm:
Oh my bad @felixdittrich92 : that being said, it's better to split PRs, and in #896, you are still missing the quotes (for people using zsh or something else than pure bash, it's needed). So perhaps we should remove the typo fix from #896? What do you think?

felixdittrich92:
yes you are right :+1: lets do this :)

PR #902, opened 2022-04-27
fix: Fixed load_pretrained_params in PyTorch when ignoring keys

frgfm:
Following up on #874, this PR introduces the following modifications:
- renamed "pop_entrys" to "ignore_keys"
- updated `load_pretrained_params` to avoid non-strict loading of wrongly sized state_dict
- fixed classification models `ignore_keys` mechanism (the keys to ignore were hardcoded in the factory function, while this function is used to build models of different size, and thus with Linear layers being named differently)
- added dedicated unittest for `ignore_keys`

Any feedback is welcome!

### DIFF
--- a/tests/pytorch/test_models_classification_pt.py
+++ b/tests/pytorch/test_models_classification_pt.py
@@ -46,9 +46,8 @@ def test_classification_architectures(arch_name, input_shape, output_size):
     # Model
     model = classification.__dict__[arch_name](pretrained=True).eval()
     _test_classification(model, input_shape, output_size)
-    # test pretrained model with different num_classes
-    model = classification.__dict__[arch_name](pretrained=True, num_classes=108).eval()
-    _test_classification(model, input_shape, output_size=(108,))
+    # Check that you can pretrained everything up until the last layer
+    classification.__dict__[arch_name](pretrained=True, num_classes=10)
 
 
 @pytest.mark.parametrize(
--- a/doctr/models/classification/magc_resnet/pytorch.py
+++ b/doctr/models/classification/magc_resnet/pytorch.py
@@ -137,7 +137,7 @@ def _magc_resnet(
     if pretrained:
         # The number of classes is not the same as the number of classes in the pretrained model =>
         # remove the last layer weights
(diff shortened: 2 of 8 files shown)


tests/pytorch/test_models_classification_pt.py
```diff
@@ -46,9 +46,8 @@ def test_classification_architectures(arch_name, input_shape, output_size):
     # Model
     model = classification.__dict__[arch_name](pretrained=True).eval()
     _test_classification(model, input_shape, output_size)
-    # test pretrained model with different num_classes
-    model = classification.__dict__[arch_name](pretrained=True, num_classes=108).eval()
-    _test_classification(model, input_shape, output_size=(108,))
+    # Check that you can pretrained everything up until the last layer
```
felixdittrich92, line 49: maybe better: Test whether a pretrained model can be initialized down to the last layer ? :)

felixdittrich92:
@frgfm  Thanks for refactoring this now it is more dynamically ;) 
LGTM only a missing import and maybe the comment ? :)

charlesmindee (commented):
Thanks for the refacto! Only  a small missing import it seems

tests/pytorch/test_models_classification_pt.py
frgfm: Not sure I understand?

Compared to previous features, my idea is that we can already assume:
- models are built correctly
- pretrained with default output classes is working

What needs testing here is that the checkpoint can be loaded on a model with different num_classes. So perhaps we could check that the bias of a given layer is indeed the one of the state_dict, but I don't see much more :) 

(and I try to keep unittests from running slower)

felixdittrich92: @frgfm sorry I just mean to change this comment :
```
# Check that you can pretrained everything up until the last layer
maybe better / clearer:
# Test whether a pretrained model can be initialized down to the last layer 
```
:sweat_smile:

felixdittrich92: @frgfm and about the unitests i think my tests from: https://github.com/mindee/doctr/pull/892/files are currently the biggest slowdown but i think we need to test each model

charlesmindee (approved):
Thanks!

PR #903, opened 2022-04-28
[Fix] imgur5k test

felixdittrich92:
This PR fixes the failing imgur5k test:
- inconsistent number of boxes samples yields to different numbers of samples in the test
- now each sample has the same number of boxes to avoid random numbers of samples (recognition test)


felixdittrich92:
@charlesmindee

charlesmindee (approved):
Thanks for the fix!

PR #904, opened 2022-04-29
[Fix]: Documentation add missing in vocabs and correct tab in sharing models

felixdittrich92:
This PR:
- adds missing entrys in vocabs docu
- fixes tab in sharing models


charlesmindee (approved):
Thanks!

Issue #847, opened 2022-03-10
Doctr api doc error on docker


dvivarelli:
### Bug description

I have cloned the project and build the image with docker compose and tried to open the documentation at http://localhost:8002/redoc but the page remains in loading.

I have checked the network tab on chrome and I saw an error:

GET http://localhost:8002/openapi.json  
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/base.py", line 25, in __call__
    response = await self.dispatch_func(request, self.call_next)
  File "/app/./app/main.py", line 26, in add_process_time_header
    response = await call_next(request)
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/base.py", line 45, in call_next
    task.result()
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/base.py", line 38, in coro
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc from None
  File "/usr/local/lib/python3.8/site-packages/starlette/exceptions.py", line 71, in __call__
    await self.app(scope, receive, sender)
  File "/usr/local/lib/python3.8/site-packages/starlette/routing.py", line 580, in __call__
    await route.handle(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/starlette/routing.py", line 241, in handle
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/starlette/routing.py", line 52, in app
    response = await func(request)
  File "/usr/local/lib/python3.8/site-packages/fastapi/applications.py", line 161, in openapi
    return JSONResponse(self.openapi())
  File "/app/./app/main.py", line 36, in custom_openapi
    openapi_schema = get_openapi(
  File "/usr/local/lib/python3.8/site-packages/fastapi/openapi/utils.py", line 410, in get_openapi
    return jsonable_encoder(OpenAPI(**output), by_alias=True, exclude_none=True)  # type: ignore
  File "pydantic/main.py", line 406, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 4 validation errors for OpenAPI
components -> schemas -> DetectionOut -> properties -> box -> items
  value is not a valid dict (type=type_error.dict)
[...]

PR #906, opened 2022-04-29
ci: Add swagger ping in API CI job

frgfm:
Following up on #847, this PR introduces the following modification:
- adds a step in CI job for API, to ping the swagger before running the unittest

Presumably, if the behaviour described in #847, the CI should fail on this PR.

Any feedback is welcome!


charlesmindee (approved):
Thanks for this!

PR #910, opened 2022-05-05
[onnx/tf] classification models export 

felixdittrich92:
Export for TF classification models

- function for easy exporting (similar to PT) and corresponding tests
- minor renaming
- adds tf2onnx on TF side
- include export ability in classification training script

NOTE: test excludes magc_resnet31 currently there are some problems: 
```
# Name:'res_net_4/magc/transform/conv2d_289/Conv2D:0_nchwc'
# Status Message: Input channels C is not equal to kernel channels * group. C: 32 kernel channels: 256 group: 1
```
All other models works fine.

Issue:
#789 

### DIFF
--- a/doctr/models/utils/pytorch.py
+++ b/doctr/models/utils/pytorch.py
@@ -92,7 +92,7 @@ def conv_sequence_pt(
     return conv_seq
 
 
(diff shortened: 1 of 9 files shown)


felixdittrich92:
For conftest the url does not exist anymore

felixdittrich92:
done

charlesmindee (approved):
Thanks for that!

PR #913, opened 2022-05-10
[api] Fixed API route response

assistechsol:
This PR:
- updates the API output, to return the text for each page

### DIFF
--- a/api/app/routes/ocr.py
+++ b/api/app/routes/ocr.py
@@ -21,4 +21,4 @@ async def perform_ocr(file: UploadFile = File(...)):
     out = predictor([img])
 
     return [OCROut(box=(*word.geometry[0], *word.geometry[1]), value=word.value)
-            for word in out.pages[0].blocks[0].lines[0].words]
+            for index in range(len(out.pages[0].blocks)) for word in out.pages[0].blocks[index].lines[0].words]


api/app/routes/ocr.py
```diff
@@ -21,4 +21,4 @@ async def perform_ocr(file: UploadFile = File(...)):
     out = predictor([img])
 
     return [OCROut(box=(*word.geometry[0], *word.geometry[1]), value=word.value)
-            for word in out.pages[0].blocks[0].lines[0].words]
+            for index in range(len(out.pages[0].blocks)) for word in out.pages[0].blocks[index].lines[0].words]
```
frgfm, line 24: This piece of code is still only checking the first page only and only the first line of each block. The route is single-page only for now, so let's fix this :) I would suggest:
```python
for block in out.pages[0].blocks for line in block.lines for word in line.words
```

frgfm: Would you mind updating the PR please? 🙏

frgfm (changes requested):
Hello @assistechsol 👋 

Thanks a lot for spotting this and opening the PR! I added a slight correction suggestion as comment!

Cheers!

api/app/routes/ocr.py
frgfm: Yes but this route is single page for now!
So this means that pages only have 1 element 👍

frgfm: Hello @SamSamhuns @assistechsol :wave: 

What do you think about my previous comment?
Let's find a solution to move forward with this PR :+1:

SamSamhuns: Sounds good, probably a good idea to mention somewhere its only for one page tho

frgfm: Yes! Mind updating the PR then? :pray:

felixdittrich92:
Hi @assistechsol :wave:, any updates ? :)

felixdittrich92:
closed with: no response

PR #914, opened 2022-05-10
-

felixdittrich92:
This PR:
- removes MAGC & SAR & MASTER and the related parts in documentation/references from main branch

motivation:
- SAR & MASTER implementations are broken on both frameworks and build currently a blocking part to continue with other things so i think we have to take the time to reimplement both maybe the other things are more stable :)

**ATTENTION:** `sar-master-backup` branch is created !

@frgfm @charlesmindee 
Any feedback is welcome ! :)


frgfm:
Before removing all of those, perhaps we should quickly check that magc_resnet is working?
The feature extractor of both models are likely to be working (and thus we could keep those)

felixdittrich92:
@frgfm sure let me take a bit more time before we remove this -> DRAFT :)

felixdittrich92:
Closed take the time and fix it :)

PR #922, opened 2022-05-18
fix: Trying to fix API template

frgfm:
Following up on #847, this PR introduces the following changes:
- updates version pin for fastAPI & uvicorn to debug API template

Any feedback is welcome!

### DIFF
--- a/api/requirements.txt
+++ b/api/requirements.txt
@@ -1,4 +1,4 @@
-fastapi>=0.65.2
-uvicorn>=0.11.1
+fastapi==0.65.2
+uvicorn==0.11.1
 python-multipart==0.0.5
 python-doctr>=0.2.0


frgfm:
Mmmh it looks like there are several dependency issues 🤔

frgfm:
@felixdittrich92 to the best of my knowledge, the dependabot is helpful to make you keep this up to date, but it doesn't take breaking changes into account 🤔 So I doubt it would work

felixdittrich92:
@frgfm mh yes .. btw pinning httptools to 0.1.2 solves the dependency conflict and i think if we run compose as root could solve the docs issue

frgfm:
> @frgfm mh yes .. btw pinning httptools to 0.1.2 solves the dependency conflict and i think if we run compose as root could solve the docs issue

I cannot run the docker correctly with my current setup, but once I get my hands on my new one, I'll be able to get back into docker and fix this 👍  (within a few weeks max)

felixdittrich92:
Hi @frgfm :wave: do you think we can close this ? .)

frgfm:
Sure!

PR #925, opened 2022-05-20
[Fix] TensorFlow SAR_Resnet31 implementation

felixdittrich92:
This PR: (Still in progress :))

- fix SARDecoder forward step

- [x] check train
- [x] check inference
- [x] cleanup / improve code

Issue:
#802

### DIFF
--- a/doctr/models/classification/resnet/tensorflow.py
+++ b/doctr/models/classification/resnet/tensorflow.py
@@ -31,7 +31,7 @@ default_cfgs: Dict[str, Dict[str, Any]] = {
         'std': (0.299, 0.296, 0.301),
         'input_shape': (32, 32, 3),
         'classes': list(VOCABS['french']),
-        'url': 'https://github.com/mindee/doctr/releases/download/v0.5.0/resnet31-5a47a60b.zip',
+        'url': None,
     },
     'resnet34': {
         'mean': (0.694, 0.695, 0.693),
--- a/doctr/models/classification/resnet/tensorflow.py
+++ b/doctr/models/classification/resnet/tensorflow.py
@@ -31,7 +31,7 @@ default_cfgs: Dict[str, Dict[str, Any]] = {
         'std': (0.299, 0.296, 0.301),
         'input_shape': (32, 32, 3),
         'classes': list(VOCABS['french']),
-        'url': None,
+        'url': 'https://github.com/mindee/doctr/releases/download/v0.5.0/resnet31-5a47a60b.zip',
     },
     'resnet34': {
         'mean': (0.694, 0.695, 0.693),
--- a/doctr/models/classification/resnet/tensorflow.py
+++ b/doctr/models/classification/resnet/tensorflow.py
@@ -27,8 +27,8 @@ default_cfgs: Dict[str, Dict[str, Any]] = {
         'url': 'https://github.com/mindee/doctr/releases/download/v0.4.1/resnet18-d4634669.zip',
     },
     'resnet31': {
-        'mean': (0.5, 0.5, 0.5),
-        'std': (1., 1., 1.),
+        'mean': (0.694, 0.695, 0.693),
+        'std': (0.299, 0.296, 0.301),
         'input_shape': (32, 32, 3),
         'classes': list(VOCABS['french']),
         'url': 'https://github.com/mindee/doctr/releases/download/v0.5.0/resnet31-5a47a60b.zip',
--- a/doctr/models/recognition/sar/tensorflow.py
+++ b/doctr/models/recognition/sar/tensorflow.py
@@ -23,10 +23,33 @@ default_cfgs: Dict[str, Dict[str, Any]] = {
         'mean': (0.694, 0.695, 0.693),
         'std': (0.299, 0.296, 0.301),
         'input_shape': (32, 128, 3),
-        'vocab': VOCABS['legacy_french'],
-        'url': 'https://github.com/mindee/doctr/releases/download/v0.3.0/sar_resnet31-9ee49970.zip',
+        'vocab': VOCABS['french'],
+        'url': None,
     },
 }
+
+
+class SAREncoder(layers.Layer, NestedObject):
+    """Implements encoder module of the SAR model
+
+    Args:
+        rnn_units: number of hidden rnn units
+        dropout_prob: dropout probability
+    """
+
+    def __init__(self, rnn_units: int, dropout_prob: float = 0.) -> None:
+
+        super().__init__()
+        self.rnn = Sequential(
+            [
+                layers.LSTM(units=rnn_units, return_sequences=True, recurrent_dropout=dropout_prob),
+                layers.LSTM(units=rnn_units, return_sequences=False, recurrent_dropout=dropout_prob),
+            ]
+        )
+
+    def call(self, x: tf.Tensor) -> tf.Tensor:
+        # (N, C)
+        return self.rnn(x)
 
 
 class AttentionModule(layers.Layer, NestedObject):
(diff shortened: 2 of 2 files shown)


felixdittrich92:
@frgfm 
training: Overall it looks good with same config as used for PT implementation it reaches after 17 epochs an exact match from ~70% 
inference: i have some problems to translate the torch.scatter_ functionality could you help with this ? :)

doctr/models/recognition/sar/tensorflow.py
```diff
@@ -124,39 +146,51 @@ def call(
...
-
-        return outputs
+                if gt is not None:
+                    #(N, embedding_units) -2 because of <bos> and <eos> (same)
+                    prev_symbol = self.embed(gt_embedding[:, t - 2])
+                else:
+                    # -1 to start at timestep where prev_symbol was initialized
+                    index = tf.argmax(logits_list[t - 1], axis=-1)
+                    # (1, embedding_units)
+                    # TODO: translate me :) prev_symbol = prev_symbol.scatter_(1, index.unsqueeze(1), 1)
```
felixdittrich92, line 174: @frgfm 
How can we do this in TF ? 
We need to update the prev_symbol last dimension at the given index with 1

felixdittrich92: ok got it

frgfm:
> @frgfm training: Overall it looks good with same config as used for PT implementation it reaches after 17 epochs an exact match from ~70% inference: i have some problems to translate the torch.scatter_ functionality could you help with this ? :)

Could you check the test perf against FUNSD or CORD ? :) 
cf. bench https://mindee.github.io/doctr/using_doctr/using_models.html#id5

doctr/models/recognition/sar/pytorch.py
```diff
@@ -184,7 +185,7 @@ def __init__(
         vocab: str,
         rnn_units: int = 512,
         embedding_units: int = 512,
-        attention_units: int = 64,
+        attention_units: int = 512,
```
frgfm, line 188: Could you comment on this change?
Is this different from the paper?

doctr/models/recognition/sar/pytorch.py
```diff
@@ -207,8 +208,14 @@ def __init__(
...
         self.encoder = SAREncoder(out_shape[1], rnn_units, dropout_prob)
-        self.decoder = SARDecoder(rnn_units, self.max_length, len(self.vocab),
-                                  embedding_units, attention_units, dropout_prob=dropout_prob)
+        self.decoder = SARDecoder(
+            rnn_units,
+            self.max_length,
+            len(self.vocab),
+            embedding_units,
+            attention_units,
+            dropout_prob=dropout_prob
```
frgfm, line 217: missing a comma here (reason behind this being that if you add a new arg, if the comma is already there the diff will be 1 line, vs if you dont --> 2 lines :) )

doctr/models/recognition/sar/tensorflow.py
```diff
@@ -250,16 +284,17 @@ def call(
         target: Optional[List[str]] = None,
         return_model_output: bool = False,
         return_preds: bool = False,
-        **kwargs: Any,
```
frgfm, line 253: that always TF to pass "training=True/False" down to each module
Is there a reason why we should remove this?

frgfm (commented):
Thanks for the PR Felix 🙏 
I added a few comments!

felixdittrich92:
> > @frgfm training: Overall it looks good with same config as used for PT implementation it reaches after 17 epochs an exact match from ~70% inference: i have some problems to translate the torch.scatter_ functionality could you help with this ? :)
> 
> Could you check the test perf against FUNSD or CORD ? :) cf. bench https://mindee.github.io/doctr/using_doctr/using_models.html#id5

@frgfm i think this would not be comparable without a full dataset (like mindee intern) :/ I test the implementations currently on a difficult toy dataset 10k train and 2.5k val (contains extrem blurred, rotated images, ~ 30 different fonts and some corrupted characters). Maybe if i found a bit more time i could train it on MJSynth

doctr/models/recognition/sar/pytorch.py
felixdittrich92: this 64 units was a mistake from my side i have had in mind 1x1x64 while computing the glimpse but it was 1x1xd and hidden_dim is in the complete model specified with 512 units :)

doctr/models/recognition/sar/tensorflow.py
felixdittrich92: I don't think there's any reason to hold it at this time as you shouldn't get into this opportunity. Also, I'm not sure if it could cause problems when exporting ONNX. And the current implementation has an enormous potential for bugs through the massive use of kwargs down into the backbone. wdyt ? :)

frgfm: I said this because in TF, there is no `training` attribute. You pass `training=True` down to the call method, which is then passed to all submodules. So for instance here, if in one of the submodules, there is a batch norm layer, we're gonna be in big trouble 😅  (because those type of layers behave quite differently between training & test times)

felixdittrich92: ok that's an really good argument :sweat_smile:

felixdittrich92:
@frgfm ok now with kwargs and check :+1: :sweat_smile:

frgfm:
FYI: this PR introduces the deprecation of resnet31 as pretrained model for TF

felixdittrich92:
> FYI: this PR introduces the deprecation of resnet31 as pretrained model for TF

@frgfm 
Yes i know, but (correct me if i'm wrong) it is currently only used as backbone for SAR (and master which will change before next release also)

trained 1 Epoch on MJSynth Subset : 500k train / 200k val
Validation loss decreased inf --> 0.215485: saving state...                                                                                                     
Epoch 1/100 - Validation loss: 0.215485 (Exact: 72.99% | Partial: 77.53%)

FUNSD:
Validation loss: 1.89441 (Exact: 44.94% | Partial: 47.41%)
CORD:
Validation loss: 2.34539 (Exact: 36.30% | Partial: 36.77%)

frgfm (commented):
Thanks Felix, I added some comments!

frgfm:
> > FYI: this PR introduces the deprecation of resnet31 as pretrained model for TF
> 
> @frgfm Yes i know, but (correct me if i'm wrong) it is currently only used as backbone for SAR (and master which will change before next release also) so i would say lets include this for next training iteration
> 
> trained 1 Epoch on MJSynth Subset : 500k train / 200k val Validation loss decreased inf --> 0.215485: saving state... Epoch 1/100 - Validation loss: 0.215485 (Exact: 72.99% | Partial: 77.53%)
> 
> FUNSD: Validation loss: 1.89441 (Exact: 44.94% | Partial: 47.41%) CORD: Validation loss: 2.34539 (Exact: 36.30% | Partial: 36.77%)
> 
> I would say for a toy run this looks really not bad and we can close the SAR issue with this PR wdyt ? :)

I agree, that's quite good 👍

doctr/models/recognition/sar/tensorflow.py
```diff
@@ -99,64 +123,77 @@ def __init__(
...
-        input_shape: Optional[List[Tuple[Optional[int]]]] = None,
+        dropout_prob: float = 0.,
     ) -> None:
 
         super().__init__()
         self.vocab_size = vocab_size
-        self.lstm_decoder = layers.StackedRNNCells(
+        self.max_length = max_length
+
+        self.embed = layers.Dense(embedding_units)
```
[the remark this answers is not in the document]
felixdittrich92: good catch :) no need

doctr/models/recognition/sar/tensorflow.py
```diff
@@ -99,64 +123,77 @@ def __init__(
         embedding_units: int,
         attention_units: int,
         num_decoder_layers: int = 2,
-        input_shape: Optional[List[Tuple[Optional[int]]]] = None,
```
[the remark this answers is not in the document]
felixdittrich92: mh only for the lstm cells but this will be initialized in step 0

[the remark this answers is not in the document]
felixdittrich92: WARNING:tensorflow:Gradients do not exist for variables ['sar/sar_decoder/stacked_rnn_cells/lstm_cell_3/kernel:0', 'sar/sar_decoder/stacked_rnn_cells/lstm_cell_3/recurrent_kernel:0', 'sar/sar_decoder/stacked_rnn_cells/lstm_cell_3/bias:0'] when minimizing the loss. If you're using `model.compile()`, did you forget to provide a `loss`argument?

doctr/models/recognition/sar/tensorflow.py
```diff
@@ -254,12 +290,19 @@ def call(
     ) -> Dict[str, Any]:
 
         features = self.feat_extractor(x, **kwargs)
-        pooled_features = tf.reduce_max(features, axis=1)  # vertical max pooling
-        encoded = self.encoder(pooled_features, **kwargs)
+        # vertical max pooling --> (N, C, W)
+        pooled_features = tf.reduce_max(features, axis=1)
+        # holistic (N, C)
+        encoded = self.encoder(pooled_features)
```
frgfm, line 296: missing the kwargs in the call 🙃

doctr/models/recognition/sar/tensorflow.py
```diff
@@ -254,12 +290,19 @@ def call(
...
+
         if target is not None:
             gt, seq_len = self.build_target(target)
             seq_len = tf.cast(seq_len, tf.int32)
-        decoded_features = self.decoder(features, encoded, gt=None if target is None else gt, **kwargs)
+
+        if kwargs.get('training', False) and gt is None:
+            raise ValueError('Need to provide labels during training for teacher forcing')
+
+        decoded_features = self.decoder(features, encoded, gt=None if target is None else gt)
```
frgfm, line 305: same here

frgfm (changes requested):
Just a few typos left!

doctr/models/classification/resnet/tensorflow.py
```diff
@@ -27,11 +27,11 @@
...
     },
     'resnet31': {
-        'mean': (0.5, 0.5, 0.5),
-        'std': (1., 1., 1.),
+        'mean': (0.694, 0.695, 0.693),
+        'std': (0.299, 0.296, 0.301),
         'input_shape': (32, 32, 3),
         'classes': list(VOCABS['french']),
-        'url': 'https://github.com/mindee/doctr/releases/download/v0.5.0/resnet31-5a47a60b.zip',
+        'url': None,
```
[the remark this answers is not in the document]
felixdittrich92: yes i have removed it after your last comment which sounds like better to remove :sweat_smile:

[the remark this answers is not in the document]
frgfm: It's better, but when we can avoid deprecation from one commit to another, it's much better for the user :) 
The SAR checkpoint won't load with the fixes, but the backbone will

frgfm (approved):
Thanks Felix 🙏

PR #931, opened 2022-05-27
Change datatype in doctr/models/preprocessor/pytorch.py

prajinkhadka:
I faced an issue that  ```samples``` parameter in ```batch_inputs``` is an object but not the list. I reckon it is because of using map. It returns an map object.

### DIFF
--- a/doctr/models/preprocessor/pytorch.py
+++ b/doctr/models/preprocessor/pytorch.py
@@ -55,6 +55,7 @@ class PreProcessor(nn.Module):
         Returns:
             list of batched samples (*, C, H, W)
         """
+        
         num_batches = int(math.ceil(len(list(samples)) / self.batch_size))
         batches = [
             torch.stack(samples[idx * self.batch_size: min((idx + 1) * self.batch_size, len(samples))], dim=0)
--- a/doctr/models/preprocessor/pytorch.py
+++ b/doctr/models/preprocessor/pytorch.py
@@ -55,8 +55,7 @@ class PreProcessor(nn.Module):
         Returns:
             list of batched samples (*, C, H, W)
         """
-
-        num_batches = int(math.ceil(len(samples) / self.batch_size))
+        num_batches = int(math.ceil(len(list(samples)) / self.batch_size))
         batches = [
             torch.stack(samples[idx * self.batch_size: min((idx + 1) * self.batch_size, len(samples))], dim=0)
             for idx in range(int(num_batches))


frgfm:
Hi @prajinkhadka 👋 

This is a bit of a thin justification for this modification 😅 
Could you share a snippet to understand where this becomes a problems? The method you're referring is taking list of tensors as inputs (cf. the typing), so we need to identify in which case this isn't a list.

Best!

prajinkhadka:
Sure ! 
I will test the scenario and post here.

prajinkhadka:
Hi @frgfm 
I ran some test and here are the results. 

The issue is from ```multithread_exec``` function in ```utils/multithreading.py```  

The ```if block``` i.e. if ```threads<2```, there we have used map function which returns a map object not the list of tensors as expected. But the else block which uses map from multiprocessing module returns the tensors as expected. So, running doctr in single threaded CPU is causing  the problem. 

Can be  verified by changing the function as :

```def multithread_exec(func: Callable[[Any], Any], seq: Iterable[Any], threads: Optional[int] = None) -> Iterable[Any]:
    """Execute a given function in parallel for each element of a given sequence

    >>> from doctr.utils.multithreading import multithread_exec
    >>> entries = [1, 4, 8]
    >>> results = multithread_exec(lambda x: x ** 2, entries)

    Args:
        func: function to be executed on each element of the iterable
        seq: iterable
        threads: number of workers to be used for multiprocessing

    Returns:
        iterable of the function's results using the iterable as inputs
    """

    threads = threads if isinstance(threads, int) else min(16, mp.cpu_count())
    # Single-thread
    results = map(func, seq)
    # # Multi-threading
    # else:
    # with ThreadPool(threads) as tp:
    #     results = tp.map(func, seq)  # type: ignore[assignment]
    print(type(results))
    return results
 ```

Thanks

frgfm:
> Hi @frgfm I ran some test and here are the results.
> 
> The issue is from `multithread_exec` function in `utils/multithreading.py`
> 
> The `if block` i.e. if `threads<2`, there we have used map function which returns a map object not the list of tensors as expected. But the else block which uses map from multiprocessing module returns the tensors as expected. So, running doctr in single threaded CPU is causing the problem.

Ah interesting, I see the problem now. That might be worth opening a dedicated issue for this

About the proposed solution, two lines after your modification, we use indexing on `samples`. So if samples is a generator at this point, that will only delay the problem. I'd argue it's better to either:
- cast the output of `multithread_exec` to a list in the `__call__` method
- change this behaviour at the source of the problem in `multithread_exec`

What do you think?

prajinkhadka:
> > Hi @frgfm I ran some test and here are the results.
> > The issue is from `multithread_exec` function in `utils/multithreading.py`
> > The `if block` i.e. if `threads<2`, there we have used map function which returns a map object not the list of tensors as expected. But the else block which uses map from multiprocessing module returns the tensors as expected. So, running doctr in single threaded CPU is causing the problem.
> 
> Ah interesting, I see the problem now. That might be worth opening a dedicated issue for this
> 
> About the proposed solution, two lines after your modification, we use indexing on `samples`. So if samples is a generator at this point, that will only delay the problem. I'd argue it's better to either:
> 
>     * cast the output of `multithread_exec` to a list in the `__call__` method
> 
>     * change this behaviour at the source of the problem in `multithread_exec`
> 
> 
> What do you think?

Yeah, I reckon it would be better to  change this behavior at the source of the problem in multithread_exec by casting the output to the list.

What do you say ?

frgfm:
> Yeah, I reckon it would be better to change this behavior at the source of the problem in multithread_exec by casting the output to the list.
> 
> What do you say ?

So the problem is that this needs to be dynamic (i.e. if the input is a tuple, then this should as well). To make it simple, either:
- we find a way to cast the output type to the same as the input type
- or we always return a generator and the user will have to cast 

:man_shrugging:

prajinkhadka:
> > So the problem is that this needs to be dynamic (i.e. if the input is a tuple, then this should as well). To make it simple, either:
> > 
> > * we find a way to cast the output type to the same as the input type
> > * or we always return a generator and the user will have to cast
> 
> I think we should go for the second options, it's safer & more uniform

yeah, second option seems to be safe.

felixdittrich92:
Hi @prajinkhadka :wave: any updates ?  Do you still want to move forward with this PR ? :)

felixdittrich92:
Closed by: #1019

PR #936, opened 2022-05-31
add a Slack Community badge

fharper:
Feel free to suggest the text, color or position changes!


felixdittrich92:
Thanks @fharper really good idea :)
I would add it at least so it's more in focus @frgfm @charlesmindee wdyt ?

frgfm:
I was thinking about reorganizing the icons as well !
Excellent idea @fharper 👍 

For the sake of visual aspect, I'd suggest to go for a color matching Slack company guidelines while staying similar to other icons from the README. Something like this for instance ?
[Docker](https://join.slack.com/t/mindee-community/shared_invite/zt-uzgmljfl-MotFVfH~IdEZxjp~0zldww)

Also I think there is a mistake in the slack URL you put (it looks like the internal one, not the community), I put the invitation URL in the above button ☝️

fharper:
@frgfm: the link is good, and I prefer the one you created (the flat design) but it won't match with the other one. I guess you'll update them like this?

frgfm (approved):
Oh I hadn't checked that there is a redirect in the URL 👍 
We'll update badges on another PR, that one is a great addition anyway !

PR #937, opened 2022-06-01
Fix links in readme 

jsn5:
Fix links to text detection and text recognition in readme


frgfm (approved):
Good catch, thanks a lot 🙏

frgfm:
@felixdittrich92 better to edit the commit message next time before merging 😉 
Something like "docs: Fixed URLs in README"

PR #941, opened 2022-06-02
[Fix] PyTorch MASTER implementation

felixdittrich92:
This PR:  (Still in progress)

- [x] fix decoding step
- [x] fix Decoder

some notes at the moment:
different from paper input set to our default (32, 128)

ftm: transformer plain implementation not nn modules

but training and inference works fine :sweat_smile: 

@frgfm feel free to grab this and take a look :+1: 

Issue:
#802

### DIFF
--- a/doctr/models/recognition/master/pytorch.py
+++ b/doctr/models/recognition/master/pytorch.py
@@ -74,7 +74,7 @@ class MASTER(_MASTER, nn.Module):
         self.vocab_size = len(vocab)
 
         self.feat_extractor = feature_extractor
-        self.positional_encoding = PositionalEncoding(self.d_model, dropout)
+        self.positional_encoding = PositionalEncoding(self.d_model, dropout, max_len=input_shape[1] * input_shape[2])
         self.embed = nn.Embedding(self.vocab_size + 3, self.d_model)
 
         self.decoder = Decoder(
@@ -165,9 +165,10 @@ class MASTER(_MASTER, nn.Module):
 
         # Encode
         features = self.feat_extractor(x)['features']
-        b, c, h, w = features.shape[:4]
-        # --> (N, H * W, C)
-        features = features.reshape((b, c, h * w)).permute(0, 2, 1)
+        b, c, h, w = features.shape
+        # (N, C, H, W) --> (N, H * W, C)
+        features = features.view(b, c, h * w).permute((0, 2, 1))
+        # add positional encoding to features
         encoded = self.positional_encoding(features)
 
         out: Dict[str, Any] = {}
@@ -210,7 +211,7 @@ class MASTER(_MASTER, nn.Module):
         b = encoded.size(0)
 
         # Padding symbol + SOS at the beginning
-        ys = torch.ones((b, self.max_length), dtype=torch.long, device=encoded.device) * self.vocab_size + 2  # pad
+        ys = torch.zeros((b, self.max_length), dtype=torch.long, device=encoded.device) * self.vocab_size + 2  # pad
         ys[:, 0] = self.vocab_size + 1  # sos
 
         # Final dimension include EOS/SOS/PAD
--- a/doctr/models/recognition/master/pytorch.py
+++ b/doctr/models/recognition/master/pytorch.py
@@ -49,8 +49,6 @@ class MASTER(_MASTER, nn.Module):
         cfg: dictionary containing information about the model
     """
 
-    feature_pe: torch.Tensor
-
     def __init__(
         self,
         feature_extractor: nn.Module,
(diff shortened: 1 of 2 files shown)


felixdittrich92:
Todo: fix wrong masking (model is cheating) / revert transformer part to nn / model test/toy benchmark

felixdittrich92:
@frgfm i have some trouble bring it to work with the nn.TransformerDecoderLayer / target masking looks good, loss does also decrease but no way to reach any matches :sweat_smile: 

last working commit without nn.TransformerDecoderLayer
https://github.com/mindee/doctr/pull/941/commits/0f96095da3f60bd44f257cc3d342b5f7202b6622
also used for the toy run + benchmark 

What do you think iterate to find a solution with PyTorch's transformer implementation (which would be definitly recommended) or use the own one ?

charlesmindee (approved):
LGTM

doctr/models/recognition/master/pytorch.py
```diff
@@ -171,14 +182,11 @@ def forward(
             gt, seq_len = torch.from_numpy(_gt).to(dtype=torch.long), torch.tensor(_seq_len)
             gt, seq_len = gt.to(x.device), seq_len.to(x.device)
 
-        if self.training:
-            if target is None:
-                raise AssertionError("In training mode, you need to pass a value to 'target'")
-            tgt_mask = self.make_mask(gt)
+            # Compute source mask and target mask
+            source_mask, target_mask = self.make_source_and_target_mask(encoded, gt)
```
frgfm, line 186: the source_mask is needed whatever the mode (training vs inference) but the target_mask is only needed in training right?

We need to avoid forgetting cases (being in training mode != being passed a target), especially for the evaluation loops for our reference scripts (the model will be in `self.training = False` but it will be given a target so that it can compute the loss)

doctr/models/recognition/master/pytorch.py
```diff
@@ -206,17 +213,19 @@ def decode(self, encoded: torch.Tensor) -> torch.Tensor:
         b = encoded.size(0)
 
         # Padding symbol + SOS at the beginning
-        ys = torch.full((b, self.max_length), self.vocab_size + 2, dtype=torch.long, device=encoded.device)
-        ys[:, 0] = self.vocab_size + 1
+        ys = torch.zeros((b, self.max_length), dtype=torch.long, device=encoded.device) * (self.vocab_size + 2)  # pad
```
frgfm, line 216: ?
previously, the tensor was full of the value `self.vocab_size + 2`
now, it's full of zeros (multiplying 0 by `self.vocab_size + 2` will yield zero :sweat_smile:

doctr/models/recognition/transformer/pytorch.py
```diff
@@ -3,89 +3,165 @@
...
+    if mask is not None:
+        scores = scores.masked_fill(mask == 0, -1e9)
+    p_attn = torch.softmax(scores, dim=-1)
+    return torch.matmul(p_attn, value), p_attn
+
+
+class PositionwiseFeedForward(nn.Module):
+    """ Position-wise Feed-Forward Network """
+
+    def __init__(self, d_model: int, ffd: int, dropout=0.1) -> None:
```
frgfm, line 62: missing typing here :)

doctr/models/recognition/transformer/pytorch.py
```diff
@@ -3,89 +3,165 @@
...
+        super(PositionwiseFeedForward, self).__init__()
+        self.first_linear = nn.Linear(d_model, ffd)
+        self.sec_linear = nn.Linear(ffd, d_model)
+        self.dropout = nn.Dropout(p=dropout)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x = torch.relu(self.first_linear(x))
+        x = self.dropout(x)
+        x = self.sec_linear(x)
+        return x
```
frgfm, line 72: This is a sequential then :) 
Let's rewrite it as such :+1:

doctr/models/recognition/transformer/pytorch.py
```diff
@@ -3,89 +3,165 @@
...
+
+
+class MultiHeadAttention(nn.Module):
+    """ Multi-Head Attention """
+
+    def __init__(self, num_heads: int, d_model: int, dropout: float = 0.1) -> None:
+        super().__init__()
+        assert d_model % num_heads == 0
```
frgfm, line 80: The error message won't be very informative :sweat_smile:

frgfm (commented):
Thanks a lot @felixdittrich92 :pray: 
Late comments for improvements :)

doctr/models/recognition/master/pytorch.py
felixdittrich92: Oh f...k got it :sweat_smile: 
In the end, it's really just about having the shape, it doesn't matter what values ​​are in there. 
I also assumed that it would only make sense to fill with the PAD value, but that's irrelevant while inference.
But to make it make some sense, I'll change it :+1:

doctr/models/recognition/master/pytorch.py
felixdittrich92: decoder only:
train:
- we need the combined target mask to ensure the model cannot 'cheat'
- source mask is optional
inference:
- normally we don't need to pass any mask while inference ... but (MASTER is weird :sweat_smile:) .. without the masks the model will not work correctly .. i can't explain 100% why but what i think is, that we also while inference need the masking to force the models predictions 'step by step' otherwise we end up in something like this: 'aaaaabbbb'

Will update this :+1: (also crnn / sar pytorch)

Issue #932, opened 2022-05-27
conflicting dependencies


kiko0217:
### Bug description

I get error conflicting dependencies when installing doctr version tf-cpu with editable flag.  

### Code snippet to reproduce the bug

```bash
git clone -b v0.5.1 --depth 1 https://github.com/mindee/doctr.git
pip install -e doctr/.[tf-cpu]
```

### Error traceback

```
ERROR: Cannot install python-doctr, python-doctr[tf-cpu]==0.5.1a0 and tensorflow-cpu==2.8.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    python-doctr[tf-cpu] 0.5.1a0 depends on keras<2.7.0
    tensorflow-cpu 2.8.0 depends on keras<2.9 and >=2.8.0rc0
    python-doctr[tf-cpu] 0.5.1a0 depends on keras<2.7.0
    tensorflow-cpu 2.9.1 depends on keras<2.10.0 and >=2.9.0rc0
    python-doctr[tf-cpu] 0.5.1a0 depends on keras<2.7.0
    tensorflow-cpu 2.9.0 depends on keras<2.10.0 and >=2.9.0rc0
    python-doctr[tf-cpu] 0.5.1a0 depends on keras<2.7.0
    tensorflow-cpu 2.8.2 depends on keras<2.9 and >=2.8.0rc0
    python-doctr[tf-cpu] 0.5.1a0 depends on keras<2.7.0
    tensorflow-cpu 2.8.1 depends on keras<2.9 and >=2.8.0rc0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

### Environment

-

### Deep Learning backend

-

frgfm:
Hello @kiko0217 👋 

Thanks for the heads up! So this constraint was introduced in #579 because of a problem on the TF/Keras side at the time! Let us check if we can loosen this 👍

frgfm:
Closed by #963

PR #945, opened 2022-06-10
[deps] upgrade to tf 2.9 and remove keras

felixdittrich92:
This PR:

- updates tf to >=2.9
- removes keras
- change mean/std magc tf

Note: this upgrade triggers a lot of mypy issues i would say we handle this in another PR @frgfm wdyt ?
Same for TF mobilenet onnx issue

Otherwise if we want to fix the mypy issues in this PR i would be really happy to get some helping hands :hand: 

Todo: reduce the evaluate_detection CI runtime


felixdittrich92:
really weird i cannot reproduce this behaviour locally :sweat:

frgfm:
I would highly suggest to only update the requirements (potentially check whether removing the version specifier of keras doesn't cause harm) to close #932

This is a rather big change (for users I mean), so I wouldn't mix it with other modif unless they're directly related :)

felixdittrich92:
> I would highly suggest to only update the requirements (potentially check whether removing the version specifier of keras doesn't cause harm) to close #932
> 
> This is a rather big change (for users I mean), so I wouldn't mix it with other modif unless they're directly related :)

@frgfm yes i agree with only update tf / maybe remove keras and fix the mypy stuff. 
But we have to decide how we want to handle the (now) failing tests for TF onnx export models (mobilenet).
We could revert it for the moment complete or only exclude mobilenet models wdyt ?

felixdittrich92:
tackled by #963

Issue #947, opened 2022-06-11
[bug] doctr 0.5.1 | pypdfium2 --> AttributeError: module 'pypdfium2' has no attribute 'render_pdf_topil'


altomator:
### Bug description

I just installed doctr on OSX (TensofFlow), and I got an issue with pypdfium2 when reading a PDF file.

### Code snippet to reproduce the bug

```
import os
import urllib.request

# Let's pick the desired backend
os.environ['USE_TF'] = '1'

import matplotlib.pyplot as plt

from doctr.io import DocumentFile
from doctr.models import ocr_predictor

# Download a sample
urllib.request.urlretrieve("https://eforms.com/download/2019/01/Cash-Payment-Receipt-Template.pdf", "test.pdf")
# Read the file
doc = DocumentFile.from_pdf("test.pdf")
print(f"Number of pages: {len(doc)}")
```

### Error traceback

```
Traceback (most recent call last):
  File "/Users/bnf/Desktop/docTR/test.py", line 22, in <module>
    doc = DocumentFile.from_pdf("test.pdf")
  File "/Users/bnf/.virtualenvs/dl4cv/lib/python3.9/site-packages/doctr/io/reader.py", line 37, in from_pdf
    return read_pdf(file, **kwargs)
  File "/Users/bnf/.virtualenvs/dl4cv/lib/python3.9/site-packages/doctr/io/pdf.py", line 42, in read_pdf
    return [np.asarray(img) for img, _ in pdfium.render_pdf_topil(file, scale=scale, **kwargs)]
AttributeError: module 'pypdfium2' has no attribute 'render_pdf_topil'
```

### Environment

DocTR version: 0.5.1
TensorFlow version: 2.9.1
PyTorch version: N/A (torchvision N/A)
OpenCV version: 4.6.0
OS: Mac OSX 10.14.6
Python version: 3.9.13
Is CUDA available (TensorFlow): No
Is CUDA available (PyTorch): N/A
CUDA runtime version: No CUDA
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA

### Deep Learning backend

>>> print(f"is_tf_available: {is_tf_available()}")
is_tf_available: True
>>> print(f"is_torch_available: {is_torch_available()}")
is_torch_available: False

felixdittrich92:
Hi @altomator,
could you please downgrade pypdfium2 to version 1.0.0 and test it again ? With the next release we support v2.0.0 :)

[...]

PR #953, opened 2022-06-17
update pypdfium2 dep and improve code quality

felixdittrich92:
This PR:

- update pypdfium to 2.1.0 (to use with statement) and set fix to <3.0.0 to avoid #947 in upcoming releases
- improve code quality


doctr/io/pdf.py
```diff
@@ -41,8 +41,5 @@ def read_pdf(
         raise TypeError("unsupported object type for argument 'file'")
 
     # Rasterise pages to PIL images with pypdfium2 and convert to numpy ndarrays
-    pdf = pdfium.PdfDocument(file, password=password)
-    pages = [np.asarray(img) for img in pdf.render_topil(scale=scale, **kwargs)]
-    pdf.close()
-
-    return pages
+    with pdfium.PdfDocument(file, password=password) as pdf:
+        return [np.asarray(img) for img in pdf.render_topil(scale=scale, **kwargs)]
```
mara004, line 45: I'm not certain what impact a `return` statement inside the context manager might have. Will this actually call `__close__()`? To be safe, I'd rather cleanly assign  `pages = [...]` and then put the return outside the context manager.

mara004: Ah, according to [this thread](https://stackoverflow.com/questions/9885217/in-python-if-i-return-inside-a-with-block-will-the-file-still-close), it's safe to do this. You never stop learning!

felixdittrich92: Should be fine close is called like in a finally block so after returning :)

mara004:
Looks good, then!

frgfm (approved):
Thanks Felix, I was about to open a PR for this specifically :)
We should update the version specifiers of other dependencies accordingly I think

felixdittrich92:
> Thanks Felix, I was about to open a PR for this specifically :)
> We should update the version specifiers of other dependencies accordingly I think

:sweat_smile:  

About the deps part let us check if we can handle this with depend-a-bot this would make it much easier (really) ^^

frgfm:
That is a different problem though: dependabot is (to the best of my knowledge) reporting version specifier updates for security reasons

Here we need to reassess the upper and lower bound of each dep: I'll take care of it, and then set the dependabot :+1:

PR #958, opened 2022-06-23
docs: Documentation the reason for keras version specifier

frgfm:
This PR simply adds a reference to the reason behind the keras version specifier.

Any feedback is welcome!

cf #932


felixdittrich92 (approved):
Thaaanks :)

Issue #957, opened 2022-06-20
[Build] Update project build config and setup depend-a-bot


felixdittrich92:
### 🚀 The feature

We should update the current build configuration to a more modern one:

- instead of requirements files and a large setup.py use pyproject.toml instead
- for some reasons i would suggest to setup depend-a-bot to tackle some dependency issues and keep it actual

@frgfm i would assign this one to you :)
Is that ok for you ? 

### Motivation, pitch

better dependencies handling
reduce the amount of configuration specific files
hold dependencies up to date

frgfm:
Yes, sounds good! Thanks for opening this tracker :)
I'll deal with this shortly!

frgfm:
I will proceed in several parts:
- [ ] moves existing configs to pyproject.toml for tools (not build) #966 
- [ ] add black formatting
- [ ] moves requirements to pyproject.toml and make version specifier stricter
- [ ] set up the bot

felixdittrich92:
@frgfm do you think we can also change CI to circleci ?

frgfm:
I could, but why do you think it would be better?
In my opinion, considering the project is on Github, and the Github CI has a lot of useful features/actions, it's easier with that :)

felixdittrich92:
Mh yes but i see as a big disadvantage that it is not possible to choose a job specific instance type (for example SAR and MASTER Onnx tests are not possible in both frameworks in fact that it kills the provided instance type by Github CI -> locally works) and i think this will not be the only case .. another think is that we would be much more flexible for example we could scale the CI to specific jobs computation need .. do you know what i mean ? :)

felixdittrich92:
It is only a idea if you have another solution in mind would be also fine :) but we need anything to test also computation expensive workflows in our CI

frgfm:
Yes I get what you mean, but the CI is already cluttered with jobs either way :sweat_smile: 
I'll have to take a look and reassess what needs testing and whether we could do it more efficiently! Just FYI, we have an unusually high amount of features being tested by the CI (in a good way)

felixdittrich92:
> Yes I get what you mean, but the CI is already cluttered with jobs either way 😅 I'll have to take a look and reassess what needs testing and whether we could do it more efficiently! Just FYI, we have an unusually high amount of features being tested by the CI (in a good way)

[...]

PR #966, opened 2022-06-28
style: Fixed mypy and moved tool configs to pyproject.toml

frgfm:
As part of #957, this PR introduces the following modifications:
- moves all tool configs to pyproject.toml (flake8 doesn't support it yet)
- fixed mypy typing & isort in the entire project

Any feedback is welcome!


doctr/models/detection/differentiable_binarization/base.py
```diff
@@ -80,9 +80,9 @@ def polygon_to_box(
                     max_size = len(p)
             # We ensure that _points can be correctly casted to a ndarray
             _points = [_points[idx]]
-        expanded_points = np.asarray(_points)  # expand polygon
+        expanded_points: np.ndarray = np.asarray(_points)  # expand polygon
         if len(expanded_points) < 1:
-            return None
+            return None  # type: ignore[return-value]
```
felixdittrich92, line 85: @frgfm 
Offtopic (has nothing to do with your PR)
But that's a bit weird i have checked the method usage and `None` isn't handled / catched which would end up in an unexpected error :sweat_smile:

doctr/models/detection/linknet/base.py
```diff
@@ -78,9 +78,9 @@ def polygon_to_box(
                     max_size = len(p)
             # We ensure that _points can be correctly casted to a ndarray
             _points = [_points[idx]]
-        expanded_points = np.asarray(_points)  # expand polygon
+        expanded_points: np.ndarray = np.asarray(_points)  # expand polygon
         if len(expanded_points) < 1:
-            return None
+            return None  # type: ignore[return-value]
```
felixdittrich92, line 83: same

felixdittrich92:
@frgfm thanks a lot that's really helpful :)

some mypy issues left

doctr/models/detection/differentiable_binarization/base.py
frgfm: I checked that because I was a bit worried and there is:
https://github.com/mindee/doctr/blob/main/doctr/models/detection/differentiable_binarization/base.py#L134

felixdittrich92: Got it and if `self.assume_straight_pages` is False than it will fail with _box = None in https://github.com/mindee/doctr/blob/main/doctr/models/detection/differentiable_binarization/base.py#L146

felixdittrich92:
@frgfm I have canceled the jobs (runs since ~2hrs to find matching packages for tf): 

https://github.com/mindee/doctr/runs/7091359034?check_suite_focus=true   #dependency war :warning:

frgfm:
All good now :+1:

felixdittrich92 (approved):
Thanks a lot really nice :+1: 
Only one question what about flake (looks like it was not triggered) ? :)

frgfm:
> Thanks a lot really nice +1 Only one question what about flake (looks like it was not triggered) ? :)

It has been, but I changed the python runtime and renamed the job, so we have to update the "required" jobs :)

felixdittrich92:
> > Thanks a lot really nice +1 Only one question what about flake (looks like it was not triggered) ? :)
> 
> It has been, but I changed the python runtime and renamed the job, so we have to update the "required" jobs :)

Yeah i see and was currently on track to check this :sweat_smile:

Issue #791, opened 2022-01-10
Release tracker - v0.6.0


fg-mindee:
This issue is to be used to track the roadmap of docTR for release v0.6.0, and collect feedback from users & contributors.

For the v0.5.2 roadmap, please see #967

- Datasets
    - [x] #587
    - [x] Unify localization target to relative coordinates #775
    - [x] #935
    - [x] #955
    - [x] #827
    - [x] #782
    - [x] #987 
    - [x] #594 #889 #891 
    - [x] #955 
    - [x] #927
    - [x] #956
    - [x] #1038 
    - [x] #1041 

- Models 
    - [x] #789
    - [x] #819
    - [x] #563
    - [x] #808
    - [x] #883 
    - [x] #544
    - [x] #836 #892 #881 #896 
    - [x] #979 #980 #976 
    - [x] #991
    - [x] #902
    - [x] #960
    - [x] #964
    - [x] #1039
    - [x] #1013 
    - [x] #1050 #1055 #1063 #1059 #1072 
    - [x] #1061
    - [x] #1022 
    - [x] #1023 

- IO
    - [x] #947
    - [x] #953
    - [x] #944 
    - [x] #1019 

- Documentation
    - [x] Improves the general appearance of the documentation (cf. #851)
    - [x] #901 
    - [x] #959
    - [x] #936
    - [x] #937
    - [x] #983 
    - [x] #856, #859, #857, #862, #861
    - [x] #905
    - [x] #904
    - [x] #921 
    - [x] #901
    - [x] #937
    - [x] #958
    - [x] #929
    - [x] #993 
    - [x] #938 
    - [x] #959 
    - [x] #1070 
    - [x] #1002 
    - [x] #1017 
    - [x] #1076 
 
- References
    - [x] Add support of AMP for training in both frameworks #263
    - [x] #933 
    - [x] #696
    - [x] #263 
    - [x] #1062
 
- Tests
    - [x] Mock dataset for testing instead of downloading subsamples #680 
    - [x] #903 
    - [x] #680 
    - [x] #1054 

- Demo
    - [x] #658 
    - [x] #951 

- Code quality
    -[x] #953 

- Build
    - [x] #957
    - [x] #454
    - [x] #932 
    - [x] #998 
    - [x] #963 
    - [x] #966 
    - [x] #1033 
    - [x] #1034 
    - [x] #1031 
    - [x] #1028
    - [x] #1030  

- CI
    - [x] #853 
    - [x] #727

- Scripts
[...]

PR #976, opened 2022-07-05
[ONNX] Part1: update current onnx export 

felixdittrich92:
This PR:

- updates current onnx export for further processing in Part2 

Part2 : add slow tests (only local testing) for each model

### DIFF
--- a/doctr/models/utils/pytorch.py
+++ b/doctr/models/utils/pytorch.py
@@ -11,7 +11,7 @@ from torch import nn
 
 from doctr.utils.data import download_from_url
 
-__all__ = ['load_pretrained_params', 'conv_sequence_pt', 'export_classification_model_to_onnx']
+__all__ = ['load_pretrained_params', 'conv_sequence_pt', 'export_model_to_onnx']
 
 
 def load_pretrained_params(
@@ -92,19 +92,26 @@ def conv_sequence_pt(
     return conv_seq
 
 
-def export_classification_model_to_onnx(model: nn.Module, model_name: str, dummy_input: torch.Tensor) -> str:
-    """Export classification model to ONNX format.
+def export_model_to_onnx(
+    model: nn.Module,
+    model_name: str,
+    dummy_input: torch.Tensor,
+    **kwargs: Any
+) -> str:
 
+    """Export model to ONNX format.
     >>> import torch
     >>> from doctr.models.classification import resnet18
     >>> from doctr.models.utils import export_classification_model_to_onnx
     >>> model = resnet18(pretrained=True)
     >>> export_classification_model_to_onnx(model, "my_model", dummy_input=torch.randn(1, 3, 32, 32))
+    >>> export_model_to_onnx(model, "my_model", dummy_input=torch.randn(1, 3, 32, 32))
 
     Args:
         model: the PyTorch model to be exported
         model_name: the name for the exported model
         dummy_input: the dummy input to the model
+        kwargs: additional arguments to be passed to torch.onnx.export
 
     Returns:
         the path to the exported model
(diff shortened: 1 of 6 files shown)


doctr/models/utils/pytorch.py
```diff
@@ -116,7 +123,10 @@ def export_classification_model_to_onnx(model: nn.Module, model_name: str, dummy
         input_names=['input'],
         output_names=['logits'],
         dynamic_axes={'input': {0: 'batch_size'}, 'logits': {0: 'batch_size'}},
-        export_params=True, opset_version=13, verbose=False
+        export_params=True,
+        opset_version=14,
```
frgfm, line 127: Perhaps a comment on the opset_version? Is there a reason to upgrade? (ideally a comment saying why we need at least this version is perfect for future contributors)

doctr/models/utils/tensorflow.py
```diff
@@ -123,33 +122,45 @@ def __repr__(self) -> str:
...
+        kwargs: additional arguments to be passed to tf2onnx
 
     Returns:
         the path to the exported model and a list with the output layer names
     """
+    large_model = kwargs.get('large_model', False)
     model_proto, _ = tf2onnx.convert.from_keras(
         model,
-        opset=13,
+        opset=14,
```
frgfm, line 153: same here

frgfm (commented):
Thanks :) A few questions in comment

doctr/models/utils/tensorflow.py
```diff
@@ -123,33 +122,45 @@ def __repr__(self) -> str:
...
     Args:
         model: the keras model to be exported
         model_name: the name for the exported model
         dummy_input: the dummy input to the model
+        kwargs: additional arguments to be passed to tf2onnx
 
     Returns:
         the path to the exported model and a list with the output layer names
     """
+    large_model = kwargs.get('large_model', False)
```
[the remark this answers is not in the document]
felixdittrich92: This is a tricky part which has took me a bit time to understood 😅 
In short: while exporting if a weight grow > 2GB it reach the limit in protobuf (only the case with MASTER) so we need to split the model into parts (external tensor storage) these splitted weights are saved in single files and the graph as __PROTO_MODEL.onnx .. all compressed in a zip .. after this onnx is able to load all these related files in one step 

Will be more clear if you see the tests in Part 2 :)

doctr/models/utils/pytorch.py
felixdittrich92: Mh there is not directly a reason .. we only should keep the latest opset version for future models (higher opset version = more operators supported) :)

doctr/models/utils/tensorflow.py
felixdittrich92: Same answer as above 😅

doctr/models/utils/pytorch.py
frgfm: Alright but then according to https://pytorch.org/docs/master/onnx.html#torch.onnx.export we should match the max opset version of the oldest PyTorch version that we support no?

felixdittrich92: Mh.. yes i agree :) 👍

felixdittrich92:
@frgfm after testing again opset 14 is the minimum version we need to ensure all operators are supported (have added a comment) otherwise the export fails for some models in both frameworks 

We should also update torch to a minimum of 1.10 (first support for opset 14 in #791 requirements update)

doctr/models/utils/tensorflow.py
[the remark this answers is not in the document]
frgfm: We need to add a comment for this, it's far from obvious haha :)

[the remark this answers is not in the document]
felixdittrich92: @frgfm  now should be ok

Issue #790, opened 2022-01-10
[models] Ensure all TensorFlow models are SavedModel exportable


fg-mindee:
Most users of the library are more interested in existing pretrained models to use for inference rather than training. For this reason, it's important to ensure we can easily export those trained models.

- General
    - [ ] Add a unittest/CI job to check that all models can be exported with ONNX

- Classification
     - [ ] VGG
    - [ ] ResNet
    - [ ] MobileNet V3
    - [ ] MAGC ResNet

- Text Detection
    - [ ] LinkNet
    - [ ] DBNet

- Text Recognition
    - [ ] CRNN
    - [ ] SAR
    - [ ] MASTER

felixdittrich92:
@frgfm update to Onnx i think that's ok !? :)

frgfm:
If that's the case, would you mind linking all the PRs to the issue description? :pray:

felixdittrich92:
Closed with #1182

PR #980, opened 2022-07-07
[ONNX/Tests] Part2 add ONNX tests for TF and PT recognition models 

felixdittrich92:
This PR:

- adds test cases to ensure all recognition models can be exported into ONNX format in both frameworks
Note: MASTER PT can be exported but not loaded with onnxruntime there is an internal ticket i have linked it in the test
- introduces slow tests which are only tested on local machine (to ensure CI instance will not be killed :sweat_smile: )

Issues:
#789 #790

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,17 +12,17 @@ style:
 # Run tests for the library
 test:
 	coverage run -m pytest tests/common/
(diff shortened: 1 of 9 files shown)


charlesmindee (approved):
Thanks! LGTM

Issue #935, opened 2022-05-31
[datasets] Filter currupted and wrong annotated files in ready to use datasets


felixdittrich92:
### Bug description

While testing #933 i have seen some errors (empty crops / non utf-8 strings / and so on)
We need to filter some invalid files/annotations

- [x]  ensure all ready to use datasets works fine with eval_detection eval_recognition scripts (TF and PT)
- [ ]  unify recognition part with recognition dataset / word generator to return string directly instead of {labels: ['string']} #954 

**detection:**

- [x] CORD  (PT/TF) #983
- [x] FUNSD (PT/TF) #983 
- [x] IC03 (PT/TF) #983 
- [x] IC13 (PT/TF) validated by: @felixdittrich92 
- [x] IIIT5K (PT/TF) validated by: @felixdittrich92 
- [x] IMGUR5K (PT/TF) validated by: @felixdittrich92 
- [x] SROIE (PT/TF) #983 
- [x] SVHN (PT/TF) validated by: @felixdittrich92
- [x] SVT (PT/TF) #955 
- [x] SynthText (PT/TF) validated by: @felixdittrich92 

**recognition:**

- [x] MJSynth (PT/TF) #956 
- [x] CORD  (PT/TF) #983 
- [x] FUNSD (PT/TF) #983 
- [x] IC03 (PT/TF) #983 
- [x] IC13 (PT/TF) validated by: @felixdittrich92 
- [x] IIIT5K (PT/TF) validated by: @felixdittrich92 
- [x] IMGUR5K (PT/TF) #1038 
- [x] SROIE (PT/TF) #983 (NOTE: 99% contains whitespaces exluding not possible)
- [x] SVHN (PT/TF) #987 
- [x] SVT (PT/TF) #955 
- [x] SynthText (PT/TF)  #1038

frgfm:
@felixdittrich92 oh :/ On a specific dataset? during training?

felixdittrich92:
@frgfm Unfortunately no it affects all detection datasets used with `use_polygons=True`. 
I only noticed it later because some datasets aren't that big and i haven't had this on track before :/

frgfm:
Alright, let's get to the bottom of this :+1:

felixdittrich92:
https://github.com/bloomberg/memray super useful to find the mem leak

felixdittrich92:
@frgfm mem leak is inside LocalizationConfusion

felixdittrich92:
[...]

PR #983, opened 2022-07-11
[datasets] update IC / SROIE / FUNSD / CORD

felixdittrich92:
This PR:

- update pre_transform for last datasets without (unification & reduce memory usage)
- ensure IC datasets are working
- filter whitespaces and non utf-8 in FUNSD and CORD
- filter sroie recognition

Issue:

#935

### DIFF
--- a/doctr/datasets/cord.py
+++ b/doctr/datasets/cord.py
@@ -97,7 +97,9 @@ class CORD(VisionDataset):
                 crops = crop_bboxes_from_image(img_path=os.path.join(tmp_root, img_path),
                                                geoms=np.asarray(box_targets, dtype=int).clip(min=0))
                 for crop, label in zip(crops, list(text_targets)):
-                    self.data.append((crop, dict(labels=[label])))
+                    # filter labels with whitespaces
+                    if ' ' not in label:
+                        self.data.append((crop, dict(labels=[label])))
             else:
                 self.data.append((
                     img_path,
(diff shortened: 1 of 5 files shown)


doctr/datasets/cord.py
```diff
@@ -97,7 +97,9 @@ def __init__(
                 crops = crop_bboxes_from_image(img_path=os.path.join(tmp_root, img_path),
                                                geoms=np.asarray(box_targets, dtype=int).clip(min=0))
                 for crop, label in zip(crops, list(text_targets)):
-                    self.data.append((crop, dict(labels=[label])))
+                    # filter labels with whitespaces
+                    if ' ' not in label:
```
frgfm, line 101: just to make sure: what is the expected behaviour here?

Because I get the feeling that:
- it was made to delete labels that are white spaces
- but here it will delete label that have whitespaces (e.g. "Hello world")

I'm not saying it's something we can easily process but it's good to be transparent on that part

doctr/datasets/funsd.py
```diff
@@ -91,7 +91,9 @@ def __init__(
                 crops = crop_bboxes_from_image(img_path=os.path.join(tmp_root, img_path),
                                                geoms=np.asarray(box_targets, dtype=np_dtype))
                 for crop, label in zip(crops, list(text_targets)):
-                    self.data.append((crop, dict(labels=[label])))
+                    # filter labels with non utf-8 characters and whitespaces
+                    if not any(char in label for char in [' ', '☑', '☐', '\uf703', '\uf702']):
```
frgfm, line 95: don't we have a better way to filter those out? 
Perhaps encoding  decoding them through utf8 to check whether we get the same thing back?

frgfm (commented):
Late review again, but here are some comments :sweat_smile:

doctr/datasets/cord.py
felixdittrich92: Ah damn, i was a bit to much focused on the behavior we need inside docTR but yes i agree we cannot do this will be changed !

And no it was exactly made to avoid every label like this: "Hello World" the dataset does not contains empty labels

PR #987, opened 2022-07-14
[datasets] revert whitespace filtering and fix svhn reco

felixdittrich92:
This PR:

- reverts the whitespace filtering in FUNSD and CORD (is to much focused then to be used only inside docTR)
- update recognition eval scripts and corresponding CI job ( batch_size = 1 as default to ensure nothing is skipped in fact of whitespaces seperated labels)
- add fix for svhn dataset (recognition)

Any feddback is welcome :hugs:

### DIFF
--- a/references/recognition/evaluate_pytorch.py
+++ b/references/recognition/evaluate_pytorch.py
@@ -146,7 +146,7 @@ def parse_args():
     parser.add_argument('--vocab', type=str, default="french", help='Vocab to be used for evaluation')
     parser.add_argument('--dataset', type=str, default="FUNSD", help='Dataset to evaluate on')
     parser.add_argument('--device', default=None, type=int, help='device')
-    parser.add_argument('-b', '--batch_size', type=int, default=32, help='batch size for evaluation')
+    parser.add_argument('-b', '--batch_size', type=int, default=1, help='batch size for evaluation')
     parser.add_argument('--input_size', type=int, default=32, help='input size H for the model, W = 4*H')
     parser.add_argument('-j', '--workers', type=int, default=None, help='number of workers used for dataloading')
     parser.add_argument('--only_regular', dest='regular', action='store_true',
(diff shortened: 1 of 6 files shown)


felixdittrich92:
@frgfm I think it would be the best to filter the labels directly .. the dataset is fix and we know what we need to exclude feels cleaner instead of any encode-decode try/except block :sweat_smile: wdyt ?

references/recognition/evaluate_pytorch.py
```diff
@@ -146,7 +146,7 @@ def parse_args():
     parser.add_argument('--vocab', type=str, default="french", help='Vocab to be used for evaluation')
     parser.add_argument('--dataset', type=str, default="FUNSD", help='Dataset to evaluate on')
     parser.add_argument('--device', default=None, type=int, help='device')
-    parser.add_argument('-b', '--batch_size', type=int, default=32, help='batch size for evaluation')
+    parser.add_argument('-b', '--batch_size', type=int, default=1, help='batch size for evaluation')
```
frgfm, line 149: isn't it best to specify the batch size in the CI but leave a decent batch size for training?

references/recognition/evaluate_tensorflow.py
```diff
@@ -124,7 +124,7 @@ def parse_args():
     parser.add_argument('arch', type=str, help='text-recognition model to evaluate')
     parser.add_argument('--vocab', type=str, default="french", help='Vocab to be used for evaluation')
     parser.add_argument('--dataset', type=str, default="FUNSD", help='Dataset to evaluate on')
-    parser.add_argument('-b', '--batch_size', type=int, default=32, help='batch size for evaluation')
+    parser.add_argument('-b', '--batch_size', type=int, default=1, help='batch size for evaluation')
```
frgfm, line 127: same here

frgfm (approved):
Thanks! just added a few questions

And about removing white spaces early on, we can if it helps! I was just raising the point that depending on the dataset, that filter might seriously reduce the number of samples.

Either way we need to make a well-informed decision :)

references/recognition/evaluate_pytorch.py
felixdittrich92: It's only eval ;) in this case we ensure while try block that no sample is skipped which can be read (does not contain whitespace and chars in vocab)

felixdittrich92: CI uses a batch size of 32 to speedup so it's not relevant if samples are skipped for testing

PR #985, opened 2022-07-12
[api] upgrade fastapi and uvicorn and configure actions

felixdittrich92:
This PR:

- upgrades fastapi and uvicorn to latest version
- add delay (for model loading) and move docs ping at the end

Any feedback is welcome :hugs: 

Issue:
#847

PR #997, opened 2022-07-26
pytest-api test: fix ping server step

odulcy-mindee:
This PR aims two solve problems:

* to fix this error on CI : https://github.com/mindee/doctr/runs/7517599231?check_suite_focus=true
`curl` command in CI is run but the API is not fully ready so the `Ping server` step often failed. There is an option `--retry-all-errors` since curl 7.71.0 but there is only curl 7.68.0 on `ubuntu-latest` (which is ubuntu 20.04) and that's not that simple to update curl to this version.
I replaced `curl` by `wget` with option `--spider`.
From man page:
```
--spider
When invoked with this option, Wget will behave as a Web spider, which means that it will not download the pages, just check that they are there.
```

* fix https://github.com/mindee/doctr/issues/847. The problem comes from Tuples on fastapi, see https://github.com/tiangolo/fastapi/issues/4168, https://github.com/tiangolo/fastapi/issues/3782 and https://github.com/tiangolo/fastapi/issues/3665#issuecomment-1019579214. Updating `fastapi` version should fix the problem.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -18,15 +18,17 @@ jobs:
 
   pytest-api:
     runs-on: ubuntu-latest
+    env:
+        PORT=8002
     steps:
       - uses: actions/checkout@v2
       - name: Build & run docker
-        run: PORT=8002 docker-compose up -d --build
+        run: docker-compose up -d --build
       - name: Ping server
-        run: curl http://localhost:8002/docs
+        run: curl http://localhost:$PORT/docs
       - name: Install dependencies in docker
         run: |
-          PORT=8002 docker-compose exec -T web python -m pip install --upgrade pip
-          PORT=8002 docker-compose exec -T web pip install -r requirements-dev.txt
+          docker-compose exec -T web python -m pip install --upgrade pip
+          docker-compose exec -T web pip install -r requirements-dev.txt
       - name: Run docker test
-        run: PORT=8002 docker-compose exec -T web pytest .
+        run: docker-compose exec -T web pytest .
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -21,7 +21,7 @@ jobs:
     env:
         # Port for web container
         # See docker-compose.yml
-        PORT=8002
+        PORT: 8002
     steps:
       - uses: actions/checkout@v2
       - name: Build & run docker
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -19,6 +19,8 @@ jobs:
   pytest-api:
     runs-on: ubuntu-latest
     env:
+        # Port for web container
+        # See docker-compose.yml
         PORT=8002
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Build & run docker
         run: docker-compose up -d --build
       - name: Ping server
-        run: curl http://localhost:$PORT/docs
+        run: wget --spider --tries=12 http://localhost:$PORT/docs
       - name: Install dependencies in docker
         run: |
           docker-compose exec -T web python -m pip install --upgrade pip
--- a/api/requirements.txt
+++ b/api/requirements.txt
@@ -1,4 +1,6 @@
-fastapi>=0.73.0
+# Fastapi: minimum version required to avoid pydantic error
+# cf. https://github.com/tiangolo/fastapi/issues/4168
+fastapi>=0.73.0 
 uvicorn>=0.11.1
 python-multipart==0.0.5
 python-doctr>=0.2.0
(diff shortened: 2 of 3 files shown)


charlesmindee (dismissed):
LGTM thanks!

odulcy-mindee:
Work on `Ping server` step also done here: https://github.com/mindee/doctr/pull/985

felixdittrich92:
@frgfm

felixdittrich92:
Hi @odulcy-mindee I think both PR's will not solve the under issue mentioned by users it is only a fix for CI 😅

odulcy-mindee:
@felixdittrich92 Yeah, I just discovered that haha

odulcy-mindee:
A deprecated decorator from `pytest_asyncio` induced failures on this step: https://github.com/mindee/doctr/runs/7522290771?check_suite_focus=true.

There has been a new release on `pytest_asyncio`. It requires to update the decorators. See a similar problem here https://github.com/pytest-dev/pytest-asyncio/issues/390.

frgfm:
Hey everyone :wave: 

Here are my humble thoughts about this:
- for multi-OS testing, I think that would clutter the CI. This type of API is usually run on a Docker or similar isolated environment, so it makes sense that this is run a Linux-based docker 
- about the underlying steps, I've started checking the proper version specifier range for all our dependencies on the API, the underlying issue should be fixed soon :)
- regarding the arg deprecation of pytest_asyncio, I've also experienced this on other projects, I'll fix this :+1: 

No need to close this PR, there are a few things we'll need :ok_hand:

odulcy-mindee:
Ok, I was able to reproduce https://github.com/mindee/doctr/issues/847 on my Linux.

As @frgfm said, I think that it's still fine to test only on a Linux-based container. I'll continue to investigate the dependencies issues.

felixdittrich92:
Nice 👍 @frgfm has also something dependency related open in #957

odulcy-mindee:
@felixdittrich92 @frgfm The problem is coming from `Tuples` on `fastapi`, see https://github.com/tiangolo/fastapi/issues/4168, https://github.com/tiangolo/fastapi/issues/3782 and https://github.com/tiangolo/fastapi/issues/3665#issuecomment-1019579214.

I tested `fastapi>=0.73.0` and it worked :partying_face:

api/requirements.txt
```diff
@@ -1,4 +1,4 @@
-fastapi>=0.65.2
+fastapi>=0.73.0
```
felixdittrich92, line 1: We should at a short reference why we have changed it behind :)
fastapi>=0.73.0  # cf. https://github.com/mindee/doctr/issues/847

felixdittrich92 (changes requested):

felixdittrich92 (dismissed):
Thanks a lot looks good for me :)
@frgfm wdyt?

odulcy-mindee:
@felixdittrich92 Yes, you're right, I'll add a small note in requirements.txt

felixdittrich92 (approved):
Thanks :)

.github/workflows/docker.yml
```diff
@@ -18,15 +18,19 @@ jobs:
...
+        # See docker-compose.yml
+        PORT: 8002
     steps:
       - uses: actions/checkout@v2
       - name: Build & run docker
-        run: PORT=8002 docker-compose up -d --build
+        run: docker-compose up -d --build
       - name: Ping server
-        run: curl http://localhost:8002/docs
+        run: wget --spider --tries=12 http://localhost:$PORT/docs
```
frgfm, line 30: Could you remind me the reason of this change? (not familiar with the spider option, the "tries" flag seems explicit)

frgfm (commented):
Thanks for the fix!
I'll check whether we can afford a looser version specifier (constraining pydantic in this case)

Something I don't get though:
- since the specifier used to be `>=0.65.2`, when the docker is built, it was taking the highest version available (I think there is no indirect constraints here), which is 0.79 (0.78 up until recently)
- so the only reason I can see that this change fixed the bug, is because the docker couldn't reuse the cache.

Is there something I missed? According to the threads, the faulty fastapi version range is within (0.66 - 0.73).

thanks!

.github/workflows/docker.yml
felixdittrich92: spider does only an 'exist' check and avoid downloading anything

odulcy-mindee: @frgfm Also, `curl` has an option `--retry-all-errors` to retry in any case but this option is only available in latest version of `curl` which can't be (easily) installed on `Ubuntu 20.04`. So it may still fail for various reasons that we can't catch. However, `wget` comes out of the box with a "retry on all errors".

odulcy-mindee: @frgfm Concerning your comment on version, it's a good remark. I didn't changed anything on my side and it didn't install the latest version of fastapi :thinking:

Issue #1003, opened 2022-07-30
Request For Adding ParSeq - text recognition model


nikokks:
### 🚀 The feature

Hello,

I mainly use the text detection and text recognition models with your framework. 

As I have seen: the most recent models that you propose in text recognition, namely MASTER and SAR, are not yet operational.

However at the text recognition level, there is a recent model that gets very impressive performances: PasrSeq.

Here are the references: 
https://github.com/open-mmlab/mmocr/blob/main/configs/textrecog/master/README.md
https://github.com/open-mmlab/mmocr/blob/main/configs/textrecog/abinet/README.md
https://paperswithcode.com/paper/scene-text-recognition-with-permuted#code
https://github.com/baudm/parseq

Would it be possible to add this recognition text to the models you propose?

Thanks a lot for your work ! 

### Motivation, pitch

I'm working with text recognition models and a recent model in state of the art outperforms on all test datasets.
I would like use this model with your framework pipelines

felixdittrich92:
Hi @nikokks 👋 ,
fyi MASTER and SAR are fixed on main branch will be released soon with 0.5.2

I agree with the ParSeq addition firstly i have had in mind to add ViTSTR but yes i think we can switch directly to ParSeq instead of this👍 

Are you maybe interested to open a PR for this model we would be happy to help with !
Otherwise we could take it in the near future we have definitly on track to add more models and i totally agree that ParSeq has to be one of it :)

frgfm:
I agree that this couldbe a good candidate for new text recognition models in 0.6.0 :)
Perhaps we should open a tracker issue for all new models request that we would stage for 0.6.0, or directly link them in the release tracker?

felixdittrich92:
@frgfm I would say a seperate issue where we can track all requested model additions (splitted in detection / recognition / TF / PT with paper/repo link) and link this issue in the release tracker. wdyt ? Do you like to open it ? :)

frgfm:
@felixdittrich92  Done :)

felixdittrich92:
@nikokks @frgfm Do you agree if we close this ticket ? Should be fine if we track it in mentioned issue :)

frgfm:
I think we should keep this:
- the PR will close this issue, and the release tracker will get the checkbox ticked

So no need to close it, and that will notify @nikokks when this gets resolved, which I guess is of interest to him :)

felixdittrich92:
👍

nikokks:
[...]

PR #1013, opened 2022-08-08
move transformers implementation to modules

felixdittrich92:
This PR:

- moves transformers from recognition to modules to make it more global available for other implementations
- fix flake8 issues

Related to:
#1003


odulcy-mindee (approved):
Thanks ! Should we also add these modifications to sphinx docs ?

felixdittrich92:
> Thanks ! Should we also add these modifications to sphinx docs ?

That's only internal used (currently for MASTER implementation) i think we do not need to add it to the docs in fact that we do not want to make it public for users :)
It's more a place for modules/layers which are widely used in different implementations so that we can reuse this for future implementations

PR #1017, opened 2022-08-10
Changes needed to be able to use doctr on AWS Lambda

mtvch:
I want to use `doctr` on `AWS Lambda`, but there is a restriction: it gives a function write access only under `/tmp` folder.
I have found two places where `doctr` claims write access outside this folder:
1. While caching downloaded models.
2. By using `ThreadPool` from python's `multiprocessing` package: it uses `/dev/shm` folder for shared memory which is not present on `AWS Lambda`.

That's why I think it would be nice to add an option for user to decide where models should be cached and whether to use `multiprocessing` or not.

### DIFF
--- a/doctr/utils/data.py
+++ b/doctr/utils/data.py
@@ -70,7 +70,10 @@ def download_from_url(
         file_name = url.rpartition('/')[-1]
 
     if not isinstance(cache_dir, str):
-        cache_dir = os.path.join(os.path.expanduser('~'), '.cache', 'doctr')
+        if os.environ['DOCTR_CACHE_DIR']:
+            cache_dir = os.environ['DOCTR_CACHE_DIR']
+        else:
+            cache_dir = os.path.join(os.path.expanduser('~'), '.cache', 'doctr')
 
     # Check hash in file name
     if hash_prefix is None:
--- a/doctr/utils/multithreading.py
+++ b/doctr/utils/multithreading.py
@@ -4,6 +4,7 @@
 # See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0.txt> for full license details.
 
 
+import os
 import multiprocessing as mp
 from multiprocessing.pool import ThreadPool
 from typing import Any, Callable, Iterable, Optional
@@ -29,8 +30,8 @@ def multithread_exec(func: Callable[[Any], Any], seq: Iterable[Any], threads: Op
 
     threads = threads if isinstance(threads, int) else min(16, mp.cpu_count())
     # Single-thread
-    if threads < 2:
-        results = map(func, seq)
+    if threads < 2 or os.environ['DOCTR_CONCURRENCY_DISABLE'] == 'true':
+        results = list(map(func, seq))
     # Multi-threading
     else:
         with ThreadPool(threads) as tp:


felixdittrich92:
Hi @mtvch :wave:,
that's a really interesting use case thanks for opening :+1: 

Some points:
- we should add a short test that it works as expected:

https://github.com/mindee/doctr/blob/main/tests/common/test_utils_multithreading.py
create test_utils_data.py in tests/common folder and add a short test 

- we need to document this for other users i would suggest to add a page to the documentation in using_doctr maybe something like 'Run on AWS' with a subpoint AWS Lambda 

About the casting to list in `multithread_exec` : #931

doctr/utils/data.py
```diff
@@ -70,7 +70,10 @@ def download_from_url(
         file_name = url.rpartition('/')[-1]
 
     if not isinstance(cache_dir, str):
-        cache_dir = os.path.join(os.path.expanduser('~'), '.cache', 'doctr')
+        if os.environ['DOCTR_CACHE_DIR']:
```
odulcy-mindee, line 73:
```suggestion
        if 'DOCTR_CACHE_DIR' in os.environ:
```

doctr/utils/multithreading.py
```diff
@@ -29,8 +30,8 @@ def multithread_exec(func: Callable[[Any], Any], seq: Iterable[Any], threads: Op
 
     threads = threads if isinstance(threads, int) else min(16, mp.cpu_count())
     # Single-thread
-    if threads < 2:
-        results = map(func, seq)
+    if threads < 2 or os.environ['DOCTR_CONCURRENCY_DISABLE'] == 'true':
```
odulcy-mindee, line 33:
```suggestion
    if threads < 2 or os.environ.get('DOCTR_CONCURRENCY_DISABLE', "").upper() in ENV_VARS_TRUE_VALUES:
```

and use
```python3
from doctr.file_utils import ENV_VARS_TRUE_VALUES
```

odulcy-mindee (dismissed):
Hello @mtvch, thank you very much for opening this PR and to bring us this use case !

I would suggest some changes in the code and also I agree with @felixdittrich92 on the parts that he highlighted

doctr/utils/multithreading.py
felixdittrich92: good catch looks better of course :)

doctr/utils/data.py
```diff
@@ -70,7 +70,10 @@ def download_from_url(
...
+            cache_dir = os.environ['DOCTR_CACHE_DIR']
+        else:
+            cache_dir = os.path.join(os.path.expanduser('~'), '.cache', 'doctr')
```
frgfm, line 76: I'd suggest a simpler version with the get method:
```python
cache_dir = os.environ.get("DOCTR_CACHE_DIR", os.path.join(os.path.expanduser("~"), ".cache", "doctr"))
```

doctr/utils/multithreading.py
frgfm: Mmmh, I'd say it would be better to disable the concurrency by passing a `threads` value inferior to 2

Relatively to docTR, this is a relatively low-level function. Also, I thought the purpose of this PR was to make docTR compatible with AWS. If that's the case, from what I understood, the only thing we have to be wary about is passing `threads < 2`. So technically, I'd argue there is nothing to change in this file.

But I understand that could be easier here, so if everyone's on board, I agree with Olivier's suggestion :+1:

doctr/utils/multithreading.py
```diff
@@ -29,8 +30,8 @@ def multithread_exec(func: Callable[[Any], Any], seq: Iterable[Any], threads: Op
...
+        results = list(map(func, seq))
```
frgfm, line 34: As mentioned by Felix, this cannot be changed in this PR. It's a complicated/dangerous change, and there is a PR for that.

After giving some thought to it, we'll go with the option where we return a generator whatever the input data structure. The user will be free to cast it to whatever they want. (to give you a dangerous/silent issue: if you have a tuple, pass it to this, and check that the output type is still a tuple and not none to perform a conditional operation, that will be a problem).

Another solution would be to dynamically cast it to the input type but I'm a bit afraid of the consequences if someone passes a sequence-like / generator-like custom data structures (it can be processed technically by the function, but casting it back will meddle with custom user code)

frgfm (changes requested):
Thanks for the PR!

I agree with most of what has already been said! A few points:
- we need some unittests for those
- the docstring of each function should reflect any behaviour change (since now env var will modify/override the behaviour)
- the casting of the multiprocessing output is being handled by another PR, it's a dangerous change that should be taken care of outside of here :) 

Cheers :v:

doctr/utils/data.py
felixdittrich92: LGTM

odulcy-mindee: Even better ! :+1:

doctr/utils/multithreading.py
odulcy-mindee: @frgfm Yes, I agree with you, it's better to handle this way: to propagate correctly the `threads` parameter.

- However, I just discovered that `threads` has not been passed in `doctr/models/preprocess/{pytorch, tensorflow}.py` so we need to propagate this parameter to higher levels -> another PR ?

- If we want to let enforce `thread=1` in AWS Lambda environment, according to [AWS Lambda Documentation about environment variables](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html), we can check that `AWS_LAMBDA_FUNCTION_NAME` is present in the environment variable.
So it should be:
```suggestion
    if threads < 2 or os.environ.get("AWS_LAMBDA_FUNCTION_NAME") is not None:
```

mtvch:
Thank you all for responses!
Can I open another PR about changing return type of `multithread_exec` function to generator?
As far as I have checked every place where it's called expects a list as a result, so this change looks easy to me

doctr/utils/multithreading.py
mtvch: This change is good in the sense user has less environment variables to configure, but every cloud platform I know offers Lambda-like service. And if another platform has similar security policy, it won't work (or user would have to manually set this variable, which is wierd). That's why I would stick to variable name like `DOCTR_DISABLE_MULTIPROCESSING`

doctr/utils/data.py
mtvch: Wow! looks good! I didn't know it accepts default value as second argument
Also maybe it's worth to check if directory is writable and if it's not hint user that they can set `DOCTR_CACHE_DIR` to writable directory. That's what `matplotlib` did and it was really helpful

felixdittrich92:
> Thank you all for responses! Can I open another PR about changing return type of `multithread_exec` function to generator? As far as I have checked every place where it's called expects a list as a result, so this change looks easy to me

Hi @mtvch :wave: , there is currently an open PR to do this #931 
I have asked if he want to continue otherwise yes of course feel free to open another PR to solve this :+1:

mtvch:
Hello, @felixdittrich92!
I have updated this pr and opened another pr #1019 about `multithread_exec` return type. Could you, please, review them?

doctr/utils/data.py
```diff
@@ -64,13 +65,17 @@ def download_from_url(
...
+    Note:
+        You can change cache directory location by using `DOCTR_CACHE_DIR` environment variable.
     """
 
     if not isinstance(file_name, str):
         file_name = url.rpartition('/')[-1]
 
     if not isinstance(cache_dir, str):
-        cache_dir = os.path.join(os.path.expanduser('~'), '.cache', 'doctr')
+        cache_dir = os.environ.get(DOCTR_CACHE_DIR_ENV_NAME, os.path.join("~", ".cache", "doctr"))
```
felixdittrich92, line 77:
```
cache_dir = os.path.expanduser(
    os.environ.get(DOCTR_CACHE_DIR, os.path.join("~", ".cache", "doctr"))
)
```
or 
```
cache_dir = os.path.expanduser(DOCTR_CACHE_DIR_PATH)
```

doctr/utils/data.py
```diff
@@ -23,6 +23,7 @@
 # matches bfd8deac from resnet18-bfd8deac.ckpt
 HASH_REGEX = re.compile(r'-([a-f0-9]*)\.')
 USER_AGENT = "mindee/doctr"
+DOCTR_CACHE_DIR_ENV_NAME = "DOCTR_CACHE_DIR"
```
felixdittrich92, line 26: I think we can remove this if there is no path to an existing folder as DOCTR_CACHE_DIR env variable set than it will use our default or we do:

`DOCTR_CACHE_DIR_PATH = os.environ.get(DOCTR_CACHE_DIR, os.path.join("~", ".cache", "doctr"))`

doctr/utils/data.py
```diff
@@ -84,8 +89,14 @@ def download_from_url(
         logging.info(f"Using downloaded & verified file: {file_path}")
         return file_path
 
-    # Create folder hierarchy
-    folder_path.mkdir(parents=True, exist_ok=True)
+    try:
```
felixdittrich92, line 92: It would be great if you add an test case for `download_from_url` in tests/common/test_utils_data.py (you need to create this file) :+1: 

you can use this url as test download : 
https://github.com/mindee/doctr/releases/download/v0.5.1/word-crop.png

doctr/utils/data.py
felixdittrich92: I would prefer the second one :)

doctr/utils/data.py
```diff
@@ -84,8 +89,14 @@ def download_from_url(
...
+        # Create folder hierarchy
+        folder_path.mkdir(parents=True, exist_ok=True)
+    except OSError:
+        print(f"""
+        Failed creating cache direcotry: {cache_dir}
```
felixdittrich92, line 97: minor typo -> directory @frgfm @odulcy-mindee are you ok with printing instead of logging in this case ?

felixdittrich92 (changes requested):
Thanks for applying the changes :+1:  
I have added some minor comments :)

@fharper can you review the minor docs change please ?

doctr/utils/data.py
felixdittrich92: BTW: there is also an use in: https://github.com/mindee/doctr/blob/26ed13221757a4f13df62bcfff86e868c1b1dbe3/doctr/datasets/datasets/base.py#L99
we should update this also if the env var is set

felixdittrich92:
```
cache_dir = os.environ.get(
    "DOCTR_CACHE_DIR", 
    os.path.join(os.path.expanduser("~"), ".cache", "doctr")
) if cache_dir is None else cache_dir
```

doctr/utils/data.py
odulcy-mindee: I would suggest to raise the exception here because if the function is unable to create the cache dir, it probably means that the python script lacks permissions to create the folder or the disk is full or any other I/O Error (`exist_ok=True` so it will not raise if the folder already exists) so there is no need to try to download and save files because it'll fail.

What about:

```python
    except OSError as e:
        error_message = f"Failed creating cache direcotry at {folder_path}"
        if DOCTR_CACHE_DIR_ENV_NAME in os.environ:
            error_message += f" using path from {DOCTR_CACHE_DIR_ENV_NAME} environment variable."
        else:
            error_message += f". You can change default cache directory using {DOCTR_CACHE_DIR_ENV_NAME} environment variable if needed."
        raise OSError(error_message) from e
```

Also note it should be `folder_path` and not `cache_dir` in error message

docs/source/using_doctr/running_on_aws.rst
```diff
@@ -0,0 +1,7 @@
+AWS Lambda
+========================
+
+AWS Lambda's security policy does not allow you to write anywhere outside `/tmp` directory.
```
frgfm, line 4: Perhaps we should add a hyperlink for people who don't know what AWS Lamba is

docs/source/using_doctr/running_on_aws.rst
```diff
@@ -0,0 +1,7 @@
+AWS Lambda
```
frgfm, line 1: we need to reference this in the `index.rst` otherwise it won't show up in the documentation

doctr/utils/data.py
frgfm: I agree :+1:

doctr/utils/data.py
frgfm: About information display, I would either go for exception raising or logging, but no print. Logging level can be set by the user as they wish, exception can be caught, `print` doesn't have that flexibility :sweat_smile:

doctr/utils/data.py
frgfm: If what we are trying to do here is to check whether the folder is writable, I think we shouldn't do a `try except`. `os.access` provides information about a folder writability for instance. If we're trying to cover other things, perhaps the OSError catching makes sense, but it would be better to avoid that!

And yes, we need a unittest for this :sweat_smile:

doctr/utils/multithreading.py
```diff
@@ -25,11 +28,18 @@ def multithread_exec(func: Callable[[Any], Any], seq: Iterable[Any], threads: Op
 
     Returns:
         iterable of the function's results using the iterable as inputs
+
+    Notes:
+        This function uses ThreadPool from multiprocessing package, which
+        uses `/dev/shm` directory for shared memory.
+        If you do not have write permissions for this directory (like in a scenario when you run doctr on AWS Lambda),
+        you might want to disable multiprocessing.
+        To achieve that, set 'DOCTR_MULTIPROCESSING_DISABLE' to 'TRUE'.
```
frgfm, line 37: I think there are more line breaks than needed :)

doctr/utils/multithreading.py
```diff
@@ -25,11 +28,18 @@ def multithread_exec(func: Callable[[Any], Any], seq: Iterable[Any], threads: Op
 
     Returns:
         iterable of the function's results using the iterable as inputs
+
+    Notes:
+        This function uses ThreadPool from multiprocessing package, which
+        uses `/dev/shm` directory for shared memory.
+        If you do not have write permissions for this directory (like in a scenario when you run doctr on AWS Lambda),
```
frgfm, line 35: "like in a scenario when you run doctr on AWS Lambda" -> "if you run `doctr` on AWS Lambda for instance"

frgfm (commented):
Thanks for the last commits! A few comments left :)

doctr/utils/data.py
mtvch: I decided to stick with catching OSError, because `os.access` doesn't help when directory does not exist

felixdittrich92:
Hi @mtvch 👋 ,
I think you missed some suggestions from above would be great if you can add the missing ones :) 👍

mtvch:
Hello, @felixdittrich92!
Excuse me, but, honestly, I don't know what I missed(

There are two things I didn't do as you suggested:
1. I didn't use `os.access` function, because it doesn't work if directory does not exist (returning `False` even if you can create directory).
2. I didn't write test for exactly downloading files, instead I used mocks and covered every line of this pr with tests.

Did you mean these two? If not, please, could you tell me what suggestions I missed and I will implement them

mtvch:
I have found one more thing I couldn't implement: using `os.environ.get` with default value in a second argument.
I decided to stick with `is None` check, because `mypy` was failing.
It seems that `os.environ.get` has `Optional[str]` specification even if I provide default value in a second argument. 
image

doctr/utils/data.py
```diff
@@ -64,13 +64,20 @@ def download_from_url(
...
+    Note:
+        You can change cache directory location by using `DOCTR_CACHE_DIR` environment variable.
     """
 
     if not isinstance(file_name, str):
         file_name = url.rpartition("/")[-1]
 
     if not isinstance(cache_dir, str):
-        cache_dir = os.path.join(os.path.expanduser("~"), ".cache", "doctr")
+        cache_dir = os.environ.get("DOCTR_CACHE_DIR")
```
felixdittrich92, line 76: @mtvch same here

```
cache_dir = os.environ.get(
    "DOCTR_CACHE_DIR", 
    str(os.path.join(os.path.expanduser("~"), ".cache", "doctr"))
)
```

doctr/utils/data.py
```diff
@@ -84,8 +91,19 @@ def download_from_url(
...
         return file_path
 
-    # Create folder hierarchy
-    folder_path.mkdir(parents=True, exist_ok=True)
+    try:
+        # Create folder hierarchy
+        folder_path.mkdir(parents=True, exist_ok=True)
+    except OSError:
+        error_message = f"Failed creating cache direcotry at {folder_path}"
+        if "DOCTR_CACHE_DIR" in os.environ:
```
felixdittrich92, line 99: `if os.environ.get("DOCTR_CACHE_DIR", ""):` no need to iter trough a list

felixdittrich92 (changes requested):
@mtvch thanks for applying all the requested changes i have added some last minor points then we are good to go :hugs:

mtvch:
> @mtvch thanks for applying all the requested changes i have added some last minor points then we are good to go hugs

@felixdittrich92
Thank you for your patience and another review!

felixdittrich92 (approved):
Thanks for all the updates. LGTM :+1:

Issue #973, opened 2022-07-03
Orientation details are not available in Json exported


gnanads:
### Discussed in https://github.com/mindee/doctr/discussions/971

<div type='discussions-op-text'>

<sup>Originally posted by **gnanads** July  1, 2022</sup>
Hi Team,

When I exported the ocr_predictor results, I found orientation and Language information are missing.

Below is the code snippet:

```
model = ocr_predictor('linknet_resnet18_rotation','crnn_vgg16_bn',pretrained=True,assume_straight_pages=False, preserve_aspect_ratio=True)
result = model([image])
export = result.export()
print(export)
```

Result json:
{'page_idx': 0, 'dimensions': (912, 617), 'orientation': {'value': None, 'confidence': None}, 'language': {'value': None, 'confidence': None}</div>

nikokks:
Hello,

the function used to get the orientation of the document uses function with open CV. 
Why did you not used the angles'median of the larger sides of the bboxes ?

PR #1022, opened 2022-08-16
Fix orientation details export

aminemindee:
Hello this PR is to deal with the problem discussed here: #973 

- [x] Add orientation computation and output it
- [x] Add language detection model
- [x] Add it to predictor and output result
- [x] Add tests


doctr/models/builder.py
```diff
@@ -272,7 +272,9 @@ def __call__(
         self,
         boxes: List[np.ndarray],
         text_preds: List[List[Tuple[str, float]]],
-        page_shapes: List[Tuple[int, int]]
+        page_shapes: List[Tuple[int, int]],
+        orientations: Optional[List[Dict[str, Any]]] = [],
```
frgfm, line 276: Two things here:
1. this is typed as `Optional`, so the default would be `None`
2. having complex data types such as tuple or list as default of a function have a few underlying issues in rare cases. To make it short, let's not do this :sweat_smile:

doctr/models/builder.py
```diff
@@ -272,7 +272,9 @@ def __call__(
...
+        languages: List[Dict[str, Any]] = [],
```
frgfm, line 277: Same here, let's make this optional

doctr/models/builder.py
```diff
@@ -288,6 +290,10 @@ def __call__(
         if len(boxes) != len(text_preds) or len(boxes) != len(page_shapes):
             raise ValueError("All arguments are expected to be lists of the same size")
 
+        if not orientations:
```
frgfm, line 293: better to be specific:
```python
if not isinstance(orientations, list):
```

frgfm, line 293: better to be specific:
```python
if not isinstance(orientations, list):
```

doctr/models/builder.py
```diff
@@ -288,6 +290,10 @@ def __call__(
...
+            orientations = [dict(value=None, confidence=None) for _ in range(len(boxes))]
+        if not languages:
```
frgfm, line 295: same here

doctr/models/builder.py
frgfm: Actually, better to do this:
```python
_orientations = orientations if isinstance(orientations, list) else [None] * len(boxes)
```

The `Page` constructor will handle None accordingly

requirements-pt.txt
```diff
@@ -15,3 +15,4 @@ Pillow>=8.3.2
 tqdm>=4.30.0
 rapidfuzz>=1.6.0
 huggingface-hub>=0.4.0
+fasttext>=0.9.2
```
frgfm, line 18: would you mind also adding a relevant upper bound in the specifier please? :pray:

frgfm (changes requested):
Thanks a lot for the PR and taking care of this!

I added a few comments but here is some high level feedback:
- I feel like this PR should be split in two: keeping this one about orientation, and opening an other for language detection (I highly suggest opening a GH issue to discuss how this should be done before going for the PR)
- the PR adds a dependency, we need to be gradually more careful about this. Hence my suggestion about design discussion in an issue :)
- you added a binary file in the PR "lid.176.ftz". I think this isn't on purpose, but we need to remove it :sweat_smile: 

Cheers :v:

requirements-pt.txt
felixdittrich92: @frgfm @aminemindee we should check If we can use `langdetect` package instead of fasttext seems to be more lightweight 👍

aminemindee: I first used langdetect but for the same small text the model will predict different languages and using `DetectorFactory.seed = 0` did not solve that issue so i went for fasttext instead. But yeah it's slower. 
For this part about language detection, we can discuss on the new pr with just that part #1023

doctr/models/builder.py
aminemindee: Yes my bad :+1: 
Done and pushed.

doctr/models/builder.py
aminemindee: Done and pushed.

doctr/models/builder.py
aminemindee: took the suggestions from the comment after this one :+1:

aminemindee: Done and pushed.

doctr/models/builder.py
aminemindee: Done and pushed.

pyproject.toml
```diff
@@ -28,6 +28,7 @@ module = [
 	"huggingface_hub.*",
 	"pypdfium2.*",
 	"rapidfuzz.*",
+	"fasttext.*",
```
felixdittrich92, line 31: remove :)

requirements-pt.txt
felixdittrich92: remove :)

requirements.txt
```diff
@@ -16,3 +16,4 @@ tensorflow-addons>=0.17.1
 tf2onnx>=1.9.2
 rapidfuzz>=1.6.0
 huggingface-hub>=0.4.0
+fasttext>=0.9.2
```
felixdittrich92, line 19: same :)

setup.py
```diff
@@ -63,6 +63,7 @@
     "rapidfuzz>=1.6.0",
     "tf2onnx>=1.9.2",
     "huggingface-hub>=0.4.0",
+    "fasttext>=0.9.2",
```
felixdittrich92, line 66: same :)

setup.py
```diff
@@ -110,6 +111,7 @@ def deps_list(*pkgs):
     deps["tqdm"],
     deps["rapidfuzz"],
     deps["huggingface-hub"],
+    deps["fasttext"],
```
felixdittrich92, line 114: same :)

felixdittrich92 (changes requested):
Thanks for applying the suggested changes from @frgfm 

Only to remove fasttext than looks good to me :+1:

requirements-pt.txt
aminemindee: ah my bad ^^ removed it

pyproject.toml
aminemindee: done and pushed.

requirements.txt
aminemindee: done and pushed.

aminemindee: done and pushed.

setup.py
aminemindee: done and pushed.

setup.py
aminemindee: done and pushed.

felixdittrich92 (changes requested):
@aminemindee sry my mistake we need to add a short docstring for `OCRPredictor ` in doctr/models/predictor/   (pytorch & tensorflow) 

something like:
detect_orientation: if True, the estimated general page orientation will be added to the predictions for each page. Doing so will slightly deteriorate the overall latency.

doctr/models/predictor/tensorflow.py
```diff
@@ -65,6 +67,13 @@ def __call__(
         origin_page_shapes = [page.shape[:2] for page in pages]
 
         # Detect document rotation and rotate pages
+        if self.detect_orientation:
+            origin_page_orientations = [estimate_orientation(page) for page in pages]
```
felixdittrich92, line 71: If detect_orientation=True and straighten_pages=True we should reuse origin_page_orientations otherwise we do it twice

tests/tensorflow/test_models_zoo_tf.py
```diff
@@ -91,6 +92,9 @@ def test_trained_ocr_predictor(mock_tilted_payslip):
                                  [0.51385817, 0.21002172]])
     assert np.allclose(np.array(out.pages[0].blocks[1].lines[0].words[-1].geometry), geometry_revised)
 
+    orientation = 30.415687561035156
```
felixdittrich92, line 95: @frgfm @charlesmindee

tests/pytorch/test_models_zoo_pt.py
```diff
@@ -57,6 +58,9 @@ def test_ocrpredictor(mock_pdf, mock_vocab, assume_straight_pages, straighten_pa
         input_page = (255 * np.random.rand(1, 256, 512, 3)).astype(np.uint8)
         _ = predictor([input_page])
 
+    orientation = 0
```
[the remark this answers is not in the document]
felixdittrich92: If i take a look inside `mock_pdf` 0 should be correct

tests/tensorflow/test_models_zoo_tf.py
felixdittrich92: this value seems to be wrong

tests/pytorch/test_models_zoo_pt.py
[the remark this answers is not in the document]
felixdittrich92: there is also a missing test for [False, False] case

tests/tensorflow/test_models_zoo_tf.py
felixdittrich92: a holy now i got it :sweat_smile: @aminemindee please move this test inside `test_ocrpredictor` :+1:

aminemindee: Done and pushed. Yeah it's better i used there in pytorch but not on tf haha my bad

tests/pytorch/test_models_zoo_pt.py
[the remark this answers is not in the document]
felixdittrich92: @aminemindee would be nice if you can add this case in this PR :)
https://github.com/mindee/doctr/blob/26ed13221757a4f13df62bcfff86e868c1b1dbe3/tests/pytorch/test_models_zoo_pt.py#L17     -> add [False, False]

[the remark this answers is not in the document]
aminemindee: Done and pushed. (added also the assume_straight_pages variable when you define the detector was missing ^^ the test failed without it )

doctr/models/predictor/tensorflow.py
aminemindee: Done and pushed. Added also the doc about the detect_orientation param :+1:

felixdittrich92 (approved):
@aminemindee Thanks a lot for applying all changes :+1:

felixdittrich92:
Hi @aminemindee :wave:, 
the Actions issue on GH side has destroyed anything .. build actions stucks in a Queue and i cannot trigger or cancel it (same for manual merge) :sweat_smile: 
Could you please push again that we can rerun CI 

git commit --allow-empty -m "trigger CI"

than push again should work without any changes :)

felixdittrich92:
Thanks :)

felixdittrich92 (approved):

felixdittrich92 (approved):

felixdittrich92:
@frgfm could you check this please i'm not able to merge this PR :thinking: (manual i get also an error: At least 1 approving review is required by reviewers with write access. :thinking: )

PR #1023, opened 2022-08-17
Feat/add language detection

aminemindee:
Hello this PR is to deal with the problem discussed here: #973  about the language detection part (It may need a rebase if the one about orientation passes before) 

- [x] Add language detection model
- [x] Add it to predictor and output result
- [x] Add tests


felixdittrich92:
Hi @aminemindee :wave:,
thanks for the PR :)

Could you maybe add a short latency benchmark with lang detection and without as comparison ?

I agree that it is a great feature (which should be inside doctr) but i see fasttext a bit troublesome have you tested : 
https://github.com/pemistahl/lingua-py ? Looks much better to me (requires only numpy and regex)  :)

About docker: `RuntimeError: Unsupported compiler -- at least C++11 support is needed!`

felixdittrich92:
> Hello @felixdittrich92 Thank you for your answers and feedback.
> 
> When you say a benchmark with using the language detection and not do you see in what shape ? A dummy test that will log the time it took and like expect with language detection to be longer ? or some other format ?
> 
> For the docker issue, you're right apparently fasttext is the one making it fail because it needs C++11. The `lingua-py` lib you suggested seems really good. I'll try it and replace fasttext with that +1

Hi @aminemindee, yeah i'm fine if you test it locally on your side and share the taken time with lang detection and without maybe for 2 - 3 test documents :)

Sounds good to me (while overflying the lib uses only N-Grams so it should also be much faster as fasttext) :+1:

aminemindee:
> > Hello @felixdittrich92 Thank you for your answers and feedback.
> > When you say a benchmark with using the language detection and not do you see in what shape ? A dummy test that will log the time it took and like expect with language detection to be longer ? or some other format ?
> > For the docker issue, you're right apparently fasttext is the one making it fail because it needs C++11. The `lingua-py` lib you suggested seems really good. I'll try it and replace fasttext with that +1
> 
> Hi @aminemindee, yeah i'm fine if you test it locally on your side and share the time it took with language detection and without maybe for 2 - 3 test documents :)
> 
> Sounds good to me (while overflying the lib uses only N-Grams so it should also be much faster as fasttext) +1

Hello @felixdittrich92, I did a benchmark using fasttext, langdetect and lingua-py using 3 pdfs with a total of 24 pages. 

| model  | timeit value | start - end time |
| --------- | --------------- | ------------------ |
| None   | 12.6 s ± 25.1 ms | 12.55 s |
| fasttext | 12.5 s ± 149 ms | 12.79 s |
| langdetect  | 12.3 s ± 96.5  | 12.57 s |
| lingua-py | 13.1 s ± 100 ms | 13.28 s|

It's weird that sometimes it's fast when using a language detection model. ( Probably it's too insignificant compared to the rest)

felixdittrich92:
Hi @aminemindee i think if we can choose it is enough to support language detection where we also support vocabs (See datasets/vocabs)

aminemindee:
> Hi @aminemindee i think if we can choose it is enough to support language detection where we also support vocabs (See datasets/vocabs)
> 
> Was lingua-py accure ?

Hello @felixdittrich92  if i understand correctly you suggest to only include the languages that are present in `doctr/datasets/vocabs.py` (english, french, portuguese, spanish, german, arabic, czech and vietnamese) right ? i'll run a benchmark with only these languages loaded

For the performance of lingua-py for the documents i tested it got the right answer with probabilities of 1.0

felixdittrich92:
> > Hi @aminemindee i think if we can choose it is enough to support language detection where we also support vocabs (See datasets/vocabs)
> > Was lingua-py accure ?
> 
> Hello @felixdittrich92 if i understand correctly you suggest to only include the languages that are present in `doctr/datasets/vocabs.py` (english, french, portuguese, spanish, german, arabic, czech and vietnamese) right ? i'll run a benchmark with only these languages loaded
> 
> For the performance of lingua-py for the documents i tested it got the right answer with probabilities of 1.0

@aminemindee yes that's what i mean :) prob sounds also good 😅 (we could also grab the vocabs via enum that it updates the language detection If we add a new vocab 👍 )

aminemindee:
> > > Hi @aminemindee i think if we can choose it is enough to support language detection where we also support vocabs (See datasets/vocabs)
> > > Was lingua-py accure ?
> > 
> > 
> > Hello @felixdittrich92 if i understand correctly you suggest to only include the languages that are present in `doctr/datasets/vocabs.py` (english, french, portuguese, spanish, german, arabic, czech and vietnamese) right ? i'll run a benchmark with only these languages loaded
> > For the performance of lingua-py for the documents i tested it got the right answer with probabilities of 1.0
> 
> @aminemindee yes that's what i mean :) prob sounds also good sweat_smile (we could also grab the vocabs via enum that it updates the language detection If we add a new vocab +1 )

Yep really good idea. I tried to do that you tell me if it's possible to do it maybe cleaner. 
I added also the params detect_orientation and detect_language to ocr_predictor function because it's the most used.

doctr/utils/lang_detect.py
```diff
@@ -0,0 +1,40 @@
...
+    Get the language with the highest probability or no language if only a few words or a high entropy
+    Args:
+        text (str): text
+    Returns:
+        The detected language in ISO 639 code and confidence score
+    """
+
+    model = LanguageDetectorBuilder.from_languages(*languages).build()
+    predictions = model.compute_language_confidence_values(text.lower())
+    if (
```
felixdittrich92, line 34: fits in one line :)

felixdittrich92 (changes requested):
Hi @aminemindee :wave:,

thanks for updating this :+1: 
I have added some minor comments :)

Would be great if you can add also the benchmark between lingua all Languages and only these which are available in vocabs  :100:

doctr/models/predictor/tensorflow.py
```diff
@@ -33,6 +34,8 @@ class OCRPredictor(NestedObject, _OCRPredictor):
             accordingly. Doing so will improve performances for documents with page-uniform rotations.
         detect_orientation: if True, the estimated general page orientation will be added to the predictions for each
             page. Doing so will slightly deteriorate the overall latency.
+        detect_language: if True, the language prediction will be added to the predictions for each
```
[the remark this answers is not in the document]
aminemindee: It can only detect the same language per page. So for a document with multiple pages it's possible to have multiple languages. you think we should add the "s" ? and should we add it also for orientation ?

doctr/utils/lang_detect.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (C) 2021-2022, Mindee.
```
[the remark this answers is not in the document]
aminemindee: Done and pushed.

doctr/utils/lang_detect.py
```diff
@@ -0,0 +1,40 @@
...
+language_dict = Language._member_map_
+languages = []
+for item in potential_languages:
+    language = language_dict.get(item, None)
+    if language:
+        languages.append(language)
+
+
+def get_language(text: str) -> Tuple[str, float]:
+    """Get languages of a textl using fasttext model.
```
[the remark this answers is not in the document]
aminemindee: Done and pushed.

doctr/models/predictor/tensorflow.py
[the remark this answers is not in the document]
felixdittrich92: ok than without "s" :) I have had in mind that we get probs for possible languages but if it is only one than without "s"

doctr/utils/lang_detect.py
```diff
@@ -0,0 +1,40 @@
...
+
+from typing import Tuple
+
+from lingua import Language, LanguageDetectorBuilder
+
+from doctr.datasets.vocabs import VOCABS
+
+__all__ = ['get_language']
+
+potential_languages = [item.upper() for item in VOCABS.keys()]
```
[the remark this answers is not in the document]
aminemindee: Yeah good idea juste need to update it if we update vocab but as it's in the same file should be okey. 
Done and pushed.

doctr/utils/lang_detect.py
aminemindee: Done and pushed.

aminemindee:
> Hi @aminemindee wave,
> 
> thanks for updating this +1 I have added some minor comments :)
> 
> Would be great if you can add also the benchmark between lingua all Languages and only these which are available in vocabs 100

For the benchmark :


model | timeit value | start - end time
-- | -- | --
None | 12.6 s ± 25.1 ms | 12.55 s
fasttext | 12.5 s ± 149 ms | 12.79 s
langdetect | 12.3 s ± 96.5 ms | 12.57 s
lingua-py | 13.1 s ± 100 ms | 13.28 s
lingua-py selected languages | 12.3 s ± 125 ms | 12.56

doctr/utils/lang_detect.py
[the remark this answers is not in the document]
felixdittrich92: yes test will raise if we forget to update LANGUAGES or VOCAB :+1:

felixdittrich92:
> > Hi @aminemindee wave,
> > thanks for updating this +1 I have added some minor comments :)
> > Would be great if you can add also the benchmark between lingua all Languages and only these which are available in vocabs 100
> 
> For the benchmark :
> 
> model	timeit value	start - end time
> None	12.6 s ± 25.1 ms	12.55 s
> fasttext	12.5 s ± 149 ms	12.79 s
> langdetect	12.3 s ± 96.5 ms	12.57 s
> lingua-py	13.1 s ± 100 ms	13.28 s
> lingua-py selected languages	12.3 s ± 125 ms	12.56

Mh .. I thought we could speed it up a bit more but it should be ok for now. :)

doctr/utils/lang_detect.py
```diff
@@ -0,0 +1,32 @@
...
+def get_language(text: str) -> Tuple[str, float]:
+    """Get languages of a text using lingua library.
+    Get the language with the highest probability or no language if only a few words or no language was identified
+    Args:
+        text (str): text
+    Returns:
+        The detected language in ISO 639 code and confidence score
+    """
+
+    model = LanguageDetectorBuilder.from_languages(*languages).build()
```
felixdittrich92, line 27: @aminemindee i think we should init the model outside of the function otherwise it will be initialized again with every call

felixdittrich92: i think this takes also some time maybe an overall speed up ?

tests/common/test_vocab_languages.py
```diff
@@ -0,0 +1,6 @@
+from doctr.datasets.vocabs import LANGUAGES, VOCABS
```
felixdittrich92, line 1: `test_dataset_vocabs.py`

doctr/utils/lang_detect.py
```diff
@@ -0,0 +1,32 @@
+# Copyright (C) 2022, Mindee.
```
felixdittrich92, line 1: @aminemindee I think we should move this to https://github.com/mindee/doctr/blob/main/doctr/models/_utils.py and move the corresponding test also

doctr/utils/lang_detect.py
aminemindee: Done and pushed. 
we gain 0.3 s :+1: 

model | timeit value | start - end time
-- | -- | --
None | 12.6 s ± 25.1 ms | 12.55 s
fasttext | 12.5 s ± 149 ms | 12.79 s
langdetect | 12.3 s ± 96.5 ms | 12.57 s
lingua-py | 13.1 s ± 100 ms | 13.28 s
lingua-py selected languages | 12 s ± 86.1 ms | 12.22

aminemindee: can you tell me where is the `with_low_accuracy_mode` i don't seem to find it.

felixdittrich92: https://github.com/pemistahl/lingua-py#95-low-accuracy-mode-versus-high-accuracy-mode

and

https://github.com/pemistahl/lingua-py/blob/305d001a331c01caa386e7a259c7ee9f98f99d4b/lingua/builder.py#L193

doctr/utils/lang_detect.py
aminemindee: Done and pushed.

doctr/utils/lang_detect.py
felixdittrich92: puh .. than we should test that `with_low_accuracy_mode` is enough

aminemindee: with `with_low_accuracy_mode` the sentence 'this is a test sentence.' get predicted as portuguese x) the RAM goes down to 0.3 gb. 

lang_detect use also 0.3 gb of RAM and it uses an AI based approach meaning the confidence score is more accurate as a probability and not a distance.

felixdittrich92: Mh i would say that 0.8 gb is to much .. so in fact of your tests do you think it is ok switching to lang_detect ? Let's try this and than let me know your decision i will go with it :)

aminemindee: From what i tested: 
- fasttext is the best one performance wise but you have to download model and it's the larger one. 
- langdetect is the easiest one to use but it's not really well done ( you have to do a try except because he return error if the text doesn't have any character he knows) 
- lingua seems really good but uses too much RAM and it's performance is not that good with the low accuracy mode. 

So for now i think langdetect is the reasonable choice, knowing we can change it easily after.

felixdittrich92 (approved):
@aminemindee thanks for all the updates:hugs:

PR #1030, opened 2022-08-25
[CI] replace header validation test with frgfm 's action

felixdittrich92:
This PR:

- replaces header validation test with @frgfm 's action (https://github.com/frgfm/validate-python-headers thanks for this :heart: )

related to: #957


doctr/models/factory/hub.py
```diff
@@ -1,7 +1,7 @@
 # Copyright (C) 2022, Mindee.
```
odulcy-mindee, line 1: Here, it's not starting by 2021. Should the action fail or should it only verify that the year is greater than `starting_year` defined in action ?

felixdittrich92: second case it need to be bigger so two ways are valid 2021-x (if file was created earlier) or 202x

odulcy-mindee: Ok thanks for clarification !

odulcy-mindee (approved):
Thanks @felixdittrich92 and @frgfm ! LGTM

felixdittrich92:
> Thanks @felixdittrich92 and @frgfm ! LGTM

Thanks goes to @frgfm :hugs:

frgfm:
Haha, glad it helped :)

PR #1031, opened 2022-08-25
[build] move requirements to pyproject.toml

felixdittrich92:
This PR:

- move requirements to pyproject.toml
- update CI
- update specifier

related to: #957 

api updates are borrowed from https://github.com/frgfm/Holocron/tree/main/api thanks @frgfm :sweat_smile:  have added comment in makefile :+1:


felixdittrich92:
NOTE: API is still broken but i think it makes sense to debug this in another PR @frgfm @aminemindee @odulcy-mindee @charlesmindee wdyt ?

felixdittrich92:
ok now we got it :sweat_smile:

.github/workflows/docker.yml
```diff
@@ -17,20 +17,36 @@ jobs:
...
+      - name: Install poetry
+        uses: abatilo/actions-poetry@v2.0.0
+        with:
+          poetry-version: 1.1.13
+      - name: Lock the requirements
+        run: |
+          cd api
+          poetry lock
+          poetry export -f requirements.txt --without-hashes --output requirements.txt
+          poetry export -f requirements.txt --without-hashes --dev --output requirements-dev.txt
```
odulcy-mindee, line 40: If possible, use `make lock` so it will also "test" that Makefile works and is sync with the CI

pyproject.toml
```diff
@@ -1,3 +1,134 @@
...
+    # docs
+    "sphinx>=3.0.0,!=3.5.0",
+    "sphinxemoji>=0.1.8",
+    "sphinx-copybutton>=0.3.1",
+    "docutils<0.18",
+    "recommonmark>=0.7.1",
+    "sphinx-markdown-tables>=0.0.15",
+    "sphinx-tabs>=3.3.0",
+    "furo>=2022.3.4",
+]
```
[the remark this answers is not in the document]
felixdittrich92: this does not work unfortunately (we have to keep the duplicates)

.github/workflows/style.yml
```diff
@@ -82,11 +82,11 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ~/.cache/pip
-          key: ${{ runner.os }}-pkg-deps-${{ matrix.python }}-${{ hashFiles('requirements.txt') }}
+          key: ${{ runner.os }}-pkg-deps-${{ matrix.python }}-${{ hashFiles('pyproject.toml') }}
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -e .[all] --upgrade
+          pip install -e .[dev] --upgrade
```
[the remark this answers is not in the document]
felixdittrich92: mypy needs all dependencies to check like numpy, cv, etc.

api/docker-compose.yml
```diff
@@ -4,9 +4,9 @@ services:
   web:
     build:
       context: .
-      dockerfile: Dockerfile-api
+      dockerfile: Dockerfile
     command: uvicorn app.main:app --reload --workers 1 --host 0.0.0.0 --port 8080
     volumes:
-      - ./api/:/usr/src/app/
+      - ./:/usr/src/app/
```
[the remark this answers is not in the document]
felixdittrich92: its only the common way where stuff is mounted in a docker nothing special :) -> some more information: https://stackoverflow.com/questions/54685942/dockerfile-for-ruby-app-why-is-workdir-specified-as-usr-src-app

.github/workflows/docker.yml
felixdittrich92: good catch :+1:

api/Dockerfile
```diff
@@ -8,27 +8,16 @@ ENV PYTHONUNBUFFERED 1
...
-COPY ./setup.py /tmp/setup.py
-COPY ./doctr /tmp/doctr
-
-RUN pip install -e /tmp/.[tf] \
-    && pip cache purge \
-    && rm -rf /root/.cache/pip
-
 # copy project
-COPY api /app
+COPY app /app/app
```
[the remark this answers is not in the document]
felixdittrich92: If we remove this we have to do all operations on the "root" folder which is not really pretty :)

.github/workflows/docker.yml
```diff
@@ -17,20 +17,36 @@ jobs:
...
-          docker-compose exec -T web python -m pip install --upgrade pip
-          docker-compose exec -T web pip install -r requirements-dev.txt
+        run: wget --spider --tries=12 http://localhost:8080/docs
       - name: Run docker test
-        run: docker-compose exec -T web pytest .
+        run: |
+          docker cp api/requirements-dev.txt api_web_1:/app/requirements-dev.txt
+          docker-compose -f api/docker-compose.yml exec -T web pip install -r requirements-dev.txt
+          docker cp api/tests api_web_1:/app/tests
+          docker-compose -f api/docker-compose.yml exec -T web pytest tests/
```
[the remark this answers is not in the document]
felixdittrich92: ok after some tests:
- lock works
- move only tests to a make ci-test : does not work no access
- same if we move the docker build inside ci-test

i would suggest to keep it :sweat_smile:

api/docker-compose.yml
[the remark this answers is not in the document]
odulcy-mindee: I mean the code is already inside the Docker container, no need to mount in `/usr/src/app/`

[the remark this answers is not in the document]
felixdittrich92: @odulcy-mindee now i got it :sweat_smile: sure done :+1:

pyproject.toml
```diff
@@ -1,3 +1,111 @@
...
+dev = [
+    "python-doctr[tf]",
+    "python-doctr[torch]",
+    "python-doctr[testing]",
+    "python-doctr[quality]",
+    "python-doctr[docs]"
+]
```
odulcy-mindee, line 95: nice ! :+1:

odulcy-mindee (approved):
LGTM ! We'll try to factorize the code between CI and Makefile another time, that's not that urgent

pyproject.toml
```diff
@@ -87,34 +87,11 @@ docs = [
...
-    "docutils<0.18",
-    "recommonmark>=0.7.1",
-    "sphinx-markdown-tables>=0.0.15",
-    "sphinx-tabs>=3.3.0",
-    "furo>=2022.3.4",
+    "python-doctr[tf]",
+    "python-doctr[torch]",
+    "python-doctr[testing]",
+    "python-doctr[quality]",
+    "python-doctr[docs]"
```
frgfm, line 94: That's dangerous because it will install the last published release. But in dev mode, you'll want to have the version of which the code is on your laptop

frgfm (commented):
Late review, but I think there is a slight problem with the "dev" extra

pyproject.toml
felixdittrich92: @frgfm that's a really good catch ... do you know an alternative way without rewriting all deps ?

frgfm: Nope that's why I rewrite all deps :sweat_smile:

frgfm: (this is not a dynamic build system yet, so that's not possible by design to the best of my knowledge)

felixdittrich92: would you like to open a PR for this ? :sweat_smile: Unfortunately im blocked until Wednesday (and the free time i have is currently filled with implementing ViT and ViTSTR in doctr :sweat_smile: )

PR #1033, opened 2022-08-29
Migrate static data from github to monitoring middleware.

marvinmindee:
Static data like models and images are now downloaded from a middleware and not from github releases, to enable more flexibility and monitor model downloads.


frgfm (commented):
Thanks @marvinmindee :) 

Quick question though:
- I understand clearly the advantage of monitoring the downloads of models (perhaps there is a feature on Github to monitor asset download but I'm not sure)
- but for the rest (images, etc.), what would be the advantage?

Thanks :pray:

marvinmindee:
> 
Hello, we also added he images so that all static data are accessed the same way and we don't have things dowloaded from multiple places (especially when we'll want to add other static files, it's better if we don't have to think about whether it should be on github/release or doctr-static). It also leaves flexibility for later!

felixdittrich92:
@frgfm @marvinmindee If we switch the models to doctr-static i agree that we should do this for all static data so we have all in one place :+1:

charlesmindee (approved):
Thanks, LGTM

PR #1034, opened 2022-08-30
[build] setup dependabot

felixdittrich92:
This PR:

- add config to setup dependabot

related to:
#957 
depends on:
#1031 

@frgfm @charlesmindee I need your help to update Settings / Security for dependabot (i have no rights to do this :sweat_smile:)


odulcy-mindee (approved):
:+1:

frgfm (commented):
Thanks! However I think we have to distinguish two things:
- security alerts (already setup I think)
- release alerts (each subdependency's) which can end up spamming us as soon as we have many deps

felixdittrich92:
@frgfm i wrote with @charlesmindee to do changes in Settings so i think this should be done (and it is limited to max 10 PRs) :)

PR #1038, opened 2022-09-01
[datasets] Fix recognition parts of SynthText and IMGUR5K

felixdittrich92:
This PR:

- fixes the mem issue for SynthText and IMGUR5K Recognition 
- now we write a new dataset for both datasets to disk (fastest way)

related to:
#935    Fix for Recognition: SynthText and IMGUR5K (not beauty but it works well - no mem issue and whole speed up)


charlesmindee (approved):
Thanks, LGTM

Issue #1001, opened 2022-07-29
[pdf] Switch PDF page rendering to an iterator format


frgfm:
As suggested by @mara004 in #1000, the PDF rendering is using a list comprehension which holds page processing while the rendering isn't complete:
https://github.com/mindee/doctr/blob/78c9105bb6c27c4a35ee9f601263da01668224b1/doctr/io/pdf.py#L45

This could be fixed by modifying it to yield the page and limit RAM usage:
```python
with pdfium.PdfDocument(file, password=password) as pdf:
    for img in pdf.render_topil(scale=scale, **kwargs): yield np.asarray(img)
```

felixdittrich92:
Clean up the minor issues a bit :sweat_smile:

felixdittrich92:
#1032

felixdittrich92:
Outdated by #1240

PR #1040, opened 2022-09-01
Switch PDF page rendering to an iterator format

felixdittrich92:
This PR:

- reduces RAM Usage while pdf page rendering

Closses:
#1001

### DIFF
--- a/doctr/io/pdf.py
+++ b/doctr/io/pdf.py
@@ -4,7 +4,7 @@
 # See LICENSE or go to <https://opensource.org/licenses/Apache-2.0> for full license details.
 
 from pathlib import Path
-from typing import Any, List, Optional
+from typing import Any, Iterator, Optional
 
 import numpy as np
 import pypdfium2 as pdfium
@@ -19,7 +19,7 @@ def read_pdf(
     scale: float = 2,
     password: Optional[str] = None,
     **kwargs: Any,
-) -> List[np.ndarray]:
+) -> Iterator[np.ndarray]:
     """Read a PDF file and convert it into an image in numpy format
 
     >>> from doctr.documents import read_pdf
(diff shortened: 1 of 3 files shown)


doctr/io/pdf.py
```diff
@@ -42,4 +42,5 @@ def read_pdf(
 
     # Rasterise pages to PIL images with pypdfium2 and convert to numpy ndarrays
     with pdfium.PdfDocument(file, password=password) as pdf:
-        return [np.asarray(img) for img in pdf.render_topil(scale=scale, **kwargs)]
+        for img in pdf.render_topil(scale=scale, **kwargs):
+            yield np.asarray(img)
```
mara004, line 46: May be a one-liner:
```python3
yield from pdf.render_topil(...)
```

mara004:
While this PR changes `read_pdf()` itself to be an iterator, you'd also have to change all uses to actually benefit from it.
Always converting to a list kind of defeats the purpose of having an iterator in the first place.

doctr/io/pdf.py
felixdittrich92: Mh looks a bit strange 😅 
`yield from np.asarray(pdf.render_topil(...))`

mara004: Right, I see! Overlooked that in the hurry. I expect it'll be better once I have released v3 so that you can use `render_tonumpy()`, see https://github.com/mindee/doctr/discussions/1032.

felixdittrich92: A got it would you link the discussion to the issue ? :) So i would close this PR until your release is finally

felixdittrich92:
> While this PR changes `read_pdf()` itself to be an iterator, you'd also have to change all uses to actually benefit from it. Always converting to a list kind of defeats the purpose of having an iterator in the first place.

Yeah that's correct... was a bit to fast with this

felixdittrich92:
Waiting until #1032

mara004:
No problem. I agree it's best to postpone this until I'm through with v3.

felixdittrich92:
> No problem. I agree it's best to postpone this until I'm through with v3.

Yeah no stress.. wrote with @charlesmindee today .. next release will be next month :)

mara004:
Good to know. I want to do a thorough reversion of my codebase so I don't need to make another API-breaking release too soon. But I do expect I'll finish it within a month at the most.

PR #1041, opened 2022-09-02
[Fix] unify recognition dataset parts return signature

felixdittrich92:
This PR:

- revert mistake for recognition dataset parts (img, label) instead of (img, dict)
- add target check in _read_sample
- now we have two cases in the library for datasets 
- more than one annotation type: targets has to be a dict with boxes / labels keys
- only one annotation type: target has to be str (label) or np.ndarray (boxes)

Hopefully that this is now a match to avoid mistakes for transformations

Closes:
#935

### DIFF
--- a/doctr/datasets/datasets/pytorch.py
+++ b/doctr/datasets/datasets/pytorch.py
@@ -20,6 +20,16 @@ __all__ = ["AbstractDataset", "VisionDataset"]
 class AbstractDataset(_AbstractDataset):
     def _read_sample(self, index: int) -> Tuple[torch.Tensor, Any]:
         img_name, target = self.data[index]
+
+        # Check target
+        if isinstance(target, dict):
+            assert "boxes" in target, "Target should contain 'boxes' key"
+            assert "labels" in target, "Target should contain 'labels' key"
+        else:
+            assert isinstance(target, str) or isinstance(
+                target, np.ndarray
+            ), "Target should be a string or a numpy array"
+
         # Read image
         img = (
             tensor_from_numpy(img_name, dtype=torch.float32)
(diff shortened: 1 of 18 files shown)


charlesmindee (approved):
thanks, LGTM

doctr/datasets/datasets/pytorch.py
```diff
@@ -20,6 +20,16 @@
...
         img_name, target = self.data[index]
+
+        # Check target
+        if isinstance(target, dict):
+            assert "boxes" in target, "Target should contain 'boxes' key"
+            assert "labels" in target, "Target should contain 'labels' key"
+        else:
+            assert isinstance(target, str) or isinstance(
+                target, np.ndarray
+            ), "Target should be a string or a numpy array"
```
frgfm, line 31: Considering that the target is unchanged since the constructor was run, that should be moved to the constructor (because otherwise it increases data reading latency)

doctr/datasets/datasets/tensorflow.py
```diff
@@ -20,6 +20,16 @@
...
+
+        # Check target
+        if isinstance(target, dict):
+            assert "boxes" in target, "Target should contain 'boxes' key"
+            assert "labels" in target, "Target should contain 'labels' key"
+        else:
+            assert isinstance(target, str) or isinstance(
+                target, np.ndarray
+            ), "Target should be a string or a numpy array"
+
```
frgfm, line 32: same here

frgfm (commented):
Thanks! Small comment though :)

doctr/datasets/datasets/pytorch.py
felixdittrich92: you mean to check it inside every dataset ? otherwise we have only access inside _read_sample and __getitem__ !? And i don't think it changes the runtime noticeably

PR #1047, opened 2022-09-05
[FIX] revert dev deps mistake

felixdittrich92:
This PR:

- reverts a mistake which would install dev build deps from last pip and not from main (thanks @frgfm :smile:)


frgfm (approved):
Thanks Felix!

felixdittrich92:
> Thanks Felix!

Thank you ! So we avoid a lot of confusion in the next release :sweat_smile:

Issue #513, opened 2021-09-29
Adding ViTSTR


felixdittrich92:
Adding Vision Transformer for scene text recognition i work currently on this (with huggingface ViT backbone) if i done and have solid results it would be a charme for me to add this model if you interested !? :) 
Same for the new unilm/TrOCR model

charlesmindee:
Hi @felixdittrich92, 

Thanks for your message, it would be a pleasure having you contributing to the lib! 

We already have a recognition model including a transformer decoder ([MASTER](https://arxiv.org/pdf/1910.02562.pdf)), but we do not have yet full transformer architectures such as ViT or TrOCR. It is on the mid-term road map, and if you would like to propose your implementation you are more than welcome to open a PR! :pray: 

Please read the [CONTRIBUTING](https://github.com/mindee/doctr/blob/main/CONTRIBUTING.md) section and feel free to look at the models already implemented in doctr :smiley:.

Thank you and have a nice day :+1:

felixdittrich92:
i will do thanks :) :+1:

charlesmindee:
Hi @felixdittrich92, do you still plan to implement this ? If not, we may close this issue to avoid a huge stack of unaddressed  ones!

felixdittrich92:
Huhu @charlesmindee :wave: ,
yes of course (maybe a bit lighter version with mobilevit) but i think ftm there are other thinks like a fix for master and sar are more important so i would say lets hold this on 1.0.0 wdyt ? :+1:

charlesmindee:
ok

chpatrick:
@felixdittrich92 Hi, are there any model weights available for ViTSTR that are compatible with doctr? :)

I saw these ones but they seem to be named differently I suppose: https://github.com/roatienza/deep-text-recognition-benchmark/releases

PR #1048, opened 2022-09-06
[DRAFT] [models] add ViTSTR in TF and PT

felixdittrich92:
This PR:

- adds VisionTransformer as module in models/modules/vision_transformer (TF/PT) 
- add Vitstr head (TF / PT)

TODOS: 

- [ ] toy run / check that all works fine
- [ ] check code quality

Any feedback is welcome :)

@frgfm wdyt ? I think it is a more flexible way instead of adding ViT as a fixed classification model (where we run into trouble with different sizes for patch_embeds) - can easily be extended needs only a head :sweat_smile: 

related to: #513 #1003


felixdittrich92:
NOTE: 
- ViT works fine (also tested with timm's ViT implementation)
- loss does not decrease (needs debugging)

felixdittrich92:
@frgfm before i continue to debug this let me know your ideas you have in mind :hugs: 
Go this way to keep ViT as module or maybe start to implement as classification model and then think how we can make the patch_embedding flexible ? wdyt ?

doctr/models/modules/transformer/pytorch.py
```diff
@@ -57,10 +57,10 @@ def scaled_dot_product_attention(
 class PositionwiseFeedForward(nn.Sequential):
     """Position-wise Feed-Forward Network"""
 
-    def __init__(self, d_model: int, ffd: int, dropout: float = 0.1) -> None:
+    def __init__(self, d_model: int, ffd: int, dropout: float = 0.1, use_gelu: bool = False) -> None:
```
frgfm, line 60: boolean value for activation selection is rather limited: are we positive that only relu & gelu can be used for such architecture types?

doctr/models/modules/transformer/tensorflow.py
```diff
@@ -74,14 +74,19 @@ def scaled_dot_product_attention(
 class PositionwiseFeedForward(layers.Layer, NestedObject):
     """Position-wise Feed-Forward Network"""
 
-    def __init__(self, d_model: int, ffd: int, dropout=0.1) -> None:
+    def __init__(self, d_model: int, ffd: int, dropout=0.1, use_gelu: bool = False) -> None:
         super(PositionwiseFeedForward, self).__init__()
+        self.use_gelu = use_gelu
```
frgfm, line 79: instantiate a `self.activation_fn` in the constructor to avoid conditional execution in the call :+1:

doctr/models/modules/vision_transformer/pytorch.py
```diff
@@ -0,0 +1,100 @@
...
+
+from ..transformer.pytorch import MultiHeadAttention, PositionwiseFeedForward
+
+__all__ = ["VisionTransformer"]
+
+
+class PatchEmbedding(nn.Module):
+    """Compute 2D patch embedding"""
+
+    # Inpired by: https://github.com/rwightman/pytorch-image-models/blob/master/timm/models/layers/patch_embed.py
```
frgfm, line 19: Just FYI: can you confirm that you made a lot of modifications?
inspired by is rather light
borrowed from is more significant

doctr/models/modules/vision_transformer/pytorch.py
```diff
@@ -0,0 +1,100 @@
...
+        assert H % self.patch_size[0] == 0, "Image height must be divisible by patch height"
+        assert W % self.patch_size[1] == 0, "Image width must be divisible by patch width"
+
+        return self.proj(x)  # BCHW
+
+
+class VisionTransformer(nn.Module):
+    """VisionTransformer architecture as described in
+    `"An Image is Worth 16x16 Words: Transformers for Image Recognition at Scale",
+    <https://arxiv.org/pdf/2010.11929.pdf>`_."""
```
frgfm, line 47: let's specify the constructor args

doctr/models/modules/vision_transformer/tensorflow.py
```diff
@@ -0,0 +1,110 @@
...
+        assert H % self.patch_size[0] == 0, "Image height must be divisible by patch height"
+        assert W % self.patch_size[1] == 0, "Image width must be divisible by patch width"
+
+        return self.proj(x, **kwargs)  # BHWC
+
+
+class VisionTransformer(layers.Layer, NestedObject):
+    """VisionTransformer architecture as described in
+    `"An Image is Worth 16x16 Words: Transformers for Image Recognition at Scale",
+    <https://arxiv.org/pdf/2010.11929.pdf>`_."""
```
frgfm, line 54: same here

frgfm (commented):
Thanks Felix! I added some comments :)

felixdittrich92:
@frgfm haha i do currently focus to implement it as classification model :sweat_smile: we have to do a decision !

felixdittrich92:
outdated by: #1050

PR #1053, opened 2022-09-08
Onnx reorder

ryanchesler:


felixdittrich92:
Hey @ryanchesler :wave:, i saw you are interested to use docTR models with Onnxruntime / Openvino :)

Currently all models can be exported to Onnx (some TF models needs a fixed batch size) up to the logits (only mobilenet in TF does not work will be fixed after 0.5.2 release) 

What we have to do is building an standalone project build which excludes TF and PT then we have to build a pipe only for onnx similar to the existing one (models needs something like a config file, that we can decide which PostProcessor to use / vocab , etc.) the currently PostProcessors implementations should be reusable ( i think it is not possible to integrate the post processing in the onnx model - a lot of opencv, shapely, ..)

If you are interessted to work on this a first PoC would be super useful where we can step into :) @frgfm

PR #1060, opened 2022-09-15
[refactor] detection script

felixdittrich92:
This PR:

- refactored detection script
- add ability to save also as xml
- add minor test

### DIFF
--- a/scripts/detect_text.py
+++ b/scripts/detect_text.py
@@ -24,54 +24,55 @@ if is_tf_available():
     if any(gpu_devices):
         tf.config.experimental.set_memory_growth(gpu_devices[0], True)
 
+IMAGE_FILE_EXTENSIONS = [".jpeg", ".jpg", ".png", ".tif", ".tiff", ".bmp"]
+OTHER_EXTENSIONS = [".pdf"]
 
-_OUTPUT_CHOICE_JSON = "json"
-_OUTPUT_CHOICE_TEXT = "txt"
 
+def _process_file(model, file_path: Path, out_format: str) -> None:
 
-def _process_file(model, file_path: Path, out_format: str) -> str:
-    if str(file_path).lower().endswith(".pdf"):
+    if out_format not in ["txt", "json", "xml"]:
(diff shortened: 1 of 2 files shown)


scripts/detect_text.py
```diff
@@ -24,95 +24,71 @@
...
+    if out_format == "json":
+        output = json.dumps(out.export(), indent=2)
+    elif out_format == "txt":
+        output = out.render()
+    elif out_format == "xml":
+        output = out.export_as_xml()
+
+    path = Path("output").joinpath(file_path.stem + "." + out_format)
+    if out_format == "xml":
+        for i, (xml_bytes, xml_tree) in enumerate(output):
```
ianardee, line 54: getting a warning here about an unused variable

scripts/detect_text.py
```diff
@@ -24,95 +24,71 @@
...
 
 
 if __name__ == "__main__":
     parsed_args = parse_args()
-    try:
-        main(parsed_args)
-    except KeyboardInterrupt:
-        print("Cancelled")
-        pass
+    main(parsed_args)
```
ianardee, line 94: why take out the keyboard interrupt?

ianardee (approved):

scripts/detect_text.py
felixdittrich92: Hi 👋 yeah we could replace it with _ but so it is more clear what it returns :)

scripts/detect_text.py
ianardee: yes ok without

scripts/detect_text.py
felixdittrich92: The XML export returns a list which contains a tuple (XML as bytes, the whole ElementTree) for each page :)

ianardee: ok clear

thanks a lot for the cleanup work on the PR :-)

felixdittrich92: No problem 🤗

Issue #991, opened 2022-07-20
Turn off automatically download backbone model


bavo96:
Hi team,

Why don't we use the `pretrained` flag instead of `pretrained_backbone` flag for using pretrained model in `backbone_fn` ? I have turned off the flag `pretrained` by passing the flag to the `ocr_predictor` but it keeps downloading the pretrained model for backbone and I have to modify the code manually. Can we combine those flags into one or pass the `pretrained_backbone` outside this function? 
![image](https://user-images.githubusercontent.com/27908949/179979379-6b6a65c0-73a9-42bf-8679-c889a38ec3ea.png)

Many thanks.

frgfm:
Hello @bavo96 :wave: 

First, quick suggestions: 
- this is more of a question topic than bug reporting, so [GH discussions](https://github.com/mindee/doctr/discussions/new) are a better fit for this :)
- especially when sharing code, send the url with a reference or copy paste & format the code, but avoid screenshots :pray: 

Now, about your question:
- first off, you can pass `pretrained_backbone=False` to the model directly (though it's true that from the recognition_predictor it's not possible as is (feel free to open a PR if you want :))
- the reason we didn't consider this is that when not using a pretrained model, you are most likely training something if you're doing anything at all. In that case, having part of the network already trained on a relevant data distribution is useful. You can freeze the backbone and train only the later parts of the model to check results quickly :+1: 

I hope this helps, cheers :v:

bavo96:
Hi @frgfm, so sorry about my posting, I'm still new to this repo and don't know where to post the question :( I'll post the reference to the code next time ❤️ Thanks for your suggestions. 
I don't mind if we use the pretrained backbone when training a model. But in reference (testing) phase, I want to load my trained model (by using `load_state_dict`) instead of downloading the pretrained model again. So in this case, the code will download a pretrained model (when creating a `doctr.models.ocr_predictor` object), and then I load my trained model (`reco_predictor.load_state_dict`), which is redundant and can be done by loading my trained model only. What do you think about this situation?

frgfm:
Hi @bavo96 :wave: 

Oh don't misinterpret my previous message, there is no problem at all :) I prefer to share feedback so that you get more familiar with the OSS community :grin: 

> in reference (testing) phase, I want to load my trained model (by using load_state_dict) instead of downloading the pretrained model again

Ok, so if I get this right, you'd like to have no download whatsoever and load your local model checkpoint directly, correct?

bavo96:
[...]

PR #1061, opened 2022-09-15
make pretrained backbone flexible in predictor

felixdittrich92:
This PR:

- solves : #991 

### DIFF
--- a/doctr/models/detection/zoo.py
+++ b/doctr/models/detection/zoo.py
@@ -51,13 +51,14 @@ def _predictor(arch: Any, pretrained: bool, assume_straight_pages: bool = True, 
             pretrained_backbone=kwargs.get("pretrained_backbone", True),
             assume_straight_pages=assume_straight_pages,
         )
-        kwargs.pop("pretrained_backbone", None)
     else:
         if not isinstance(arch, (detection.DBNet, detection.LinkNet)):
             raise ValueError(f"unknown architecture: {type(arch)}")
 
         _model = arch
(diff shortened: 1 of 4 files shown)


doctr/models/detection/zoo.py
```diff
@@ -46,14 +46,20 @@ def _predictor(arch: Any, pretrained: bool, assume_straight_pages: bool = True,
...
+            assume_straight_pages=assume_straight_pages,
+        )
     else:
         if not isinstance(arch, (detection.DBNet, detection.LinkNet)):
             raise ValueError(f"unknown architecture: {type(arch)}")
 
         _model = arch
         _model.assume_straight_pages = assume_straight_pages
 
+    kwargs.pop("pretrained_backbone", None)
```
frgfm, line 61: no need for the second arg since we don't capture the output

frgfm (approved):
Thanks Felix! A slight comment but that's quite alright

doctr/models/detection/zoo.py
felixdittrich92: mh .. i have tried this but without success.
However, I don't remember what exactly the error was :sweat_smile:

felixdittrich92:
will be merged after #1063 to avoid failing CI

doctr/models/detection/zoo.py
frgfm: oh perhaps it fails if the key doesn't exist without the default value

PR #1072, opened 2022-09-21
[models] ViT add checkpoints and some rework to use pretrained ViT backbone in ViTSTR

felixdittrich92:
This PR: 

- make pretrained checkpoints available for ViT (TF and PT)
- rework patchify make it dynamic to the input shape
- rework pretrained backbone usage for ViTSTR with checkpoints
- fix issue with bicubic scale and onnxruntime (use bilinear : minor cost on quality and speed up)

All added checkpoints >98% acc

Any feedback is welcome

PR #1073, opened 2022-09-26
update version for minor release

felixdittrich92:
This PR:

- update version to 0.6.0a0 for minor release next week
- release tracker will be updated today - done :)

:hugs: 

CI will pass after merge #1072

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ from pathlib import Path
 from setuptools import setup
 
 PKG_NAME = "python-doctr"
-VERSION = os.getenv("BUILD_VERSION", "0.5.2a0")
+VERSION = os.getenv("BUILD_VERSION", "0.6.0a0")
 
 
 if __name__ == "__main__":


setup.py
```diff
@@ -9,7 +9,7 @@
 from setuptools import setup
 
 PKG_NAME = "python-doctr"
-VERSION = os.getenv("BUILD_VERSION", "0.5.2a0")
+VERSION = os.getenv("BUILD_VERSION", "0.6.0a0")
```
frgfm, line 12: Unless you intend to release 0.6.0, I suggest doing this after the actual release (otherwise it will mess up dev installations ;))

frgfm (commented):
I suggest holding this until the actual release ;)

setup.py
felixdittrich92: ok :+1: i have updated the release tracker moved all from 0.5.2 in 0.6.0 and open stuff to 0.7.0 could you take a look ?

felixdittrich92: let's move after the release to 0.7.0a0 (the current state seems to be clean (mostly) clean related to bugs so i think it is ok)

felixdittrich92:
@odulcy-mindee ok last one that we can release it as 0.6.0 and set it to 0.7.0a= with the post release PR :)

setup.py
felixdittrich92: @frgfm @odulcy-mindee a short summary why i think it is more a minor release instead a patch release:
(release tracker 0.5.2 is merged in 0.6.0 tracker and updated)

**new features:**

- all datasets can now be used also for recognition task (inclusive documentation)
- full integration with HF hub (pushing and loading inclusive documentation)
- updated documentation
- updated project build structure
- integrated language detection
- orientation can now also added to the output
- demo is now also PyTorch compatible
- new classification model ViT (TF and PT)
- new recognition model ViTSTR (TF and PT)
- (experimential - feature which is not completly done) all models can be exported to ONNX up to the logits (only TF mobilenet left for 0.7.0)

**bug fixes:**

- MASTER (TF and PT) refactored & fixed
- SAR (TF and PT) refactored & fixed
- some minor fixes and improvements (easier custom model loading, and so on)

odulcy-mindee: > Unless you intend to release 0.6.0, I suggest doing this after the actual release (otherwise it will mess up dev installations ;))
I'm not sure to understand. What would be the problem ?

felixdittrich92: @frgfm have had in mind that we want to release 0.5.2 so it would be correct to do it after the release but we want to release 0.6.0 (minor release insted patch) so after releasing we set the version to 0.7.0 which will be the next minor release

felixdittrich92: so it is correct to increase the number to 0.6.0 before we release and afterwards to 0.7.0

odulcy-mindee: So we skip 0.5.2 in profit of 0.6.0 ?

felixdittrich92: yep all issues from 0.5.2 tracker are included in 0.6.0 now + new features which was planned for 0.6.0 and are done

odulcy-mindee: Okey ! That's fine for me

felixdittrich92: normal flow should be 0.1.0 (features / minor) -> 0.1.1 (bug fixes patch) -> 0.2.0 -> .... (0.5.2 was only planned for the architecture fixes from MASTER (PT + TF) and SAR (PT + TF) and are included in 0.6.0)

odulcy-mindee (approved):
:rocket: :rocket: :rocket:

PR #1075, opened 2022-09-26
upgrade pypdfium2

felixdittrich92:
This PR:

- short update to pypdfium >= 3.0.0


felixdittrich92:
@mara004

doctr/io/pdf.py
```diff
@@ -42,4 +42,4 @@ def read_pdf(
 
     # Rasterise pages to PIL images with pypdfium2 and convert to numpy ndarrays
     with pdfium.PdfDocument(file, password=password) as pdf:
-        return [np.asarray(img) for img in pdf.render_topil(scale=scale, **kwargs)]
+        return [img[0] for img in iter(pdf.render_to(pdfium.BitmapConv.numpy_ndarray, scale=scale, **kwargs))]
```
mara004, line 45: Is there a specific reason why you added `iter()`?

doctr/io/pdf.py
```diff
@@ -42,4 +42,4 @@ def read_pdf(
 
     # Rasterise pages to PIL images with pypdfium2 and convert to numpy ndarrays
     with pdfium.PdfDocument(file, password=password) as pdf:
-        return [np.asarray(img) for img in pdf.render_topil(scale=scale, **kwargs)]
+        return [img[0] for img in pdf.render_to(pdfium.BitmapConv.numpy_ndarray, scale=scale, **kwargs)]
```
mara004, line 45: I find the name `img` a bit confusing, because the variable doesn't really store the image, but the result, which is a tuple of the form (image, colour_format). What about renaming it to `result` or doing `img for img, _ in ...`?

doctr/io/pdf.py
mara004: Nice wording!

doctr/io/pdf.py
felixdittrich92: sec case makes sense :)

mara004:
LGTM

felixdittrich92:
Ah damn i forgot to reverse the order (BGR to RGB) @mara004 there was an arg for it if i have it correctly in mind ?

mara004:
It's `rev_byteorder=True`

pyproject.toml
```diff
@@ -37,7 +37,7 @@ dependencies = [
     "scipy>=1.4.0,<2.0.0",
     "h5py>=3.1.0,<4.0.0",
     "opencv-python>=4.5.0,<5.0.0",
-    "pypdfium2>=2.1.0,<3.0.0",  # cf. https://github.com/mindee/doctr/issues/947
+    "pypdfium2>=3.0.0,<4.0.0",
```
odulcy-mindee, line 40: Maybe we can pin:
```
pypdfium2~=3.1.0
```

What do you think @felixdittrich92 ? Do you prefer less constraints in `pyproject` as it has been done with other dependencies ?

doctr/io/pdf.py
```diff
@@ -28,6 +29,7 @@ def read_pdf(
     Args:
         file: the path to the PDF file
         scale: rendering scale (1 corresponds to 72dpi)
+        rgb_mode: if True, the output will be RGB, otherwise BGR
         password: a password to unlock the document, if encrypted
         kwargs: additional parameters to :meth:`pypdfium2.PdfDocument.render_topil`
```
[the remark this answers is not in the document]
felixdittrich92: good catch i missed this :+1:

pyproject.toml
felixdittrich92: @odulcy-mindee i think this is fine as it is :) but to be sure :
@mara004 in future releases with breaking changes you increase the major version number correct ?

doctr/io/pdf.py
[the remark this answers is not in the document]
mara004: Right, thanks! I only looked at the diff and this wasn't part of it...

pyproject.toml
felixdittrich92: the only thing we want to avoid are breaking changes which leads to issues in doctr so keeping a bigger range of minor and patch updates should be ok :)

mara004: Yes. Maybe I might change some really insignificant parameter without a major update if there's a problem with it, but document opening and rendering methods will remain the same within one major version.

odulcy-mindee: Okey, it's fine for me, thanks for the quick reply @mara004 !

odulcy-mindee (approved):

doctr/io/pdf.py
```diff
@@ -28,8 +29,9 @@ def read_pdf(
     Args:
         file: the path to the PDF file
         scale: rendering scale (1 corresponds to 72dpi)
+        rgb_mode: if True, the output will be RGB, otherwise BGR
```
mara004, line 32: Can doctr work with BGR input, too?

felixdittrich92: working yes but it would not be correct :sweat_smile: but we need to keep it flexible for users which to anything else with the loaded file

felixdittrich92: Example:
x = DocumentFile.from_pdf(..)
do with the numpy arrays what you want to do :)

mara004: Yep, I see

PR #1080, opened 2022-09-29
ViTSTR disable pretrained backbone by default

felixdittrich92:
This PR:

- disable using pretrained backbone in VITSTR by default
- after some tests it runs much better without using the pretrained backbone


charlesmindee (approved):
thanks

PR #1081, opened 2022-09-29
chore: apply post release modifications v0.6.0

felixdittrich92:
This PR:

- add post release modifications
- set dev to 0.7.0


charlesmindee (approved):
thanks!

odulcy-mindee (approved):

setup.py
```diff
@@ -9,7 +9,7 @@
 from setuptools import setup
 
 PKG_NAME = "python-doctr"
-VERSION = os.getenv("BUILD_VERSION", "0.6.0a0")
+VERSION = os.getenv("BUILD_VERSION", "0.7.0a0")
```
frgfm, line 12: 0.6.1a0 out of safety ;)

frgfm (commented):
I think we shouldn't stress ourselves by setting a version index too far. If we release X.Y.0, the dev version should be something like X.Y.1a0 (otherwise if we release a patch, that will mess up some dev installations for a few people)

setup.py
felixdittrich92: yeah i set it to 0.7.0 because we have no backlog. But you are right :)

PR #1097, opened 2022-10-13
Feat: Make detection training and inference Multiclass

aminemindee:
Hello, this PR is to make the detection part of DocTR multiclass.

:warning: This PR has a breaking change on the use of DocTR. The attribute `blocks` of class `Page` is no longer a list of blocks but a dictionnary with keys being the class names. 

- [x] Training detection model tf/torch working with new data format to include class and also old data format
- [x] Inference with ocr_predictor working tf/torch
- [x] visulization and export
- [x] middleware and hf api not changed should work without changes
- [x] minimal API: some changes in the docker file and doc to make it work 

A first review of the code was done on my fork. you can find it here: https://github.com/aminemindee/doctr/pull/4 

Any feedback is welcome!

### DIFF
--- a/api/app/routes/kie.py
+++ b/api/app/routes/kie.py
@@ -1,0 +1,31 @@
+# Copyright (C) 2021-2022, Mindee.
+
+# This program is licensed under the Apache License 2.0.
+# See LICENSE or go to <https://opensource.org/licenses/Apache-2.0> for full license details.
+
+from typing import Dict, List
+
+from fastapi import APIRouter, File, UploadFile, status
+
+from app.schemas import OCROut
+from app.vision import kie_predictor
+from doctr.io import decode_img_as_tensor
+
+router = APIRouter()
+
+
+@router.post("/", response_model=Dict[str, List[OCROut]], status_code=status.HTTP_200_OK, summary="Perform KIE")
+async def perform_kie(file: UploadFile = File(...)):
+    """Runs docTR KIE model to analyze the input image"""
+    img = decode_img_as_tensor(file.file.read())
+    out = kie_predictor([img])
+
+    return {
+        class_name: [
+            OCROut(box=(*word.geometry[0], *word.geometry[1]), value=word.value)
+            for block in out.pages[0].predictions[class_name]
+            for line in block.lines
+            for word in line.words
+        ]
+        for class_name in out.pages[0].predictions.keys()
+    }
--- a/api/app/routes/kie.py
+++ b/api/app/routes/kie.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2021-2022, Mindee.
+# Copyright (C) 2022, Mindee.
 
 # This program is licensed under the Apache License 2.0.
 # See LICENSE or go to <https://opensource.org/licenses/Apache-2.0> for full license details.
--- a/api/app/routes/kie.py
+++ b/api/app/routes/kie.py
@@ -22,10 +22,8 @@ async def perform_kie(file: UploadFile = File(...)):
 
     return {
         class_name: [
-            OCROut(box=(*word.geometry[0], *word.geometry[1]), value=word.value)
-            for block in out.pages[0].predictions[class_name]
-            for line in block.lines
-            for word in line.words
+            OCROut(box=(*prediction.geometry[0], *prediction.geometry[1]), value=prediction.value)
+            for prediction in out.pages[0].predictions[class_name]
         ]
         for class_name in out.pages[0].predictions.keys()
     }
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -10,7 +10,7 @@ authors = ["Mindee <contact@mindee.com>"]
 license = "Apache-2.0"
 
 [tool.poetry.dependencies]
-python = ">=3.8.2,<3.11"
+python = ">=3.8.2,<3.11"  # pypdfium2 needs a python version above 3.8.2
 tensorflow = ">=2.9.0,<3.0.0"
 tensorflow-addons = ">=0.17.1"
 python-doctr = { version = ">=1.0.0", extras = ['tf'] }
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -10,10 +10,10 @@ authors = ["Mindee <contact@mindee.com>"]
 license = "Apache-2.0"
 
 [tool.poetry.dependencies]
-python = ">=3.8,<3.11"
+python = ">=3.8.2,<3.11"
 tensorflow = ">=2.9.0,<3.0.0"
 tensorflow-addons = ">=0.17.1"
-python-doctr = ">=0.2.0"
+python-doctr = { version = ">=0.6.0", extras = ['tf'] }
 # Fastapi: minimum version required to avoid pydantic error
 # cf. https://github.com/tiangolo/fastapi/issues/4168
 fastapi = ">=0.73.0"
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.masonry.api"
 
 [tool.poetry]
 name = "doctr-api"
-version = "0.5.2a0"
+version = "1.0.1a0"
 description = "Backend template for your OCR API with docTR"
 authors = ["Mindee <contact@mindee.com>"]
 license = "Apache-2.0"
@@ -13,7 +13,7 @@ license = "Apache-2.0"
 python = ">=3.8.2,<3.11"
 tensorflow = ">=2.9.0,<3.0.0"
 tensorflow-addons = ">=0.17.1"
-python-doctr = { version = ">=0.6.0", extras = ['tf'] }
+python-doctr = { version = ">=1.0.0", extras = ['tf'] }
 # Fastapi: minimum version required to avoid pydantic error
 # cf. https://github.com/tiangolo/fastapi/issues/4168
 fastapi = ">=0.73.0"
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -13,7 +13,7 @@ license = "Apache-2.0"
 python = ">=3.8.2,<3.11"  # pypdfium2 needs a python version above 3.8.2
 tensorflow = ">=2.9.0,<3.0.0"
 tensorflow-addons = ">=0.17.1"
-python-doctr = { version = ">=1.0.0", extras = ['tf'] }
+python-doctr = { version = ">=0.7.0", extras = ['tf'] }
 # Fastapi: minimum version required to avoid pydantic error
 # cf. https://github.com/tiangolo/fastapi/issues/4168
 fastapi = ">=0.73.0"
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.masonry.api"
 
 [tool.poetry]
 name = "doctr-api"
-version = "1.0.1a0"
+version = "0.7.1a0"
 description = "Backend template for your OCR API with docTR"
 authors = ["Mindee <contact@mindee.com>"]
 license = "Apache-2.0"
--- a/doctr/datasets/datasets/base.py
+++ b/doctr/datasets/datasets/base.py
@@ -8,8 +8,6 @@ import shutil
 from pathlib import Path
 from typing import Any, Callable, List, Optional, Tuple, Union
 
-import numpy as np
-
 from doctr.io.image import get_img_shape
 from doctr.utils.data import download_from_url
 
@@ -57,11 +55,11 @@ class _AbstractDataset:
             img = self.img_transforms(img)
 
         if self.sample_transforms is not None:
-            if isinstance(target, np.ndarray):
-                img, target = self.sample_transforms(img, target)
-            elif isinstance(target, dict):
+            if isinstance(target, dict):
                 for k, v in target.items():
                     img, target[k] = self.sample_transforms(img, v)
+            else:
+                img, target = self.sample_transforms(img, target)
 
         return img, target
 
--- a/doctr/datasets/datasets/base.py
+++ b/doctr/datasets/datasets/base.py
@@ -8,6 +8,8 @@ import shutil
 from pathlib import Path
 from typing import Any, Callable, List, Optional, Tuple, Union
 
+import numpy as np
+
 from doctr.io.image import get_img_shape
 from doctr.utils.data import download_from_url
 
@@ -55,7 +57,11 @@ class _AbstractDataset:
             img = self.img_transforms(img)
 
         if self.sample_transforms is not None:
-            img, target = self.sample_transforms(img, target)
+            if isinstance(target, np.ndarray):
+                img, target = self.sample_transforms(img, target)
+            elif isinstance(target, dict):
+                for k, v in target.items():
+                    img, target[k] = self.sample_transforms(img, v)
 
         return img, target
 
--- a/doctr/datasets/datasets/base.py
+++ b/doctr/datasets/datasets/base.py
@@ -8,6 +8,8 @@ import shutil
 from pathlib import Path
 from typing import Any, Callable, List, Optional, Tuple, Union
 
+import numpy as np
+
 from doctr.file_utils import copy_tensor
 from doctr.io.image import get_img_shape
 from doctr.utils.data import download_from_url
@@ -56,7 +58,7 @@ class _AbstractDataset:
             img = self.img_transforms(img)
 
         if self.sample_transforms is not None:
-            if isinstance(target, dict):
+            if isinstance(target, dict) and all([isinstance(item, np.ndarray) for item in target.values()]):
                 img_transformed = copy_tensor(img)
                 for class_name, bboxes in target.items():
                     img_transformed, target[class_name] = self.sample_transforms(img, bboxes)
--- a/doctr/datasets/datasets/base.py
+++ b/doctr/datasets/datasets/base.py
@@ -56,8 +56,10 @@ class _AbstractDataset:
 
         if self.sample_transforms is not None:
             if isinstance(target, dict):
+                img_transformed = img.copy()
                 for class_name, bboxes in target.items():
-                    img, target[class_name] = self.sample_transforms(img, bboxes)
+                    img_transformed, target[class_name] = self.sample_transforms(img, bboxes)
+                img = img_transformed
             else:
                 img, target = self.sample_transforms(img, target)
 
--- a/doctr/datasets/datasets/base.py
+++ b/doctr/datasets/datasets/base.py
@@ -56,8 +56,8 @@ class _AbstractDataset:
 
         if self.sample_transforms is not None:
             if isinstance(target, dict):
-                for k, v in target.items():
-                    img, target[k] = self.sample_transforms(img, v)
+                for class_name, bboxes in target.items():
+                    img, target[class_name] = self.sample_transforms(img, bboxes)
             else:
                 img, target = self.sample_transforms(img, target)
 
--- a/doctr/datasets/datasets/base.py
+++ b/doctr/datasets/datasets/base.py
@@ -8,6 +8,7 @@ import shutil
 from pathlib import Path
 from typing import Any, Callable, List, Optional, Tuple, Union
 
+from doctr.file_utils import copy_tensor
 from doctr.io.image import get_img_shape
 from doctr.utils.data import download_from_url
 
@@ -56,7 +57,7 @@ class _AbstractDataset:
 
         if self.sample_transforms is not None:
             if isinstance(target, dict):
-                img_transformed = img.copy()
+                img_transformed = copy_tensor(img)
                 for class_name, bboxes in target.items():
                     img_transformed, target[class_name] = self.sample_transforms(img, bboxes)
                 img = img_transformed
--- a/doctr/datasets/datasets/pytorch.py
+++ b/doctr/datasets/datasets/pytorch.py
@@ -25,6 +25,10 @@ class AbstractDataset(_AbstractDataset):
         if isinstance(target, dict):
             assert "boxes" in target, "Target should contain 'boxes' key"
             assert "labels" in target, "Target should contain 'labels' key"
+        elif isinstance(target, tuple):
+            assert isinstance(target[0], str) or isinstance(
+                target[0], np.ndarray
+            ), "Target should be a string or a numpy array"
         else:
             assert isinstance(target, str) or isinstance(
                 target, np.ndarray
--- a/doctr/datasets/datasets/pytorch.py
+++ b/doctr/datasets/datasets/pytorch.py
@@ -26,6 +26,7 @@ class AbstractDataset(_AbstractDataset):
             assert "boxes" in target, "Target should contain 'boxes' key"
             assert "labels" in target, "Target should contain 'labels' key"
         elif isinstance(target, tuple):
+            assert len(target) == 2
             assert isinstance(target[0], str) or isinstance(
                 target[0], np.ndarray
             ), "first element of the tuple should be a string or a numpy array"
--- a/doctr/datasets/datasets/pytorch.py
+++ b/doctr/datasets/datasets/pytorch.py
@@ -28,7 +28,8 @@ class AbstractDataset(_AbstractDataset):
         elif isinstance(target, tuple):
             assert isinstance(target[0], str) or isinstance(
                 target[0], np.ndarray
-            ), "Target should be a string or a numpy array"
+            ), "first element of the tuple should be a string or a numpy array"
+            assert isinstance(target[1], list), "second element of the tuple should be a list"
         else:
             assert isinstance(target, str) or isinstance(
                 target, np.ndarray
--- a/doctr/file_utils.py
+++ b/doctr/file_utils.py
@@ -10,13 +10,16 @@ import logging
 import os
 import sys
 
+CLASS_NAME: str = "words"
+
+
 if sys.version_info < (3, 8):
     import importlib_metadata
 else:
     import importlib.metadata as importlib_metadata
 
 
-__all__ = ["is_tf_available", "is_torch_available"]
+__all__ = ["is_tf_available", "is_torch_available", "CLASS_NAME"]
 
 ENV_VARS_TRUE_VALUES = {"1", "ON", "YES", "TRUE"}
 ENV_VARS_TRUE_AND_AUTO_VALUES = ENV_VARS_TRUE_VALUES.union({"AUTO"})
--- a/doctr/file_utils.py
+++ b/doctr/file_utils.py
@@ -88,3 +88,11 @@ def is_torch_available():
 
 def is_tf_available():
     return _tf_available
+
+
+def copy_tensor(x):
+    if is_tf_available():
+        import tensorflow as tf
+        return tf.identity(x)
+    elif is_torch_available():
+        return x.detach().clone()
--- a/doctr/file_utils.py
+++ b/doctr/file_utils.py
@@ -93,6 +93,7 @@ def is_tf_available():
 def copy_tensor(x):
     if is_tf_available():
         import tensorflow as tf
+
         return tf.identity(x)
     elif is_torch_available():
         return x.detach().clone()
--- a/doctr/file_utils.py
+++ b/doctr/file_utils.py
@@ -19,7 +19,7 @@ else:
     import importlib.metadata as importlib_metadata
 
 
-__all__ = ["is_tf_available", "is_torch_available", "CLASS_NAME"]
+__all__ = ["is_tf_available", "is_torch_available", "CLASS_NAME", "copy_tensor"]
 
 ENV_VARS_TRUE_VALUES = {"1", "ON", "YES", "TRUE"}
 ENV_VARS_TRUE_AND_AUTO_VALUES = ENV_VARS_TRUE_VALUES.union({"AUTO"})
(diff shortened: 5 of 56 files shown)


odulcy-mindee:
Test in CI `docker / pytest-api` will work after DocTR release :+1:

odulcy-mindee:
@felixdittrich92 Yes, it seems weird written like that haha.
It's due to these:
- https://github.com/mindee/doctr/pull/1097/files#diff-40f8f985c548e2b57b803d17f42833f8a4d94603ca5b5e0bf48de51872809400R16 -> Docker Image tries to install doctr from master which does not support multiclass
- https://github.com/mindee/doctr/pull/1097/files#diff-fac0b9b6096c5c832d53bfd0a96fd1311dc556b4dfaea23f4b3415a6cef33cd4 and https://github.com/mindee/doctr/pull/1097/files#diff-32f93bf544faa578de445d7ac8fe1428e2062d2bc0f854ed511bc96fa8236d35 which are files that @aminemindee needed to change for multiclass support

@aminemindee following @felixdittrich92's remark, I just saw that we need to enforce DocTR to be the latest version in the `pyproject.tom` in the API, otherwise it may still try to install an older version not supporting Multiclass

felixdittrich92:
@odulcy-mindee @aminemindee you need also to increase the major version specifier in root pyproject.toml (breaking change) and sort the milestones/open tickets :) 
After scrolling trough the changes i saw also that the xml export changes will break the PDF/A notebook fyi :)
I will try to review this tomorrow careful but for my understanding what is the general use case for multiclass text detection ? printed/handwritten ?

doctr/io/elements.py
```diff
@@ -311,65 +316,66 @@ def export_as_xml(self, file_title: str = "docTR - XML export (hOCR)") -> Tuple[
             },
         )
         # iterate over the blocks / lines / words and create the XML elements in body line by line with the attributes
-        for block in self.blocks:
```
felixdittrich92, line 314: It is a while ago i wrote this :sweat_smile: but we should test that it still works as expected with the PDF/A notebook

doctr/models/_utils.py
```diff
@@ -161,3 +161,39 @@ def get_language(text: str) -> Tuple[str, float]:
     if len(text) <= 1 or (len(text) <= 5 and lang.prob <= 0.2):
         return "unknown", 0.0
     return lang.lang, lang.prob
+
+
+def convert_list_dict_to_dict_list(list_dict: List[Dict[str, Any]]) -> Dict[str, List[Any]]:
```
felixdittrich92, line 166: Maybe something shorter ?

```
def invert_between_dict_of_lists_and_list_of_dicts(
    x: Union[List[Dict[str, Any]], Dict[str, List[Any]]]
) -> Union[List[Dict[str, Any]], Dict[str, List[Any]]]:

    if isinstance(x, dict):
        assert (
            len(set([len(v) for v in x.values()])) == 1
        ), "All the lists in the dictionnary should have the same length."
        return [dict(zip(x, t)) for t in zip(*x.values())]
    elif isinstance(x, list):
        return {k: [dic[k] for dic in x] for k in x[0]}
    else:
        raise TypeError(f"Expected input to be either a dict or a list, got {type(input)} instead.")
```

references/detection/train_pytorch.py
```diff
@@ -155,11 +157,14 @@ def evaluate(model, val_loader, batch_transforms, val_metric, amp=False):
...
         # Compute metric
         loc_preds = out["preds"]
-        for boxes_gt, boxes_pred in zip(targets, loc_preds):
-            if args.rotation and args.eval_straight:
-                # Convert pred to boxes [xmin, ymin, xmax, ymax]  N, 4, 2 --> N, 4
-                boxes_pred = np.concatenate((boxes_pred.min(axis=1), boxes_pred.max(axis=1)), axis=-1)
-            val_metric.update(gts=boxes_gt, preds=boxes_pred[:, :4])
+        for target, loc_pred in zip(targets, loc_preds):
+            if isinstance(target, np.ndarray):
+                target = {CLASS_NAME: target}
```
felixdittrich92, line 162: Should this name not come from the dataset specified names ? :thinking:

references/detection/train_tensorflow.py
```diff
@@ -115,11 +117,14 @@ def evaluate(model, val_loader, batch_transforms, val_metric):
...
         # Compute metric
         loc_preds = out["preds"]
-        for boxes_gt, boxes_pred in zip(targets, loc_preds):
-            if args.rotation and args.eval_straight:
-                # Convert pred to boxes [xmin, ymin, xmax, ymax]  N, 4, 2 --> N, 4
-                boxes_pred = np.concatenate((boxes_pred.min(axis=1), boxes_pred.max(axis=1)), axis=-1)
-            val_metric.update(gts=boxes_gt, preds=boxes_pred[:, :4])
+        for target, loc_pred in zip(targets, loc_preds):
+            if isinstance(target, np.ndarray):
+                target = {CLASS_NAME: target}
```
felixdittrich92, line 122: same ?

doctr/io/elements.py
```diff
@@ -215,7 +220,7 @@ class Page(Element):
     """Implements a page element as a collection of blocks
 
     Args:
-        blocks: list of block elements
+        blocks: Dictionary with list of block elements for each detection class
```
felixdittrich92, line 223: I still try to find a solution to avoid this breaking change but i think in front of this a short explanation would help me to understand the use case for multi class in text detection or better what you want to reach with it :sweat_smile: 

Is it not possible to add another key value pair like` 'class_names': 'words'` instead of changing the output structure ?

doctr/datasets/datasets/base.py
```diff
@@ -55,7 +56,13 @@ def __getitem__(self, index: int) -> Tuple[Any, Any]:
             img = self.img_transforms(img)
 
         if self.sample_transforms is not None:
-            img, target = self.sample_transforms(img, target)
+            if isinstance(target, dict):
+                img_transformed = copy_tensor(img)
+                for class_name, bboxes in target.items():
```
felixdittrich92, line 61: An OCRDataset comes also with dict as target and has no class names (string labels and bbox annotations) so if we transform samples in this case this should normally fail

references/detection/README.md
```diff
@@ -57,7 +57,30 @@ labels.json
      ...
 }
 ```
+If you want to train a model with multiple classes, you can use the following format
```
felixdittrich92, line 60: This should be a bit more clear:
multiclass box a ['printed', 'handwritten'] or box a ['printed'] box b ['handwritten']

felixdittrich92 (commented):
Hi @aminemindee @odulcy-mindee (seems you have both worked on it :sweat_smile:) thanks for the PR i added a first general review without going to deep and left some questions and remarks :)

Due to the breaking change and the amount of changes, someone should have a second look at it. @frgfm ?

references/detection/README.md
felixdittrich92: ok i think now i got it :sweat_smile: 

for example class_name_1 == 'printed' contains all boxes for printed and class_name_2 == 'handwritten' contains all boxes for handwritten.

Maybe a better format:

polygons: [....]
classes: ['printed', 'handwritten', 'printed', 'printed', ...] 

So we can avoid to introduce a new dataset format then it should also be possible to avoid the breaking change only to add a new key value pair like 'class': 'printed' @aminemindee @odulcy-mindee wdyt ?

aminemindee:
> @odulcy-mindee @aminemindee you need also to increase the major version specifier in root pyproject.toml (breaking change) and sort the milestones/open tickets after merge/release :) After scrolling trough the changes i saw also that the xml export changes will break the PDF/A notebook fyi :) I will try to review this tomorrow careful but for my understanding what is the general use case for multiclass text detection ? printed/handwritten ?

Hello @felixdittrich92  the general use case of multiclass text detection is when you want to find not just the words in a document but some specific infomations. An example, in a receipt from a restaurant you want to find the name of the restaurant and the total you paid. then you can have two classes and predict different bounding boxes for each class.

felixdittrich92:
> > @odulcy-mindee @aminemindee you need also to increase the major version specifier in root pyproject.toml (breaking change) and sort the milestones/open tickets after merge/release :) After scrolling trough the changes i saw also that the xml export changes will break the PDF/A notebook fyi :) I will try to review this tomorrow careful but for my understanding what is the general use case for multiclass text detection ? printed/handwritten ?
> 
> Hello @felixdittrich92 the general use case of multiclass text detection is when you want to find not just the words in a document but some specific infomations. An example, in a receipt from a restaurant you want to find the name of the restaurant and the total you paid. then you can have two classes and predict different bounding boxes for each class.

Ok got it so in the end like the LayoutLM model (without the text input)

references/detection/train_pytorch.py
aminemindee: for the training part, you can still use a data format where you don't give any class name. in that case, targets are not a dict so in that case i change it here. But now that i'm looking at it i think it's better to just add the CLASS_NAME in the dataset and not handle it outside. right ?

doctr/io/elements.py
aminemindee: hmm something like a class_names is the same len as blocks and give the class name of each block ?

doctr/io/elements.py
aminemindee: i looked at it a little bit. in google colab the installation using git doesn't seem to work for me. However, the code should work just the output will be slightly different i think.

doctr/datasets/datasets/pytorch.py
```diff
@@ -25,6 +25,11 @@ def _read_sample(self, index: int) -> Tuple[torch.Tensor, Any]:
         if isinstance(target, dict):
             assert "boxes" in target, "Target should contain 'boxes' key"
             assert "labels" in target, "Target should contain 'labels' key"
+        elif isinstance(target, tuple):
```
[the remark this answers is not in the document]
aminemindee: In here, for now the only moment when target is a tuple is coming from detection dataset when there is multiple classes to keep that information and include it into the target in the _pre_transform. 

yes we should be careful but for now this is the only case where it will be used.

doctr/models/_utils.py
aminemindee: Love it ! Changed it :+1:

doctr/datasets/datasets/base.py
aminemindee: Yes that's true :fearful:  i'm trying to think how to handle it . open to any suggestions x)

references/detection/train_pytorch.py
felixdittrich92: I think we need to avoid this CLASS_NAME completly ... to explain a box can be annotated with different labels like:
box 1: name, box 2: salary i do not understand correctly why we need to modify in the end the blocks output structure 
i think what we want is this:

```
(words): [
          Word(value='ABC Gmbh.', confidence=0.91, class_name='name'),
          Word(value='1200', confidence=0.99, class_name='salary),
        ]
```
right ?

doctr/datasets/datasets/base.py
felixdittrich92: As described in the other comments if we handle it like it is done with object_detection (classes comes from model.cfg which is flexible and can be updated with classes from dataset) we could remove this 
Maybe thinking again about this dataset format:
polygons: [....]
classes: ['printed', 'handwritten', 'printed', 'printed', ...]  ? @frgfm @odulcy-mindee @charlesmindee wdyt ?

api/pyproject.toml
```diff
@@ -4,16 +4,16 @@ build-backend = "poetry.masonry.api"
 
 [tool.poetry]
 name = "doctr-api"
-version = "0.5.2a0"
+version = "1.0.1a0"
```
frgfm, line 7: :warning: This is the version of the package, not of poetry
I'd suggest reverting back

api/pyproject.toml
```diff
@@ -4,16 +4,16 @@ build-backend = "poetry.masonry.api"
...
 description = "Backend template for your OCR API with docTR"
 authors = ["Mindee <contact@mindee.com>"]
 license = "Apache-2.0"
 
 [tool.poetry.dependencies]
-python = ">=3.8,<3.11"
+python = ">=3.8.2,<3.11"
```
frgfm, line 13: Could you add a comment to specify the reason of the minor version? (much easier to reassess later when we have a comment saying why)

api/pyproject.toml
frgfm: Actually, I realize you want to bump docTr to 1.0.0. Perhaps that might be easier to discuss this in an issue and lower the amount of modifications in this specific PR?

api/pyproject.toml
```diff
@@ -4,16 +4,16 @@ build-backend = "poetry.masonry.api"
...
 tensorflow = ">=2.9.0,<3.0.0"
 tensorflow-addons = ">=0.17.1"
-python-doctr = ">=0.2.0"
+python-doctr = { version = ">=1.0.0", extras = ['tf'] }
```
frgfm, line 16: Correct me if I'm wrong, but I think this will result as double version specifiers. If we consider extras, we need to remove the "tensorflow" & cie specifiers

doctr/datasets/datasets/base.py
frgfm: I think we should split the topics of this PR, and discuss this in several issues/discussions to find a proper implementations. There are several breaking change modifications in the PR over quite a lot of files, it's quite hard to ensure the review quality will be both high and exhaustive :sweat_smile:

doctr/datasets/datasets/pytorch.py
```diff
@@ -25,6 +25,11 @@ def _read_sample(self, index: int) -> Tuple[torch.Tensor, Any]:
...
+            assert isinstance(target[0], str) or isinstance(
+                target[0], np.ndarray
+            ), "first element of the tuple should be a string or a numpy array"
+            assert isinstance(target[1], list), "second element of the tuple should be a list"
```
frgfm, line 32: The problem with this type of condition is that if the target is a tuple with a single element the code will crash. Either we have to add `assert len(target) == 2` or change that a bit

frgfm (commented):
Hello everyone :wave:

Thanks a lot of the PR! It's only my opinion but I think we have too many things at stake in the PR to be able to say whether it's OK to merge or not. I would suggest pausing this one shortly and opening GH issues for the following topics:
- dataset target format modification for multiple class of text/objects
- versioning of docTR (when is 1.0.0 within reach, etc.)
- minimal API modifications
- export format modifications
- detection model modifications
- training/evaluation

_I might have forgotten a few_


Once a topic has reached a good implementation idea / way to handle it, we can cherry pick from this PR or reimplement it :) 

I already added comments for the section I checked, but then thought there is too much to handle in a single PR :sweat_smile: If you prefer to go with that one in a single PR, no worries!

Cheers :v:

doctr/io/elements.py
charlesmindee: Actually blocks are not resolved, so this sub object in the `Page` might be obsolete, we could either rename it to `Layer` to have a more meaningful page structure and keep it like that or add a `Layer` object between` Page` and `Block` in the hierarchy, so that a class would be on a layer and inside each layer we can resolve blocks and lines.

doctr/models/_utils.py
```diff
@@ -161,3 +161,18 @@ def get_language(text: str) -> Tuple[str, float]:
     if len(text) <= 1 or (len(text) <= 5 and lang.prob <= 0.2):
         return "unknown", 0.0
     return lang.lang, lang.prob
+
+
+def invert_between_dict_of_lists_and_list_of_dicts(
```
charlesmindee, line 166: `invert_data_structure` with a clear docstring which explains the typing  maybe ? so that we keep a short and clear name ?

doctr/file_utils.py
```diff
@@ -10,13 +10,16 @@
 import os
 import sys
 
+CLASS_NAME: str = "words"
```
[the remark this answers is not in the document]
charlesmindee: I think it is ok to have this default single class and use the class of the datasets otherwise to build the model, but the user should be able to modify this CLASS_NAME to a specific value like "words", "numbers", "default_class", "handwritten"... so it shouldn't be hardcoded here but I rather see it as a parameter given by the user either in the config or when building the mode, WDYT ?

charlesmindee (commented):
A few remarks, but minor

references/detection/train_pytorch.py
aminemindee: The multiclass detection model is prediction bounding boxes that represent the blocks. so in theory the whole block is of that class and the words inside are of that class. A user will want to get all the words of the same block of that class and he can get that directly with the blocks as dict approach. different words in the same block will always have the same class also.

doctr/file_utils.py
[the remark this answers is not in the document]
felixdittrich92: Hi @charlesmindee yep sounds good to me 👍 

I see it much more problematic to modify the blocks output structure .. the results should be added to each predicted word like

'house', conf: '0.99', class: 'word'

Instead of:

Blocks: 
'word':  { Line .. Words ..},
'salary': { Line .. Words ..}

Wdyt ?

doctr/models/_utils.py
felixdittrich92: Sounds good :)

api/pyproject.toml
aminemindee: i don't think i understand your comment can you be more clear? 
we added the extras because now doctr if you don't install the extras "tf" you don't have tf2onnx library and things like that, that are needed to run the code.

doctr/models/_utils.py
aminemindee: Done

doctr/datasets/datasets/pytorch.py
aminemindee: added it.

api/pyproject.toml
aminemindee: Yes good idea ! added it.

aminemindee:
> Hello everyone wave
> 
> Thanks a lot of the PR! It's only my opinion but I think we have too many things at stake in the PR to be able to say whether it's OK to merge or not. I would suggest pausing this one shortly and opening GH issues for the following topics:
> 
>     * dataset target format modification for multiple class of text/objects
> 
>     * versioning of docTR (when is 1.0.0 within reach, etc.)
> 
>     * minimal API modifications
> 
>     * export format modifications
> 
>     * detection model modifications
> 
>     * training/evaluation
> 
> 
> _I might have forgotten a few_
> 
> Once a topic has reached a good implementation idea / way to handle it, we can cherry pick from this PR or reimplement it :)
> 
> I already added comments for the section I checked, but then thought there is too much to handle in a single PR sweat_smile If you prefer to go with that one in a single PR, no worries!
> 
> Cheers v

Hello @frgfm i understand that this PR is kind of big with a lot of changes. The goal is to have it available as soon as possible. i know that the review won't be exhaustive but we can see this 1.0.0 release as experimental and once it's there we can improve things. After this PR, i will work on improving the training part ( changing losses maybe and also adding some augmentations and things of the sort)

api/pyproject.toml
frgfm: Yes sure, I wasn't referring to the extra but the version index specifier
You switched it from `>=0.2.0` to `>=1.0.0` which is a big deal :sweat_smile:

frgfm: Regarding the extra, I'd argue it's up for discussion but the API doesn't need to convert models to ONNX

Ideally, we'd like it to run on ONNX, but some postprocessing ops are not compatible yet. So the alternative is to use doctr, and then carefully add as little deps as possible to run it

frgfm:
Hi @aminemindee :)

> i know that the review won't be exhaustive but we can see this 1.0.0 release as experimental and once it's there we can improve things. 

I could agree with your suggestion if it were 0.7.0, but 1.0.0 in software dev, at least to the best of my experience & knowledge, is a mark of stability in the API (in the broad sense) and performance milestone. So jumping from 0.6.0 to 1.0.0 looks a bit hasty to me (even the 0.7.0 release tracker is full of things that would be worth considering before pushing a 1.0.0)

To put this into context, the day the community saw PyTorch jump from 0.4.0 to 1.0.0, it was because they added the C++ API, the JIT compiler (which was their target for the first 1.0.0 release). We knew that this could now be considered to run in production. Here, at least I think the release roadmap should follow a structure readable by the community (cf. release trackers) :sweat_smile: 

> After this PR, i will work on improving the training part ( changing losses maybe and also adding some augmentations and things of the sort)

Sounds good :+1:

doctr/datasets/datasets/base.py
aminemindee: I added another check in addition to if it's a dict to ensure it's the target coming from detection.

felixdittrich92:
About the versioning, i agree with @frgfm 1.0.0 should be the first stable (production ready) version specifier. 
On the other hand with the breaking change we need to make it clear for end users. 
I would still suggest to rethink some parts so that we can implement it without modifying the `Page` structure (for example as extra attribute `class` for each `Word`) cf. https://github.com/mindee/doctr/blob/acb9f64b11ebad8e53ac60737fcde8dbd3158a22/doctr/io/elements.py#L57

api/pyproject.toml
aminemindee: I agree. but for now how doctr is done to run the code in the api tf2onnx is needed

aminemindee:
https://github.com/mindee/doctr/pull/1097#discussion_r996300298

> About the versioning, i agree with @frgfm 1.0.0 should be the first stable (production ready) version specifier. On the other hand with the breaking change we need to make it clear for end users. I would still suggest to rethink some parts so that we can implement it without modifying the `Page` structure and avoid the breaking change (for example as extra attribute `class` for each `Word`) cf.
> 
> https://github.com/mindee/doctr/blob/acb9f64b11ebad8e53ac60737fcde8dbd3158a22/doctr/io/elements.py#L57

I don't think adding an attribute `class` to `Word` would be ideal, because for a user who wants all the blocks of words from a certain class will have to iterate over every block line and words to get them. 

I think the main problem here is that, once the detection model is not trying to detect all words in a page, the fact that `Page` is a list of `Block` is not logical anymore because it doesn't constitute the page, but just give the information we found on the page. 

So i agree with @charlemindee, we can add a new object `Layer` but instead of between  `Page` and `Blocks` but at the same level of Blocks with the same information just as a dictionary. WDYT @fg-mindee @charlesmindee @felixdittrich92 ?

felixdittrich92:
> [#1097 (comment)](https://github.com/mindee/doctr/pull/1097#discussion_r996300298)
> 
> > About the versioning, i agree with @frgfm 1.0.0 should be the first stable (production ready) version specifier. On the other hand with the breaking change we need to make it clear for end users. I would still suggest to rethink some parts so that we can implement it without modifying the `Page` structure and avoid the breaking change (for example as extra attribute `class` for each `Word`) cf.
> > https://github.com/mindee/doctr/blob/acb9f64b11ebad8e53ac60737fcde8dbd3158a22/doctr/io/elements.py#L57
> 
> I don't think adding an attribute `class` to `Word` would be ideal, because for a user who wants all the blocks of words from a certain class will have to iterate over every block line and words to get them.
> 
> I think the main problem here is that, once the detection model is not trying to detect all words in a page, the fact that `Page` is a list of `Block` is not logical anymore because it doesn't constitute the page, but just give the information we found on the page.
> 
> So i agree with @charlemindee, we can add a new object `Layer` but instead of between `Page` and `Blocks` but at the same level of Blocks with the same information just as a dictionary. WDYT @fg-mindee @charlesmindee @felixdittrich92 ?

Mh .. we should keep in mind that it is an feature task core is still OCR (where a user wants to get all the extracted text at once) so i think it would be totally ok if the user needs to iterate trough the results ... but yeah that's my opinion 😅 let's wait on @frgfm @charlesmindee 👍

aminemindee:
Hello again ! After discussing it internally, i have a new suggestion to add the detection multiclass without a big breaking change. 

I added a new kie_predictor ( like ocr_predictor) that will handle the different classes and has a dict like output, while keeping ocr_predictor as it was. 
Also, we're not going major anymore but just to 0.7.0 and maybe what was already tracked for the 0.7.0 will become for 0.8.0 i guess.

So a summary of the changes: 

- Addition of kie_predictor and some io classes to use it (KIEDocument, KIEPage, ..)
- ocr_predictor has the same output as before, no breaking change here
- Detection model training as it was before with new dataformat in input and outputs a dict 
- Test were fixed and added tests for kie_predictor
- kie route was added on api

WDYT @felixdittrich92  @odulcy-mindee  @frgfm @charlesmindee ?

doctr/models/detection/linknet/base.py
```diff
@@ -156,78 +156,93 @@ class _LinkNet(BaseModel):
...
             raise AssertionError("the expected dtype of target 'boxes' entry is 'np.float32'.")
-        if any(np.any((t[:, :4] > 1) | (t[:, :4] < 0)) for t in target):
+        if any(np.any((t[:, :4] > 1) | (t[:, :4] < 0)) for tgt in target for t in tgt.values()):
             raise ValueError("the 'boxes' entry of the target is expected to take values between 0 & 1.")
 
-        h, w = output_shape
-        target_shape = (len(target), h, w, 1)
+        h: int
+        w: int
+        if is_tf_available():
```
felixdittrich92, line 179: We should avoid the permanent usage of torch / tf available better:

check shape: channels last => Tf / channels first => PT   (especially if we think further -> Onnx this conditions have to be removed completly if we want to reuse Pre-/Postprocessing

felixdittrich92 (commented):
Hi @aminemindee thanks for the changes i added a few comments but in general there are some questions / remarks and the PR is still much to big to review it careful :sweat_smile:

doctr/models/kie_predictor/pytorch.py
```diff
@@ -0,0 +1,175 @@
+# Copyright (C) 2021-2022, Mindee.
```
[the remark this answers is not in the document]
aminemindee: I agree with you that a lot of things are duplicated for now. But as there is no breaking change, it's something that we can work on after this is released and make it better. ( remove the redundancies, maybe have a abstract predictor more general and inherit from it, ...) 

For the fact that the PR is too big and opening dicsussion/issue about some part, unfortunately we don't have time to do that. We want the multiclass feature to be available as soon as possible, even if it's not perfect and improve it after but at least have it running.

doctr/models/kie_predictor/base.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (C) 2021-2022, Mindee.
```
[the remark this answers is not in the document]
aminemindee: done

doctr/models/detection/linknet/base.py
aminemindee: Added the info from where the method was used.

doctr/models/detection/differentiable_binarization/pytorch.py
```diff
@@ -112,20 +114,26 @@ class DBNet(_DBNet, nn.Module):
...
         deform_conv: bool = False,
-        num_classes: int = 1,
         assume_straight_pages: bool = True,
         exportable: bool = False,
         cfg: Optional[Dict[str, Any]] = None,
+        class_names: Optional[List[str]] = None,
     ) -> None:
 
         super().__init__()
+        if class_names:
```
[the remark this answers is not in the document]
aminemindee: Removed.

api/app/routes/kie.py
```diff
@@ -0,0 +1,31 @@
+# Copyright (C) 2021-2022, Mindee.
```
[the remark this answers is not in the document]
aminemindee: Yes done

aminemindee:
https://github.com/mindee/doctr/pull/1097#discussion_r1006613171 

in here it's just checking that target has the right format, for the transformations when target is a tuple ( only in detection) the _pre_transform handles it. Can you be more precise to what's troublesome ?

doctr/models/kie_predictor/pytorch.py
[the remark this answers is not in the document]
felixdittrich92: Mh .. i agree if you want to use it "now" but sometimes it is better to take the time instead of brute forcing things :sweat_smile: the refactoring will take some more time as doing it now in an better way and i am sure it will break again things which are done in your PR. :thinking: 

Unfortunately, I don't see the PR in good condition right now, which is why I wouldn't / can't approve it, but in the end this is not my decision, if you have discussed this internally (@odulcy-mindee @charlesmindee).

felixdittrich92:
> [#1097 (comment)](https://github.com/mindee/doctr/pull/1097#discussion_r1006613171)
> 
> in here it's just checking that target has the right format, for the transformations when target is a tuple ( only in detection) the _pre_transform handles it. Can you be more precise to what's troublesome ?

Sure sorry for the lazy comment :sweat_smile: 

We have all the time tried to reduce / keep simple the data formats which we need to handle afterwards in 

```
img_transforms: Optional[Callable[[Any], Any]] = None,
sample_transforms: Optional[Callable[[Any, Any], Tuple[Any, Any]]] = None,
pre_transforms: Optional[Callable[[Any, Any], Tuple[Any, Any]]] = None,
```

Adding things that are only needed for a specific case increases the complexity in this case instead of unifying it.
I have really spent some time to bring all the datasets related stuff in an easy unified format to make at reusable, easy to understand and especially unified. :sweat_smile: 
If I remember correctly, the last stand was that we wanted to convert everything into a dictionary.

charlesmindee (dismissed):
Thanks, everything LGTM just add the entries in the docs for predictor/IO and update the readme with a KIE section

charlesmindee (dismissed):
Thanks!

charlesmindee (dismissed):
LGTM, thanks for the improvements

charlesmindee (approved):

PR #1110, opened 2022-11-30
feat: :sparkles:  expose thresh bin thresh in DBNet

charlesmindee:
This PR exposes the `bin_thresh` in the DBNet args so that one can build a predictor with a binarization threshold for the segmentation task adapted to his use case.
It can be done this way:
```

from doctr.models import detection_predictor, db_resnet50

predictor = detection_predictor(db_resnet50(pretrained=True, bin_thresh=0.1))
```

The threshold is left at 0.3 by default.
Any feedback is welcome!


felixdittrich92 (commented):
@charlesmindee you can pass the model directly into `ocr_predictor` without wrapping it in `detection_predictor` :)

scripts/detect_text.py
```diff
@@ -82,6 +89,7 @@ def parse_args():
     )
     parser.add_argument("path", type=str, help="Path to process: PDF, image, directory")
     parser.add_argument("--detection", type=str, default="db_resnet50", help="Text detection model to use for analysis")
+    parser.add_argument("--bin_thresh", type=float, default=0.3, help="binarization threshold for the detection model")
```
ianardee, line 92:
```suggestion
    parser.add_argument("--bin-thresh", type=float, default=0.3, help="Binarization threshold for the detection model.")
```

aminemindee (dismissed):

aminemindee (approved):

Issue #1020, opened 2022-08-16
GPU inference error using db_resne50_rotation


harindercnvrg:
### Bug description

Hello, guys I am getting the following error by running the snipped of code. I don't  get the same error if I use db_resnet50 instead of db_resnet50_rotation. Help please! I am using the version 0.5.1

### Code snippet to reproduce the bug

```
from doctr.io import DocumentFile
from doctr.models import ocr_predictor
tic = time.time()
doc = DocumentFile.from_pdf("/content/file.pdf")
print(f"Number of pages: {len(doc)}")
predictor = ocr_predictor('db_resnet50_rotation','crnn_vgg16_bn',pretrained=True,assume_straight_pages=False).cuda()
result = predictor(doc)
```

### Error traceback

```
RuntimeError: Input type (torch.FloatTensor) and weight type (torch.cuda.FloatTensor) should be the same or input should be a MKLDNN tensor and weight is a dense tensor
```

### Environment
```
!pip install python-doctr[torch]
!sudo apt-get install fonts-freefont-ttf -y
!pip3 install --no-build-isolation -U pypdfium2==1.0.0
```
### Deep Learning backend
```
is_tf_available: False
is_torch_available: True
```

harindercnvrg:
First, could you elaborate on "I don't get the same error if I use db_resnet50 instead of db_resnet50_rotation". What error do you get if you use that one?
Yes, please look below:

frgfm:
Thanks for that! The traceback gave the source of the problem away: the input tensor isn't being moved to the device of the model as it should.

I'll fix this shortly then :) 
I'll let you know here when it's merged!

felixdittrich92:
Hi @harindercnvrg should work now (on main) 👋

minouei-kl:
Hi
I'm getting a similar error using db_resne50_rotation with version 0.7.1a0.
it is working fine with db_resne50.
Error traceback
```
Traceback (most recent call last):
  File "ocr.py", line 42, in <module>
    result = model(image)
  File "/home/minouei/miniconda3/envs/mmocr3/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1102, in _call_impl
    return forward_call(*input, **kwargs)
  File "/home/minouei/miniconda3/envs/mmocr3/lib/python3.8/site-packages/torch/autograd/grad_mode.py", line 28, in decorate_context
    return func(*args, **kwargs)
[...]

PR #1039, opened 2022-09-01
[Fix] rotation classifier input move to model device

felixdittrich92:
This PR:

- fix moving rotation classifier input to cuda device if given

Closes:
#1020

PR #1125, opened 2023-01-09
fix classification model cuda move

odulcy-mindee:
Related to https://github.com/mindee/doctr/pull/1039, closes https://github.com/mindee/doctr/issues/1020


felixdittrich92 (approved):
@odulcy-mindee thanks for the quick fix :smile:

Issue #1126, opened 2023-01-12
API docker build 


meliascosta:
### Bug description

I tried to run the command `make lock` but I get the following error: 

**Because doctr-api depends on python-doctr (>=0.7.0) which doesn't match any versions, version solving failed.**

It looks like the pyproject.toml file is already prepared to use version 0.7.0 but the latest version published on PyPI is 0.6.0

### Code snippet to reproduce the bug

```bash
make lock
```

### Error traceback

*Because doctr-api depends on python-doctr (>=0.7.0) which doesn't match any versions, version solving failed.

### Environment

DocTR version: N/A
TensorFlow version: N/A
PyTorch version: N/A (torchvision N/A)
OpenCV version: N/A
OS: Ubuntu 20.04.5 LTS
Python version: 3.9.7
Is CUDA available (TensorFlow): N/A
Is CUDA available (PyTorch): N/A
CUDA runtime version: Could not collect
GPU models and configuration: Could not collect
Nvidia driver version: Could not collect
cuDNN version: Could not collect

### Deep Learning backend

Not applicable this is before installation

odulcy-mindee:
Hello @meliascosta,

Thanks for this issue. Indeed, your diagnostic is correct:
> It looks like the pyproject.toml file is already prepared to use version 0.7.0 but the latest version published on PyPI is 0.6.0

We have not made a release on `PyPI` yet. I need to check with the team when we'll release it.

In the meantime, I fixed this problem in https://github.com/mindee/doctr/issues/1126

PR #1148, opened 2023-02-15
fix: :wrench: docker api use GitHub repository

odulcy-mindee:
Fix https://github.com/mindee/doctr/issues/1126

Now, the docker container in `api` folder uses directly DocTR GitHub repository instead of the wheel on PyPI. This is done until `0.7.0` is released.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -34,7 +34,4 @@ jobs:
         run: wget --spider --tries=12 http://localhost:8080/docs
       - name: Run docker test
         run: |
-          docker cp api/requirements-dev.txt api_web_1:/app/requirements-dev.txt
-          docker-compose -f api/docker-compose.yml exec -T web pip install -r requirements-dev.txt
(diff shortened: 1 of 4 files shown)


felixdittrich92 (approved):
makes sense :)
Thanks @odulcy-mindee

aminemindee (dismissed):
LGTM

felixdittrich92:
@odulcy-mindee job should be update https://github.com/mindee/doctr/blob/main/.github/workflows/docker.yml seems make lock is missing

odulcy-mindee:
> @odulcy-mindee job should be updated https://github.com/mindee/doctr/blob/main/.github/workflows/docker.yml seems make lock is missing

@felixdittrich92 well, it's not needed anymore, I don't know why there is still these instructions, because `make lock` occurs in the container now. We only need to execute `pytest` in the container now :ok_hand:

felixdittrich92 (approved):
Thanks :)

PR #1149, opened 2023-02-15
lint: :lipstick: mypy and black

odulcy-mindee:
Fix lint with `mypy` and `black`.

Also replaced `use_auth_token` by `token` in `doctr/models/factory/hub.py`, see https://github.com/huggingface/huggingface_hub/pull/1122

### DIFF
--- a/api/.gitignore
+++ b/api/.gitignore
@@ -1,0 +1,2 @@
+poetry.lock
+requirements*
--- a/api/Makefile
+++ b/api/Makefile
@@ -9,17 +9,17 @@ lock:
 
 # Run the docker
 run:
-	docker-compose up -d --build
+	docker compose up -d --build
 
 # Run the docker
 stop:
-	docker-compose down
+	docker compose down
 
 # Run tests for the library
 test:
-	docker-compose up -d --build
+	docker compose up -d --build
 	docker cp requirements-dev.txt api_web_1:/app/requirements-dev.txt
-	docker-compose exec -T web pip install -r requirements-dev.txt
+	docker compose exec -T web pip install -r requirements-dev.txt
 	docker cp tests api_web_1:/app/tests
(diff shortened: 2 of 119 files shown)


felixdittrich92 (changes requested):
Thanks @odulcy-mindee 

`api/imgjpg `can be removed and `api/requirements-dev.txt` and `api/requirements.txt` should be added to gitignore :)

odulcy-mindee:
> Thanks @odulcy-mindee
> 
> `api/imgjpg `can be removed and `api/requirements-dev.txt` and `api/requirements.txt` should be added to gitignore :)

Oops, good catch ! Thanks !

felixdittrich92 (approved):
Thanks :)

Are you sure that `docker compose` instead of `docker-compose `in a Makefile is valid ?

odulcy-mindee:
> Are you sure that `docker compose` instead of `docker-compose `in a Makefile is valid ?

Yeah, now `docker compose` is drop-in replacement of `docker-compose`, see: https://docs.docker.com/compose/compose-v2/#differences-between-compose-v1-and-compose-v2. It's now a plugin which is already in `docker`, we do not need to have a standalone binary.

See also https://github.com/docker/compose#about-update-and-backward-compatibility

felixdittrich92:
> > Are you sure that `docker compose` instead of `docker-compose `in a Makefile is valid ?
> 
> Yeah, now `docker compose` is drop-in replacement of `docker-compose`, see: https://docs.docker.com/compose/compose-v2/#differences-between-compose-v1-and-compose-v2. It's now a plugin which is already in `docker`, we do not need to have a standalone binary.
> 
> See also https://github.com/docker/compose#about-update-and-backward-compatibility

Ah nice and good to know ty :+1:

PR #1179, opened 2023-04-20
[build] replaces flake8 with ruff

felixT2K:
This PR:

- depends on: #1180 
- replaces flake8 with ruff (same config as flake8 before)

Benefits:
- config compatible with pyproject.toml
- much faster as flake8

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -20,11 +20,11 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
           architecture: x64
-      - name: Run flake8
+      - name: Run ruff
         run: |
-          pip install flake8
-          flake8 --version
-          flake8
+          pip install ruff
+          ruff --version
+          ruff check --diff .
 
   black:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -7,7 +7,7 @@ on:
     branches: main
 
 jobs:
-  flake8:
+  ruff:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
--- a/docs/build.sh
+++ b/docs/build.sh
@@ -5,14 +5,7 @@ function deploy_doc(){
     fi
     COMMIT=$(git rev-parse --short HEAD)
     echo "Creating doc at commit" $COMMIT "and pushing to folder $2"
-    # Hotfix
-    if [ "$1" \< "dcbb21f" ]
-        then
-            pip install -U ..
-            pip install rapidfuzz==2.15.1
-        else
-            pip install -U ..
-        fi
+    pip install -U ..
     if [ ! -z "$2" ]
     then
         if [ "$2" == "latest" ]; then
--- a/docs/build.sh
+++ b/docs/build.sh
@@ -5,7 +5,14 @@ function deploy_doc(){
     fi
     COMMIT=$(git rev-parse --short HEAD)
     echo "Creating doc at commit" $COMMIT "and pushing to folder $2"
-    pip install -U ..
+    # Hotfix
+    if [ "$1" \< "dcbb21f" ]
+        then
+            pip install -U ..
+            pip install rapidfuzz==2.15.1
+        else
+            pip install -U ..
+        fi
     if [ ! -z "$2" ]
     then
         if [ "$2" == "latest" ]; then
--- a/doctr/version.py
+++ b/doctr/version.py
@@ -1,0 +1,6 @@
+# Copyright (C) 2021-2023, Mindee.
+
+# This program is licensed under the Apache License 2.0.
+# See LICENSE or go to <https://opensource.org/licenses/Apache-2.0> for full license details.
+
+__version__ = '0.6.1a0'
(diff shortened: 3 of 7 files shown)


.github/workflows/style.yml
```diff
@@ -20,11 +20,11 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
           architecture: x64
-      - name: Run flake8
+      - name: Run ruff
```
frgfm, line 23: Thanks! Let's rename the job on line 10 as well :pray:

doctr/version.py
frgfm, line 1: Indeed, let's change this but this is for another PR (+ adding it to gitignore)

pyproject.toml
```diff
@@ -170,6 +170,18 @@ src_paths = ["doctr", "tests", "scripts", "references", "demo", "docs", "api"]
 skip_glob = "**/__init__.py"
 known_third_party = ["tensorflow", "torch", "torchvision", "wandb", "fastprogress"]
 
+[tool.ruff]
+ignore = ["E402", "F403", "E731"]
+exclude = [".github", ".git", "venv*", "docs", "build"]
```
frgfm, line 175: same here (if there are python script in .github or docs, it might be nice to have those checked by ruff?)

setup.py
```diff
@@ -9,7 +9,7 @@
 from setuptools import setup
 
 PKG_NAME = "python-doctr"
-VERSION = os.getenv("BUILD_VERSION", "0.7.1a0")
+VERSION = os.getenv("BUILD_VERSION", "0.6.1a0")
```
frgfm, line 12: Thanks for that fix, but I also suggest to keep that for another PR :)

frgfm (commented):
Thanks a lot @felixT2K :pray: 

Two high-level comments:
- loving the suggestion to switch to ruff, I fully agree with that. I only added a few comments to make sure we have no regression from flake8 previous config
- for the rest, everything is on point, but let's make one objective per PR and focus that one on ruff if that's ok :smiley:

pyproject.toml
```diff
@@ -170,6 +170,18 @@ src_paths = ["doctr", "tests", "scripts", "references", "demo", "docs", "api"]
 skip_glob = "**/__init__.py"
 known_third_party = ["tensorflow", "torch", "torchvision", "wandb", "fastprogress"]
 
+[tool.ruff]
+ignore = ["E402", "F403", "E731"]
```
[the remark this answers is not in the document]
felixdittrich92: The removed patterns seems to be invalid (.flake8 was unchecked) i will take a look again

doctr/version.py
felixdittrich92: Is already added in gitignore don't know how this could happen before 😅

docs/build.sh
```diff
@@ -5,7 +5,14 @@ function deploy_doc(){
...
     echo "Creating doc at commit" $COMMIT "and pushing to folder $2"
-    pip install -U ..
+    # Hotfix
+    if [ "$1" \< "dcbb21f" ]
+        then
+            pip install -U ..
+            pip install rapidfuzz==2.15.1
+        else
+            pip install -U ..
+        fi
```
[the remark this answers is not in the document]
felixdittrich92: Will open another one for the Version and this

.github/workflows/style.yml
felixdittrich92: Good catch i totally missed this :D

pyproject.toml
felixdittrich92: Sure i have no use Case in mind currently but why not :)

felixT2K:
@frgfm Do you think it would make sense to add pre-commit also ? :) 
(I personally like it one of my current colleagues hates it like hell 😂 )
I was so free to merge this minor stuff directly in #1180 think that's fine

pyproject.toml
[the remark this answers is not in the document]
felixT2K: yep for the removed onces it says: Unknown rule selector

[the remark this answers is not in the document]
felixT2K:
```
By default, Ruff enables Flake8's E and F rules. Ruff supports all rules from the F category, and a [subset](https://beta.ruff.rs/docs/rules/#error-e) of the E category, omitting those stylistic rules made obsolete by the use of an autoformatter, like [Black](https://github.com/psf/black).
```

pyproject.toml
```diff
@@ -70,7 +70,7 @@ testing = [
     "requests>=2.20.0",
 ]
 quality = [
-    "flake8>=3.9.0",
+    "ruff>=0.0.260",
```
frgfm, line 73: out of security, I would have added an upper bound as well 0.1.0 for instance to avoid troubles with the CI later but nothing major :+1:

frgfm (approved):
Nice, thanks for the migration Felix :+1: 
(I just noticed @felixdittrich92 & @felixT2K are the same haha)

felixdittrich92:
> Nice, thanks for the migration Felix 👍 (I just noticed @felixdittrich92 & @felixT2K are the same haha)

Yeah it's my work profile 😅

Issue #978, opened 2022-07-07
[models] Replace current tf mobilenetv3 implementation with tf.keras.applications implementation


felixdittrich92:
### Bug description

- implementation is still troublesome on CPU (grouped convolutions)
- is not ONNX exportable (some strange error messages)
- cannot be pushed to the huggingface hub (raise with: model contains binary data - invalid model)

### Code snippet to reproduce the bug

-

### Error traceback

-

### Environment

-

### Deep Learning backend

-

aminemindee:
Hello Felix, I started working on this issue and i have some questions/remarks
- Using the keras Implementation directly i can see that the mobilenetv3 small has a different number of parameters from the keras implementation compared to doctr implementation. 
- is the weights for mobilenetv3 the pretrained ones (from hub) or are weights trained by mindee? 
- if it's the second case should we use keras implementation but reconcile the model to have the same parameters as the one in doctr to be able to reload weights?  ( Doing that means reusing the `InvertedResidual` class)

These are some questions i kind of face when trying to look at it. 
Feel free to give any other information or suggestions about it 
Thank you

felixdittrich92:
Hi @aminemindee :wave:,
nice that you have started to work on it :+1: 
One thing in front of your questions we can solve this issue in a draft PR but have to wait before we merge it until 0.5.2 is released otherwise it would be a big breaking change for current users of doctr (why ? this change affects on TF side db_mobilenet_v3_large / crnn_mobilenet_v3_small / crnn_mobilenet_v3_large which needs to be retrained after changing the backbone) 

Now your questions:
- i think it makes more sense to compare the torchvision implementation we use in doctr and the keras one :) 
- trained from mindee ( with charactergenerator - see classification training script ) 
- i would suggest to train these models from scratch (with mindee data) there is already an issue pinned for 0.6.0 #969 

Additional to the backbone change we need to ensure that the above models continue to work well by testing them with toy runs

Let me know what you think :hugs:

felixdittrich92:
Hi @aminemindee, are you still interested to solve this ? :) So we could assign it to you

aminemindee:
> Hi @aminemindee, are you still interested to solve this ? :) So we could assign it to you

Hello @felixdittrich92 i will get back to it once i finish a project i'm working on to make the detection model training part working with multiple classes and then maybe retrain with that so i can't really work on it at the moment.

felixdittrich92:
[...]

PR #1182, opened 2023-04-21
[tests/TF/build] enable missing classification onnx tests and set tensorflow lower bound to 2.11

felixT2K:
This PR:

- upgrades the minor version of tf2onnx to >=1.14 (which supports TF 2.11)
- upgrades the minor version of tensorflow to >= 2.11 (first version where the issue seems to be fixed)
- enables tests

Closes:
#978 and #790 🚀

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,9 +88,9 @@ docs = [
 ]
 dev = [
     # Tensorflow
-    "tensorflow>=2.11.0,<3.0.0",  # cf. https://github.com/mindee/doctr/issues/454
+    "tensorflow>=2.11.0,<3.0.0",  # cf. https://github.com/mindee/doctr/pull/1182
     "tensorflow-addons>=0.17.1",
-    "tf2onnx>=1.14.0",
+    "tf2onnx>=1.14.0,<2.0.0",
     # PyTorch
     "torch>=1.8.0",
     "torchvision>=0.9.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,9 +88,9 @@ docs = [
 ]
 dev = [
     # Tensorflow
-    "tensorflow>=2.9.0,<3.0.0",  # cf. https://github.com/mindee/doctr/issues/454
+    "tensorflow>=2.11.0,<3.0.0",  # cf. https://github.com/mindee/doctr/issues/454
     "tensorflow-addons>=0.17.1",
-    "tf2onnx>=1.9.2",
+    "tf2onnx>=1.14.0",
     # PyTorch
     "torch>=1.8.0",
     "torchvision>=0.9.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,9 +54,9 @@ dependencies = [
 
 [project.optional-dependencies]
 tf = [
-    "tensorflow>=2.11.0,<3.0.0",  # cf. https://github.com/mindee/doctr/issues/454
+    "tensorflow>=2.11.0,<3.0.0",  # cf. https://github.com/mindee/doctr/pull/1182
     "tensorflow-addons>=0.17.1",
-    "tf2onnx>=1.14.0",
+    "tf2onnx>=1.14.0,<2.0.0",
 ]
 torch = [
     "torch>=1.8.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,9 +54,9 @@ dependencies = [
 
 [project.optional-dependencies]
 tf = [
-    "tensorflow>=2.9.0,<3.0.0",  # cf. https://github.com/mindee/doctr/issues/454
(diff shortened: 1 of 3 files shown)


felixdittrich92:
@frgfm Are you fine with increasing the minimum versions to fix this ? I think it would be also the time to drop py3.6 and 3.7 support and focus on 3.9 and 3.10 (especially CI) with >=3.8,<4

odulcy-mindee:
@felixdittrich92 Should it fix:
```
tensorflow.python.framework.errors_impl.InvalidArgumentError: {{function_node __wrapped__Conv2DBackpropInput_device_/job:localhost/replica:0/task:0/device:CPU:0}} Gradients for grouped convolutions are not supported on CPU. Please file a feature request if you run into this issue. Computed input depth 576 doesn't match filter input depth 1 [Op:Conv2DBackpropInput]
```
?

felixT2K:
> @felixdittrich92 Should it fix:
> 
> ```
> tensorflow.python.framework.errors_impl.InvalidArgumentError: {{function_node __wrapped__Conv2DBackpropInput_device_/job:localhost/replica:0/task:0/device:CPU:0}} Gradients for grouped convolutions are not supported on CPU. Please file a feature request if you run into this issue. Computed input depth 576 doesn't match filter input depth 1 [Op:Conv2DBackpropInput]
> ```
> 
> ?

@odulcy-mindee
Unfortunately no, the problem should actually be fixed since TF 2.9. I also tested it locally and it persists. However, only during training on the CPU. Training on the GPU and inference on the CPU and GPU runs without any problems.

Ref.: https://discuss.tensorflow.org/t/error-message-for-grouped-convolution-backprop-on-cpu-is-uninformative/6323

But this PR fixes the failing onnx exports for TF mobilenet. After this we have finally all models available for exporting into onnx format (up to the logits)

felixT2K:
@olivmindee fine with the changes in this PR ? I opened #1183 for the grouped conv issue 👍🏼

frgfm:
> @frgfm Are you fine with increasing the minimum versions to fix this ? I think it would be also the time to drop py3.6 and 3.7 support and focus on 3.9 and 3.10 (especially CI) with >=3.8,<4

Regarding the version specifiers, yes that's alright but for future debugging purposes, I suggest updating the PR description to specify which new features we want to get out of those recent releases (easier to track down problems in a specific version later on).

About the python version, let me open a quick PR to check that (Python version upgrade can sometimes be tricky when the project on large platform-specific builds like DL frameworks) :+1:

pyproject.toml
```diff
@@ -54,9 +54,9 @@ dependencies = [
 
 [project.optional-dependencies]
 tf = [
-    "tensorflow>=2.9.0,<3.0.0",  # cf. https://github.com/mindee/doctr/issues/454
+    "tensorflow>=2.11.0,<3.0.0",  # cf. https://github.com/mindee/doctr/issues/454
```
frgfm, line 57: Open question: should we update the comment to reflect what we aim to fix with that?

I've already found it much more useful than I anticipated to have the references to the multiple issues that led to choosing the version specifiers.

pyproject.toml
```diff
@@ -54,9 +54,9 @@ dependencies = [
...
     "tensorflow-addons>=0.17.1",
-    "tf2onnx>=1.9.2",
+    "tf2onnx>=1.14.0",
```
frgfm, line 59: Same here (if there's any) and perhaps add an upper bound? (<2.0.0 I would suggest)

tests/tensorflow/test_models_classification_tf.py
```diff
@@ -98,10 +98,9 @@ def test_crop_orientation_model(mock_text_box):
         ["resnet50", (32, 32, 3), (126,)],
         ["magc_resnet31", (32, 32, 3), (126,)],
         ["vit_b", (32, 32, 3), (126,)],
-        # Disabled for now
-        # ["mobilenet_v3_small", (512, 512, 3), (126,)],
-        # ["mobilenet_v3_large", (512, 512, 3), (126,)],
-        # ["mobilenet_v3_small_orientation", (128, 128, 3), (4,)],
+        ["mobilenet_v3_small", (512, 512, 3), (126,)],
+        ["mobilenet_v3_large", (512, 512, 3), (126,)],
+        ["mobilenet_v3_small_orientation", (128, 128, 3), (4,)],
```
frgfm, line 103: Finally haha :grin:

frgfm (commented):
Thanks again! Only a few suggestions for later debugging :ok_hand:

pyproject.toml
felixT2K: added ref to PR :)

felixdittrich92 (dismissed):

felixdittrich92 (approved):

PR #1184, opened 2023-04-27
[build] drop py3.6/3.7 support and update CI default to py3.8/3.9

felixT2K:
This PR:

- drops support for python 3.6 and 3.7 (latest packages mostly have a mimum python version from 3.8 for example pytorch 2.0)
- update CI also to be up to date

Any feedback is welcome :)

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest]
         python: ["3.8", "3.10"]
         framework: [tensorflow, pytorch]
     steps:
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -17,9 +17,9 @@ jobs:
         framework: [tensorflow, pytorch]
     steps:
       - uses: actions/checkout@v2
-      - if: matrix.os == 'macos-latest'
-        name: Install MacOS prerequisites
-        run: brew install cairo pango gdk-pixbuf libffi
+      #- if: matrix.os == 'macos-latest'
+        #name: Install MacOS prerequisites
+        #run: brew install cairo pango gdk-pixbuf libffi
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python: [3.7, 3.8]
+        python: ["3.8", "3.10"]
         framework: [tensorflow, pytorch]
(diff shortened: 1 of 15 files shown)


felixdittrich92:
@charlesmindee If we want to merge it you should disable the required checks in the GitHub actions merge it and enable the same updated checks again as required (you are the only one with repo Admin rights) :)

.github/workflows/builds.yml
```diff
@@ -13,27 +13,27 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python: [3.7, 3.8]
+        python: ["3.8", "3.10"]
```
frgfm, line 16: I suggest testing consecutive versions instead. I see two reasons behind this:
- for another project, I recently took a look at the most used python versions, and 3.8 was the top one. Adoption is generally smooth, meaning there are way more people using 3.9 than 3.10.
- once we drop support of 3.8, it would be nice to already have a history of test runs on 3.9

.github/workflows/demo.yml
```diff
@@ -13,27 +13,27 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python: [3.8]
+        python: ["3.10"]
```
frgfm, line 16: Adding tests for multiple version is one thing, but here we cannot guarantee anymore that this works for python 3.8. I'd suggest keeping 3.8 :)

.github/workflows/docker.yml
frgfm, line 1: Same here

pyproject.toml
```diff
@@ -59,8 +59,9 @@ tf = [
     "tf2onnx>=1.9.2",
 ]
 torch = [
-    "torch>=1.8.0",
-    "torchvision>=0.9.0",
+    "torch>=1.12.0",
+    "torchvision>=0.13.0",
+    "onnx>=1.12.0"
```
frgfm, line 64: I suggest not modifying lib version specifiers in this PR

frgfm (commented):
Thanks, I was about to open a PR haha
I added a few comments but here are my high-level thoughts:
- dropping support of 3.6 & 3.7, and adding 3.9 & 3.10 is something
- moving the main version support from 3.8 to 3.10 is another
- I'd suggest staying on 3.8 which is still the most used python version (at least a few months back)

Best !

.github/workflows/demo.yml
felixT2K: yeah i agree in this case .. :) 👍🏼

pyproject.toml
felixT2K: moved to other PR

felixT2K:
@frgfm i personally prefer especially for tests to have 2er steps (3.8 | 3.10 | 3.12 ..) but i am also fine to test it always on the most used python version :)  

But all tests was also running with python 3.10 so good to know :)

Now it is up to @charlesmindee :)

odulcy-mindee (approved):
Thanks @felixdittrich92 !!

PR #1201, opened 2023-05-24
[CI] change old cache action and skip TF classification onnx export temporarily

felixT2K:
This PR:

- update last missing cache action
- remove fixed opset version which are not more needed after changing the minimum pt / tf version)
- temporarily skip on CI for TF classification onnx export (tf2onnx 1.14 seems to raise a memory leak which kills the CI machine) - locally all tests passes for me, but yeah the tf resnet >= 31 now takes ~16GB RAM while converting 
Not sure if it comes from TF itself or tf2onnx

Maybe setting up a larger runner: @charlesmindee @odulcy-mindee (or we skip more tf onnx tests ..ftm)
https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners

Any feedback is welcome 🤗

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -87,7 +87,7 @@ jobs:
           python-version: ${{ matrix.python }}
           architecture: x64
       - name: Cache python modules
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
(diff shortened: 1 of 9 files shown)


odulcy-mindee (approved):
LGTM, thanks Felix !

PR #1203, opened 2023-05-29
Arm m1 gpu

ArthurKantor:
* Puts the detection and selection of GPU availablebility into one place.
* adds support for arm mps in addition to the existing cpu and cuda hardware.


ArthurKantor:
wrong repo - sorry

PR #1234, opened 2023-06-29
[docs] update benchmark page

felixdittrich92:
This PR:

- updates models/benchmark page
- add benchmarks for new available models
- set aspect ratio to False to avoid eval script changes atm
- update onnx also in tf build which was already added for the dev build


felixT2K:
TODO: 

- update half precsion docs (only full support on gpu)

ArsalanYounus007:
Can you please add MS OCR to the comparison table as well?

felixT2K:
> Can you please add MS OCR to the comparison table as well?

Hey 👋🏼 ,
you mean the Azure Vision API (OCR) ?
I will check this 👍🏼

felixdittrich92:
> Can you please add MS OCR to the comparison table as well?

Added benchmark for Azure form recognizer (v3.2). I have tested the standard OCR API also but this performs really poor on document types (because it's made for scene text in the wild) so i decided to add only the form recognizer benchmark to be fair :)

ArsalanYounus007:
Thank you for considering MS Azure Read API.
I am a little confused about "this performs really poor on document types" because we use this for many projects and it performs very well on scanned document images.

Just to confirm, You tried using Read API ?? I have shared the URL that we use.
https://learn.microsoft.com/en-us/azure/ai-services/document-intelligence/concept-read?view=doc-intel-3.1.0

felixdittrich92:
> Thank you for considering MS Azure Read API. I am a little confused about "this performs really poor on document types" because we use this for many projects and it performs very well on scanned document images.
> 
> Just to confirm, You tried using Read API ?? I have shared the URL that we use. https://learn.microsoft.com/en-us/azure/ai-services/document-intelligence/concept-read?view=doc-intel-3.1.0

The added benchmark is from document intelligence and before i have tested the vision api ocr endpoint

PR #1237, opened 2023-07-03
chore: apply PIL major changes and increase min version specifier

felixT2K:
This PR:

- increases the PIL min version to >= 10.0.0
- fix for the breaking change


felixT2K:
merged without further review because: critical and minor change

felixdittrich92 (approved):

Issue #1207, opened 2023-05-30
issue when using with multiprocessing


msaoudallah:
### Bug description

i have an issue which is i need to use `doctr` whithin a subprocess so that memory is released when the prossed is finished. my problem is when i start the inference from the PDF i get the following output 
![image](https://github.com/mindee/doctr/assets/6372325/cc9cf1d2-fa34-46b5-8f5c-66ca976bd4fe)

i have a flask server and when i send the first request to be processed this happens, but the subsequent requests complete successfully 
another note is i used this application within docker container and i don't have GPU 

 

### Code snippet to reproduce the bug

```python
with concurrent.futures.ProcessPoolExecutor(max_workers =8) as executor:
                    result = executor.submit(self.get_xml_custom, fileobj).result()
```
please note that this line is executed from parent class and the `self.get_xml_custom` is excuted from child class

### Error traceback

no errors just this line 

```
free(): invalid pointer
```

### Environment

DocTR version: v0.6.0
TensorFlow version: N/A
PyTorch version: 1.10.0+rocm4.2 (torchvision 0.11.1+rocm4.2)
OpenCV version: 4.6.0
OS: Debian GNU/Linux 11 (bullseye)
Python version: 3.8.16
Is CUDA available (TensorFlow): N/A
Is CUDA available (PyTorch): No
CUDA runtime version: Could not collect
GPU models and configuration: Could not collect
Nvidia driver version: Could not collect
cuDNN version: Could not collect

### Deep Learning backend

is_tf_available: False
is_torch_available: True

felixdittrich92:
Hi @msaoudallah 👋,

Please try to set a env var to:
`DOCTR_MULTIPROCESSING_DISABLE=TRUE`

msaoudallah:
hello @felixdittrich92 ,
can you please provide an example of how this would work ?  and would it help releasing memory after execution finish ?

felixdittrich92:
Hey @msaoudallah 👋,
You can put `os.environ['DOCTR_MULTIPROCESSING_DISABLE'] = 'TRUE'` on top of your code. :)
If i remember it has solved a possible memory leak also but i am not 100% sure :)

felixdittrich92:
Hi @msaoudallah 👋,
has this solved your issue ? :)

Srini138:
[...]

PR #1239, opened 2023-07-04
[chore]: Pypdfium2 compatibility fix

felixT2K:
This PR:

- applies changes to be compatible with pypdfium2 >= 4.0.0


felixdittrich92:
@mara004 temp fix for compatibility

doctr/io/pdf.py
```diff
@@ -31,16 +30,12 @@ def read_pdf(
...
 
-    if isinstance(file, Path):
-        file = str(file)
-
     # Rasterise pages to numpy ndarrays with pypdfium2
-    pdf = pdfium.PdfDocument(file, password=password)
-    renderer = pdf.render_to(pdfium.BitmapConv.numpy_ndarray, scale=scale, rev_byteorder=rgb_mode, **kwargs)
-    return [img for img, _ in renderer]
+    pdf = pdfium.PdfDocument(file, password=password, autoclose=True)
+    return [page.render(scale=scale, rev_byteorder=rgb_mode, **kwargs).to_numpy() for page in pdf]
```
mara004, line 41: No, please don't do this. This will result in a major slowdown compared to the multi-page renderer, as mentioned in the other thread.

mara004: just write a tempfile, as I already said, no?

felixdittrich92: Ah sry i missed this part in your comment 🙈 do you know how big the impact is ? I think there shouldn't be much difference from current usage !?
I would really don't like to start with tricks like this just to ensure compatibility if we probably rewrite the io module after the next release one way or another.

mara004: No problem. The performance advantage depends on use case and host system: If `len(pdf) >= n_processes`, the multi-page renderer will be factor `n_processes` faster.

If you only use it with single-page PDFs, then the multi-page renderer is actually slower ATM. (It should really handle that as special case and render the one page directly instead of setting up a process pool. I wanted to change that a while ago but got distracted by other tasks...)

mara004: > I would really don't like to start with tricks like this just to ensure compatibility if we probably rewrite the io module after the next release one way or another.

Yeah, I understand. Shouldn't be much code, though. Just an if check to save bytes into a file and re-define input to the file name.

FWIW, you can also just stay on v3 until we get to the rewrite.

felixdittrich92: Sure personally i would try to avoid any tempfile usage inside librarys. The problem is some users have mentioned memory leaks while using doctr with pdf inputs (i don't have ad the time to profile it) but if there was an issue with closing your new autoclose arg should fix it (maybe).

felixdittrich92: @odulcy-mindee @frgfm wdyt ?

mara004: > but if there was an issue with closing your new autoclose arg should fix it (maybe).

An issue with closing causing a memory leak? I don't think so, leastways not in the code that affects doctr.
Also I don't think there's a new autoclose argument?

However, it is possible that whatever is wrong with parallel bytes rendering might have caused other corruption / memory leaks...

mara004: > Sure personally i would try to avoid any tempfile usage inside librarys.

Yes, in general I fully agree, but in this case it's kinda different since multiprocessing is involved.
Duplicating a large byte stream with each worker process isn't good either. A tempfile should actually be a performance/memory advantage here, regardless of the bug.

Maybe pypdfium2 could use shared memory here, though. Theoretically that might be even better, but I think it's a difficult topic...

felixdittrich92: Mh `autoclose=True` i don't remember if it was available in 3.0.0 ^^

mara004: Yes I think it was, and this is only to have pypdfium2 close buffer input when the document handle is closed. Sorry about the ambiguity.

mara004: BTW the memory leaks issue is #1207, right?

felixdittrich92: > BTW the memory leak issue is #1207, right?

Mh yes and #1208 (same author)

felixdittrich92: I would now wait for another opinion. 
To ensure compatibility, I would take the version upgrade in any case, since we will probably not start with the refactoring before the end of this year. Which is why now the question would be open whether we want to leave it at the current or if the input bytes is about tempfile (as mara has suggested) want to go !? 🙂

mara004: > I would now wait for another opinion.

Yeah, I'd appreciate in any case if you could delay this a bit so I have time to investigate the shared memory approach first (and also update the document-level renderer to handle single pages).

felixdittrich92: Sounds good to me :)

mara004: I did some linear vs parallel rendering experiments today. The results are highly astonishing.
It looks like you're right and I was wrong all along: performance seems to be almost equal, with linear rendering actually slightly faster according to my preliminary testing. Apparently the overhead caused by process management and data serialization must be huge...

Linear rendering is stable with all input types.

Parallel rendering only works with paths, anything else either crashes or skyrockets memory load, eventually freezing the system. And even with paths memory consumption is elevated compared to linear rendering.

I'm confused; but evidently this is the reality. If I haven't overlooked anything, I guess you may well go ahead with this PR. Maybe I should deprecate the process pool based renderer. Lol.

mara004: I could still dive into shared memory some day, but am not sure if it's worth it in this light...

felixdittrich92: Hey @mara004 👋 ,
Thanks for sharing your findings. 
Ok that was just a guess on my part, as I noticed similar behavior in another context before. 😅 
But nice if you can take some insight from the test for your library as well. 👍

felixdittrich92: I think too. Until we go ahead with generators we can keep this as robust solution for the moment. :)

mara004: Simply using linear rendering should be OK for doctr for now.

However, it turns out the situation is more complicated for pypdfium2's CLI.
I figured the problem was that we had a fairly expensive PIL save call in the main process, not in workers, which meant the process pool queued up results in memory without limit while the generator in the main process lacked behind. (That also may be something to take into account for the upcoming refactoring of doctr...)
Now when including the save call in workers parallelization is still a major advantage for our CLI.

However, I don't know yet how parallel vs linear would compare with rendering only (i.e. no post-processing and no bitmap transfer).

odulcy-mindee (approved):

PR #1246, opened 2023-07-08
docs: :pencil2: fix typos in documentation

odulcy:


docs/README.md
felixdittrich92, line 1: Can we move this inside https://github.com/mindee/doctr/blob/main/CONTRIBUTING.md please :)

docs/README.md
```diff
@@ -0,0 +1,22 @@
...
+```bash
+# Make sure you are at the root of the repository before executing these commands
+python -m pip install --upgrade pip
+pip install -e .[tf]
+pip install -e .[docs]
+```
+
+Now you can build the documentation by running:
+```bash
+make html
```
felixdittrich92, line 15: We don't have `make html` it's `make docs-single-page`

docs/source/using_doctr/using_models.rst
```diff
@@ -330,14 +330,18 @@ For reference, here is the JSON export for the same `Document` as above::
...
 
   xml_output = result.export_as_xml()
   for output in xml_output:
       xml_bytes_string = output[0]
       xml_element = output[1]
 
-For reference, here is a sample XML byte string output::
+For reference, here is a sample XML byte string output:
+
+.. code-block:: xml
```
felixdittrich92, line 344: Does this render correctly ?

docs/README.md
```diff
@@ -0,0 +1,22 @@
...
+```
+
+Afterwards, you can view the documentation by opening `_build/html/index.html` in your web browser.
+
+## Modifying Documentation
+
+The DocTR documentation is built using `sphinx`. If you make changes to a file, you need to run `make html` again. Please note that files that have not been modified will not be rebuilt. If you want to force a complete rebuild, you can delete the `_build` directory. Additionally, you may need to clear your web browser's cache to see the modifications.
```
felixdittrich92, line 22: DocTR -> docTR :)

felixdittrich92 (changes requested):

docs/README.md
odulcy: What do you mean ? `make html` works on my side :sweat_smile:

docs/README.md
odulcy: Oh yeah, sure

docs/source/using_doctr/using_models.rst
odulcy: Yep, but maybe it's due to `make html`:
image

docs/README.md
odulcy: :see_no_evil:

docs/source/using_doctr/using_models.rst
felixdittrich92: Nice 👍

docs/README.md
felixdittrich92: Mh we don't have a command html in Makefile 😅

felixdittrich92: How can this work xD docs-single-version builds only the actual docs (for testing)

odulcy: Because we have the following `Makefile`:
```
# Catch-all target: route all unknown targets to Sphinx using the new
# "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
%: Makefile
	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```

`html` is forwarded to `sphinx-build` where `html` is a subcommand, that's why it works

felixdittrich92: ok then we should decide on a command and adapt the Makefile accordingly 👍

odulcy-mindee: Ok, so I just checked both command. `make html` invokes `sphinx-build` with builder named `html`: https://www.sphinx-doc.org/en/master/man/sphinx-build.html.
But `html` is the default builder, so `make html` here or `make docs-single-version` at root repository produces exactly the same thing. So let's keep `make docs-single-version`

tests/common/test_utils_geometry.py
```diff
@@ -231,7 +231,7 @@ def test_extract_rcrops(mock_pdf):  # noqa: F811
     abs_boxes = deepcopy(rel_boxes)
     abs_boxes[:, :, 0] *= doc_img.shape[1]
     abs_boxes[:, :, 1] *= doc_img.shape[0]
-    abs_boxes = abs_boxes.astype(np.int)
+    abs_boxes = abs_boxes.astype(np.int64)
```
felixT2K, line 234: can be reverted only rebase on main branch :)

felixT2K (changes requested):

PR #1252, opened 2023-07-17
[chore]: Replace `tensorflow_addons` 

felixdittrich92:
This PR:

- replaces `tensorflow_addons` which is end of life (May 2024) --> https://github.com/tensorflow/addons/issues/2807
- add hotfix for docs build
- maybe some space for improvements

Any feedback is welcome :)

### DIFF
--- a/doctr/models/classification/vit/tensorflow.py
+++ b/doctr/models/classification/vit/tensorflow.py
@@ -8,7 +8,6 @@ from typing import Any, Dict, Optional, Tuple
 
 import tensorflow as tf
 from tensorflow.keras import Sequential, layers
-from tensorflow.keras.activations import gelu
 
 from doctr.datasets import VOCABS
 from doctr.models.modules.transformer import EncoderBlock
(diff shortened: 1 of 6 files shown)


doctr/models/classification/vit/tensorflow.py
```diff
@@ -38,6 +38,17 @@
...
 
+class GELU(layers.Layer):
+    """Gaussian Error Linear Unit activation function"""
+
+    def __init__(self, approximate: bool = False, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self.approximate = approximate
+
+    def call(self, x: tf.Tensor) -> tf.Tensor:
+        return gelu(x, approximate=self.approximate)
```
odulcy-mindee, line 49: Using `https://www.tensorflow.org/api_docs/python/tf/keras/layers/Activation` with `tf.nn.gelu` should work, no ?

felixdittrich92: Good catch i will test it but it should work

doctr/transforms/modules/tensorflow.py
```diff
@@ -385,11 +384,14 @@ def extra_repr(self) -> str:
...
-        sigma = random.uniform(self.std[0], self.std[1])
-        return tfa.image.gaussian_filter2d(
-            img,
-            filter_shape=self.kernel_shape,
-            sigma=sigma,
+        return tf.squeeze(
+            _gaussian_filter(
+                img[tf.newaxis, ...],
+                kernel_size=self.kernel_shape,
+                sigma=np.random.uniform(self.std[0], self.std[1]),
```
odulcy-mindee, line 391: For performance reason, it would be better to keep `random.uniform` if you sample only one number

felixdittrich92: But random.uniform is tracked as possible security issue (codacity log)

felixT2K: https://github.com/mindee/doctr/runs/15116301840

felixdittrich92:
@odulcy-mindee wdyt ? :)

odulcy-mindee (approved):

PR #1254, opened 2023-07-18
fix: :bug: fix bug when training object detection

aminemindee:
- Small quick fix to make training object detection working again.


felixdittrich92:
Thanks for the PR @aminemindee :)
Can we maybe add a short comment above the condition that it's only the case for a detection KIE dataset ?

BTW: `file_utils.py` seems not the right place for `copy_tensor` can we move it to `https://github.com/mindee/doctr/tree/main/doctr/models/utils`  and add a short test ? (Because it's small but uncovered currently)

odulcy-mindee (approved):

Issue #1256, opened 2023-07-19
[Docs] Add a Page for training and loading custom models


felixdittrich92:
### 🚀 The feature

Add a documentation page which links to references Readme's and add code snippets for TF and PT how to load custom trained models 

snippets should show also:  

- passing correct vocab while loading custom model
- passing correct class_names while loading (if KIE)
- disabed pretrained and pretrained_backbone to prevent download

### Motivation, pitch

A lot of people ask how to train their own models and then load them.
We should add a page in the documentation linking to the README's in the references folder.

In addition, we should deposit code examples (for TF and PT) how to load your trained model into docTR.
It should also be clearly evident that the vocab used must be transferred to the model when loading it if it deviates from the default config. We should do the same for KIE, where you have to give the class names to the model when loading.

pretraind + pretrained_backbone should be false in this case to prevent the download

PR #1257, opened 2023-07-20
feat: :sparkles: ClearML support for TensorFlow

odulcy-mindee:
This PR introduces [ClearML](https://clear.ml/pricing/) support in references script, for TensorFlow only (PyTorch will be done in another PR).

At the moment, ClearML captures **everything**, meaning that logs, model checkpoints, etc. are tracked. Do we want to keep it like that or do we want to block everything [auto_connect_frameworks](https://clear.ml/docs/latest/docs/references/sdk/task/#taskinit) in `Task.init` by default ? If so, how do we expose parameters to users to tune this behavior ?

### DIFF
--- a/references/detection/train_tensorflow.py
+++ b/references/detection/train_tensorflow.py
@@ -313,7 +313,7 @@ def main(args):
     if args.clearml:
         from clearml import Task
 
-        task = Task.init(project_name="docTR/text-detection", task_name=exp_name)
+        task = Task.init(project_name="docTR/text-detection", task_name=exp_name, reuse_last_task_id=False)
         task.upload_artifact("config", config)
 
     if args.freeze_backbone:
--- a/references/detection/train_tensorflow.py
+++ b/references/detection/train_tensorflow.py
@@ -354,10 +354,10 @@ def main(args):
             from clearml import Logger
 
             logger = Logger.current_logger()
-            logger.report_single_value(name="val_loss", value=val_loss)
-            logger.report_single_value(name="recall", value=recall)
-            logger.report_single_value(name="precision", value=precision)
-            logger.report_single_value(name="mean_iou", value=mean_iou)
+            logger.report_scalar(title="Validation Loss", series="val_loss", value=val_loss, iteration=epoch)
+            logger.report_scalar(title="Precision Recall", series="recall", value=recall, iteration=epoch)
+            logger.report_scalar(title="Precision Recall", series="precision", value=precision, iteration=epoch)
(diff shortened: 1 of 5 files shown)


references/detection/train_tensorflow.py
```diff
@@ -359,9 +374,9 @@ def parse_args():
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
     )
 
-    parser.add_argument("train_path", type=str, help="path to training data folder")
-    parser.add_argument("val_path", type=str, help="path to validation data folder")
     parser.add_argument("arch", type=str, help="text-detection model to train")
+    parser.add_argument("--train_path", type=str, help="path to training data folder")
+    parser.add_argument("--val_path", type=str, help="path to validation data folder")
```
odulcy-mindee, line 379: I changed that so it's the same options than `recognition`. I'll do the same for PyTorch in another PR.

felixdittrich92 (dismissed):
Thanks a lot @odulcy-mindee 🙏
We should keep the default config which WE use already for wandb. For specific use cases the scripts are open to be modified by users :)

felixdittrich92:
@odulcy-mindee three small things:

For detection it should be `train_path` instead of `--train_path` because it's required we have no other option like the generators for classicication and recognition :)

Point 2: 
Can we import wandb also after the condition ?

Point 3: 
weight decay should be configurable 

:)

references/detection/train_tensorflow.py
felixdittrich92: See comment :)

odulcy-mindee:
@felixdittrich92  
> Point 1: For detection it should be `train_path` instead of `--train_path` because it's required we have no other option like the generators for classicication and recognition :)
Oh yeah, I just wanted to have same way to provide the dataset. I can revert or make it mandatory with `¶equired=True`

> Point 2: Can we import wandb also after the condition ?
Sure !

> Point 3: weight decay should be configurable (or removed from the config If unused) ?
Ok !

felixdittrich92:
@odulcy-mindee 
I would be fine with `required=True` 👍

odulcy-mindee:
@felixdittrich92 I removed `weight-decay` (it was unused) and added name to the scheduler, tell me what you think about it

felixdittrich92 (approved):
Thanks LGTM now :)
Only one question does clear ml also ask for login like wandb ?

felixdittrich92:
Otherwise we should add a short reference but we can do this in #1256

odulcy-mindee:
@felixdittrich92 it looks for a `clearml.conf` in home. If the file does not exist, clearml raises an error and provides a link to doc to do the setup :+1:

PR #1258, opened 2023-07-20
[quality] move copy tensor function

felixT2K:
This PR:

- moves `copy_tensor` from `file_utils.py` to model/ùtils.py`


odulcy-mindee (approved):
Thanks!

PR #1263, opened 2023-07-24
[docs] custom model load

felixdittrich92:
This PR:

- Add page in docs with a ref to references README's for custom training and how to load custom models (which is the most frequent asked question :sweat_smile:)


odulcy-mindee (approved):
Thanks, that's a good idea !

PR #1264, opened 2023-07-24
[chore] increase version and update maintainers

felixT2K:
This PR:

- increase version to 0.7.0a0 (next release will be minor and not patch)
- add Olivier and myself  😅 also to the list of maintainers

I will cleanup the issue trackers in the next days 👍🏼 

Any feedback is welcome 🤗

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,6 +9,8 @@ authors = [{name = "Mindee", email = "contact@mindee.com"}]
 maintainers = [
     {name = "François-Guillaume Fernandez"},
     {name = "Charles Gaillard"},
+    {name = "Olivier Dulcy"},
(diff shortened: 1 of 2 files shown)


felixdittrich92:
ok issue trackers are updated (for 0.8.0 we can decide later what we want to keep or maybe move to a later release)

pyproject.toml
odulcy-mindee, line 1: Haha thanks

setup.py
odulcy-mindee, line 1: I need to discuss that with the others internally

felixdittrich92: @odulcy-mindee how does it look ? :)

odulcy-mindee (approved):

PR #1267, opened 2023-07-24
[Fix] fix obj det train and suppress endless warning prints

felixdittrich92:
This PR:

- fix object detection training also with pretrained (change to torchvision's new weights loading -> pretrained is deprecated)
- fix the endless warnings from each torchvision resize operation (warning that `antialias` will change by default to `True` in v0.17)

Any feedback is welcome

### DIFF
--- a/doctr/transforms/modules/pytorch.py
+++ b/doctr/transforms/modules/pytorch.py
@@ -26,7 +26,7 @@ class Resize(T.Resize):
         preserve_aspect_ratio: bool = False,
         symmetric_pad: bool = False,
     ) -> None:
-        super().__init__(size, interpolation)
+        super().__init__(size, interpolation, antialias=True)
         self.preserve_aspect_ratio = preserve_aspect_ratio
         self.symmetric_pad = symmetric_pad
 
@@ -64,7 +64,7 @@ class Resize(T.Resize):
                     tmp_size = (self.size, max(int(self.size / actual_ratio), 1))
 
             # Scale image
-            img = F.resize(img, tmp_size, self.interpolation)
+            img = F.resize(img, tmp_size, self.interpolation, antialias=True)
             raw_shape = img.shape[-2:]
             if isinstance(self.size, (tuple, list)):
                 # Pad (inverted in pytorch)
--- a/doctr/models/obj_detection/faster_rcnn/pytorch.py
+++ b/doctr/models/obj_detection/faster_rcnn/pytorch.py
@@ -5,7 +5,7 @@
 
 from typing import Any, Dict
 
-from torchvision.models.detection import FasterRCNN, faster_rcnn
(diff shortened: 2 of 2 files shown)


odulcy-mindee (approved):

doctr/transforms/modules/pytorch.py
```diff
@@ -26,7 +26,7 @@ def __init__(
         preserve_aspect_ratio: bool = False,
         symmetric_pad: bool = False,
     ) -> None:
-        super().__init__(size, interpolation)
+        super().__init__(size, interpolation, antialias=True)
```
tvasenin, line 29: This change of `antialias` parameter makes detection of single-digit words consistently worse in some documents in our dataset.

@felixdittrich92 Is it possible to change this back to `antialias=False` (at least until new models are trained with this set to `True`) or at least make it configurable?

felixT2K: Hi @tvasenin 👋🏼,

Thanks for the feedback.
Are you sure that the problem is antialising ?
We have changed the default value from `preserve_aspect_ratio=False` to `True` see https://github.com/mindee/doctr/releases/tag/v0.7.0

I think reverting it is not an option because in torch under the hood Pillow will always use antialising.
Making it configurable if you use a custom preprocessor would be possible.
But i suggest to test it in front of this change with the old behaviour  `ocr_predictor(...., preserve_aspect_ratio=False)`

felixT2K: @odulcy-mindee What do you think ?
I would prefer to update (resume) the "old" models instead of reverting this because we have also added some more augmentations in the pipeline which should make the models more robust.
@tvasenin FYI we started already with the training period :)

odulcy-mindee: @felixT2K Yeah, I also prefer to udpate the "old" models as you said!

tvasenin: @felixT2K @odulcy-mindee Thanks for the update, waiting for the new release with models trained with the new `antialias` value!

As for release `v0.7.0`, while `preserve_aspect_ratio=True` indeed improves results, changing `antialias` value for tensor resize (both affected resize operations in the current code are tensor-only) breaks the behavior for the models shipped with this release.
As a workaround, we had to downgrade to the previous commit 0a47726bdcdc3fc2237a770fed9839f668cb2d39 and thus can wait for the updated models (though we'd still prefer temporary revert and have `v0.7.1` released as a hotfix).

tvasenin: > FYI we started already with the training period :)

@felixT2K Any updates regarding re-trained models with new `antialias` value?

felixdittrich92: @tvasenin mostly done -> https://github.com/mindee/doctr/issues/1364  :)

Issue #783, opened 2022-01-03
Images not displayed on the PyPi page.


bashirmindee:
### Bug description

The images concerning Visualization sample, Synthesis sample, Demo app and TFJS demo aren't displayed on the PyPi page.
![image](https://user-images.githubusercontent.com/91077790/147921168-46217bc4-bb70-4295-afac-a34d7b868388.png)

However they are well displayed in the Github readme.

### Code snippet to reproduce the bug

.

### Error traceback

.

### Environment

google chrome 96.0.4664.110 linux ubuntu 20.04

fg-mindee:
Thanks for reporting this @bashirmindee :pray: 
I've already seen this on pypi, and apparently, it's a common issue. Here is a StackOverflow thread about this: https://stackoverflow.com/questions/41983209/how-do-i-add-images-to-a-pypi-readme-that-works-on-github

fg-mindee:
The reason behind this seems to be because we used release attachments for the picture, and thus we can't use the `raw` URL described in the StackOverFlow thread. I'll check whether there is something to do about this, without having to add the pictures in the git history!

felixdittrich92:
@odulcy-mindee Looks like my fix was useless unfortunately :sweat_smile:

PR #1268, opened 2023-07-24
[PyPi] Fix image display

felixdittrich92:
This PR:

- should fix the displayed images on https://pypi.org/project/python-doctr/  after the next release

#783


felixT2K:
@odulcy-mindee Another option would be that you upload the images on your file server and we replace the urls.

odulcy-mindee (approved):
@felixdittrich92 it's good enough, thanks for the fix!

PR #1271, opened 2023-07-27
[Fix] add ignore keys if classes differ - KIE training

felixdittrich92:
This PR:

- make it possible to use `--pretrained` if you use a pretrained det model and train on a KIE dataset
- does nothing change for "normal" det training and inference
- same way we have handled it for recognition on differing vocabs

### DIFF
--- a/doctr/models/detection/differentiable_binarization/pytorch.py
+++ b/doctr/models/detection/differentiable_binarization/pytorch.py
@@ -289,6 +289,7 @@ def _dbnet(
     fpn_layers: List[str],
     backbone_submodule: Optional[str] = None,
     pretrained_backbone: bool = True,
+    ignore_keys: Optional[List[str]] = None,
     **kwargs: Any,
 ) -> DBNet:
     # Starting with Imagenet pretrained params introduces some NaNs in layer3 & layer4 of resnet50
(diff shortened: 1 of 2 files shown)


doctr/models/detection/differentiable_binarization/pytorch.py
```diff
@@ -312,7 +313,12 @@ def _dbnet(
     model = DBNet(feat_extractor, cfg=default_cfgs[arch], **kwargs)
     # Load pretrained parameters
     if pretrained:
-        load_pretrained_params(model, default_cfgs[arch]["url"])
+        # The number of class_names is not the same as the number of classes in the pretrained model =>
+        # remove the layer weights
```
odulcy-mindee, line 317: Can you elaborate ?

felixT2K: Yes, just like with the recognition models, if we want to fine-tune a pre-trained model on another vocab we have to reset (resp. resize embeddings) the classifier head as well as the embeddings, otherwise it leads to a shape mismatch. 
We have the same thing now when training on a KIE data set, since the number of classes varies, before that in normal detection training it is always one class (text) so it didn't matter.

felixT2K: This will reset the Conv"head" which makes it also possible to use '--pretrained' for KIE detection training without an impact on the normal detection training and inference otherwise '--pretrained' raises a shape mismatch while init the model with the pretrained checkpoint

felixT2K: as example for crnn:

```
if pretrained:
        # The number of classes is not the same as the number of classes in the pretrained model =>
        # remove the last layer weights
        _ignore_keys = ignore_keys if _cfg["vocab"] != default_cfgs[arch]["vocab"] else None
        load_pretrained_params(model, _cfg["url"], ignore_keys=_ignore_keys)
        
 ignore_keys=["linear.weight", "linear.bias"],
 ```

PR #1279, opened 2023-08-10
[predictor] aspect ratio true by default

felixdittrich92:
This PR:

- change keeping the aspect ratio by default to True which will especially on images which are smaller as the default 1024x1024 improve the results

Ref.: #330  (already discussed with @charlesmindee)


tests/tensorflow/test_models_zoo_tf.py
odulcy-mindee, line 1: Maybe add a similar test with `preserve_aspect_ratio=True` please, wdyt

odulcy-mindee (commented):
Fine to modify the default value

tests/tensorflow/test_models_zoo_tf.py
felixdittrich92: Now we have unified tests for TF and PT :)

odulcy-mindee (approved):
Thank you very much ! :rocket:

Issue #11, opened 2021-01-12
[docs] Add documentation building dependencies


fg-mindee:
Some basic documentation with proper installation and usage installations would be greatly beneficial to the library. This would be the privileged mean of communication with non-developer audiences.

Having it being automatically built using something similar to sphinx would be efficient considering all docstring will have a compatible format.

PR #1287, opened 2023-08-25
[chore]: clean dependencies

felixT2K:
This PR:

- updates API dependencies
- remove upper bounds for dev dependencies
- increase minimum tf2onnx version to fix the incompatibility with >tf2.12

### DIFF
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -17,10 +17,6 @@ python-doctr = {git = "https://github.com/mindee/doctr.git", extras = ['tf'], br
 # cf. https://github.com/tiangolo/fastapi/issues/4168
 fastapi = ">=0.73.0"
 uvicorn = ">=0.11.1"
-#python-multipart = "==0.0.5"
-# Protobuf: pin version otherwise errors
-# See https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates
-#protobuf= "<=3.20.2"
 
 [tool.poetry.dev-dependencies]
 pytest = ">=5.3.2"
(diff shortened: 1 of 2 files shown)


odulcy-mindee (commented):
I don't know why it passes the tests but [there is an incompatibility between some packages](https://github.com/mindee/doctr/actions/runs/5995836517/job/16259453924?pr=1287):

```
...
#11 137.0 ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
#11 137.0 cleo 2.0.1 requires rapidfuzz<3.0.0,>=2.2.0, but you have rapidfuzz 3.2.0 which is incompatible.
...
```

See `Build & run docker` steps

felixT2K:
> I don't know why it passed the tests but [there is an incompatibility between some packages](https://github.com/mindee/doctr/actions/runs/5995836517/job/16259453924?pr=1287):
> 
> ```
> ...
> #11 137.0 ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
> #11 137.0 cleo 2.0.1 requires rapidfuzz<3.0.0,>=2.2.0, but you have rapidfuzz 3.2.0 which is incompatible.
> ...
> ```
> 
> See `Build & run docker` steps

Yeah i saw this also but it runs locally inside docker also without issues ... i think we override repidfuzz with our >=3.0.0 version and it seems not to break anything inside cleo or there is no use of it

felixT2K:
@odulcy-mindee checked again it overrides it:  rapidfuzz-3.2.0   is installed and all works 😅

felixT2K:
cleo lib ref: https://github.com/python-poetry/cleo/commit/57c6b9ee79164c099aed385413566ac3f3d2af4e (as i wrote on LinkedIn ^^)

odulcy-mindee (approved):

PR #1288, opened 2023-08-25
Change in way how model is saved during Distributed Data Parallel

venkatapathy:
There needs a slight change for saving a model during DistributedDataParallel. Check the out the following links:
1. https://stackoverflow.com/questions/61037819/pytorch-how-to-save-and-load-model-from-distributeddataparallel-learning
2. https://stackoverflow.com/questions/68386282/error-loading-model-saved-with-distributed-data-parallel
3. https://discuss.pytorch.org/t/how-to-save-a-single-final-model-from-distributed-multi-gpu-training/130983

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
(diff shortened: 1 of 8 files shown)


venkatapathy:
I just want the changeindpp branch to be committed

PR #1310, opened 2023-09-11
docs: :pencil2: fix images on pypi

odulcy-mindee:
Attempt to fix https://github.com/mindee/doctr/issues/783

@felixdittrich92 Unfortunately, your previous fix failed because it's a link to an asset in a GitHub Release, which behaves differently than `https://raw.githubusercontent.com/...`. GitHub Release seems to always return an answer with `content-type=application/octet-stream`, which fools PyPi. Let's try by putting the file directly in the repo.

~~Also, I re-added `pip install` at the top of the README.md, I think it makes sense to have it on top to show it's (now) easy to install an up-to-date version of docTR. This point is open to discussion, of course.~~


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -73,4 +73,4 @@ body:
 - type: markdown
   attributes:
     value: >
-      Thanks for helping us improve the library!
```
felixT2K, line 76: please revert :)

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -30,4 +30,4 @@ body:
 - type: markdown
   attributes:
     value: >
-      Thanks for contributing 🎉
```
felixT2K, line 33: same

README.md
```diff
@@ -1,17 +1,25 @@
...
 
+
```
felixT2K, line 7: don't think the pypi badge and installation guide inside the docs are enough ? :)

notebooks/README.rst
```diff
@@ -14,4 +14,4 @@ Here are some notebooks compiled for users to better leverage the library capabi
 +--------------------------------------------------------------------------------------------------------+----------------------------------------------+---------------------------------------------------------------------------------------------------------------------+
 | `[Artefact detection] <https://github.com/mindee/notebooks/blob/main/doctr/artefact_detection.ipynb>`_ | Object detection for artefacts in documents  | .. image:: https://colab.research.google.com/assets/colab-badge.svg                                                 |
 |                                                                                                        |                                              |        :target: https://colab.research.google.com/github/mindee/notebooks/blob/main/doctr/artefact_detection.ipynb  |
-+--------------------------------------------------------------------------------------------------------+----------------------------------------------+---------------------------------------------------------------------------------------------------------------------+
```
felixT2K, line 17: same :)

felixT2K (changes requested):
Hey @odulcy-mindee 👋🏼,

thanks for the fix.
I have always tried to avoid exactly this case because with each installation you will also download the images 😅 
But at the end i think without uploading the images to a file server we have no chance to fix this unfortunately :/ 
(In short i agree that the only way is to keep the images in a docs folder just a few style changes please 👍🏼 )

README.md
felixT2K: @odulcy-mindee https://github.com/mindee/doctr#installation

odulcy-mindee:
> I have always tried to avoid exactly this case because with each installation you will also download the images 😅

Using a git-clone, yes :/
But for PyPi, it's fine

.github/ISSUE_TEMPLATE/bug_report.yml
odulcy-mindee: Don't why, it's pre-commit which did that

felixT2K:
> > I have always tried to avoid exactly this case because with each installation you will also download the images 😅
> 
> Using a git-clone, yes :/ But for PyPi, it's fine

Yes, sorry, it was probably a bit unclear 😅

README.md
odulcy-mindee: @felixT2K We can add a link to this header instead, wdyt ?

felixT2K: That would be fine but than moving it as first point after the Quicktour headline ?

.github/ISSUE_TEMPLATE/bug_report.yml
felixT2K: I see my pre-commit was out of date (after autoupdate it does the same ^^)

README.md
odulcy-mindee: @felixT2K Well, in the end, I reverted that. I'm eager to update the README.md but let's do that in another PR.

felixdittrich92 (approved):
Thanks ☺️

PR #1335, opened 2023-09-29
[FIX] clip overflowing probs

felixdittrich92:
This PR:

- clip overflowing probs 
- @odulcy-mindee as mentioned


odulcy-mindee (approved):
Thanks !

PR #1348, opened 2023-10-12
TF change antialias to true

felixT2K:
This PR: 

- changes `antialias` also for TF to True

NOTE tests:

- super minimal difference (< 1e-6) in the expected output  (e.g.: 5.00001 instead of 5.0  /  1.00002 instead of 1.0) - 
I don't think there's much we can do about that (the results are correct and close to PT see below)

### DIFF
--- a/doctr/models/preprocessor/tensorflow.py
+++ b/doctr/models/preprocessor/tensorflow.py
@@ -106,7 +106,7 @@ class PreProcessor(NestedObject):
                 x = tf.image.convert_image_dtype(x, dtype=tf.float32)
             # Resizing
             if (x.shape[1], x.shape[2]) != self.resize.output_size:
-                x = tf.image.resize(x, self.resize.output_size, method=self.resize.method)
+                x = tf.image.resize(x, self.resize.output_size, method=self.resize.method, antialias=True)
 
             batches = [x]
 
--- a/doctr/models/preprocessor/tensorflow.py
+++ b/doctr/models/preprocessor/tensorflow.py
@@ -106,7 +106,9 @@ class PreProcessor(NestedObject):
                 x = tf.image.convert_image_dtype(x, dtype=tf.float32)
             # Resizing
             if (x.shape[1], x.shape[2]) != self.resize.output_size:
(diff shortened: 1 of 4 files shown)


felixT2K:
10 x abs

both with antialiasing:
```
opencv_tf_pil_tf_diff:  108826.1
opencv_torch_pil_torch_diff:  1859450
opencv_tf_opencv_torch_diff:  6922.742
pil_torch_pil_tf_diff:  7202.461
```
opencv_tf_pil_tf_diff
opencv_torch_pil_torch_diff
opencv_tf_opencv_torch_diff
pil_torch_pil_tf_diff


only pt with antialiasing:
```
opencv_tf_pil_tf_diff:  428563.7
opencv_torch_pil_torch_diff:  1859450
opencv_tf_opencv_torch_diff:  470293.66
pil_torch_pil_tf_diff:  183107.75
```
opencv_tf_pil_tf_diff
opencv_torch_pil_torch_diff
opencv_tf_opencv_torch_diff
pil_torch_pil_tf_diff

felixT2K:
observation over 30 images:

straight pages: no visible changes
rotated pages: narrow text is detected somewhat less well (should be fixed if we do the next run with rotated augmentation)

@charlesmindee @odulcy-mindee wdyt ?

odulcy-mindee:
@felixdittrich92 I'm not sure to understand the benchmark. Why do you compare tf and torch with PIL and OpenCV ? Is there also antialiasing enabled by default in these frameworks when you downsize an image ?

felixT2K:
> @felixdittrich92 I'm not sure to understand the benchmark. Why do you compare tf and torch with PIL and OpenCV ? Is there also antialiasing enabled by default in these frameworks when you downsize an image ?

Correct after this PR we use in TF and PT antialias so the images are now much more close also for training runs

felixT2K:
Add the end what we want is to work on the same images in both frameworks in the whole pipe inference and training
and with antialias it's the closest we can get atm (in pytorch we use antialising already since 0.7.0)

odulcy-mindee (approved):
Thanks !

PR #1355, opened 2023-10-17
hebrew letters

uriva:


doctr/datasets/vocabs.py
```diff
@@ -13,7 +13,7 @@
     "digits": string.digits,
     "ascii_letters": string.ascii_letters,
     "punctuation": string.punctuation,
-    "currency": "£€¥¢฿",
+    "currency": "£€¥¢฿₪",
```
felixT2K, line 16: @uriva This needs to be added to Line 47 otherwise it would break our existing models :)

felixT2K (changes requested):
Thansk for the PR @uriva only one minor change please :)

uriva:
ptal

felixdittrich92 (approved):
Thanks :)

PR #1360, opened 2023-10-26
[Fix]  tf augmentations

felixT2K:
This PR:

- fix `ToGray` output channels issue
- unify recognition augs from pt and tf (except RandomPerspective)


odulcy-mindee (approved):
Thanks Felix !

PR #1365, opened 2023-10-30
[misc] increase to 0.8.0 and temp pin onnx

felixT2K:
This PR:

- increase version to 0.8.0 (docker feature / orientation improvement / etc. -> minor instead of patch)
- temp pin onnx until https://github.com/onnx/tensorflow-onnx/pull/2252 is fixed


felixdittrich92:
@odulcy-mindee forced to fix CI

felixdittrich92 (approved):

PR #1322, opened 2023-09-18
feat: :sparkles: Official docker images for docTR

odulcy-mindee:
# TODO

- [x] TF Dockerfile
Still heavy TBH:
```
REPOSITORY                                 TAG                                              IMAGE ID       CREATED             SIZE
test_doctr_gpu                             latest                                           936c9e203093   17 minutes ago      8.22GB
```
- [x] PT Dockerfile
- [ ] CI job to publish docTR image

PR #1373, opened 2023-11-05
Relax Pillow and OpenCV version bounds.

nh2:
For Pillow, the change in commit

e04e183b - chore: apply PIL major changes and increase min version specifier (#1237)

that does the `font.getsize()` -> `font.getbbox()` migration is definitely already available in Pillow 9.2.0, see

https://pillow.readthedocs.io/en/stable/deprecations.html#font-size-and-offset-methods

> Deprecated since version 9.2.0.


felixdittrich92:
Hi @nh2 👋,
Thanks for the PR.
What's the reason to increase the lower bound of opencv ? :)

felixdittrich92:
Please update your branch from main :)

felixdittrich92 (approved):
Thanks sounds great @nh2 👍 
If your mentioned PR is merged feel free to open a follow up PR we should add it to the installation docs as option :)

nh2:
> If your mentioned PR is merged feel free to open a follow up PR we should add it to the installation docs as option :)

@felixdittrich92 Ah yes, I just spotted from  #1322 that you're interested in packaging options. Nix can be a convenient alterntaive to docker.

Here's the one-liner invocation that uses the WIP commit from my PR; it drops you into a Python shell where doctr is working, with all dependencies reproducibly pinned (assuming [Nix is installed](https://nixos.org/download)):

```sh
NIX_PATH=nixpkgs=https://github.com/NixOS/nixpkgs/archive/6c069433b46584559bdadcb4d74fb760308aabb0.tar.gz nix-shell -p 'python3.withPackages (ps: with ps; [ doctr ])' --run python3
```

```
>>> import doctr
>>> doctr.__version__
'0.7.0a0'
```

PR #1389, opened 2023-11-22
feat: :sparkles: use `tqdm` instead of `fastprogress` in reference scripts

odulcy-mindee:
I have not been able to run `train_pytorch.py` script on Colab. An idea is to use `tqdm` instead of `fastprogress`.

Related to https://github.com/mindee/doctr/discussions/1388 and https://github.com/mindee/doctr/discussions/1106

### DIFF
--- a/references/classification/train_pytorch.py
+++ b/references/classification/train_pytorch.py
@@ -15,7 +15,6 @@ import time
 import numpy as np
 import torch
 import wandb
-from fastprogress.fastprogress import master_bar, progress_bar
 from torch.nn.functional import cross_entropy
 from torch.optim.lr_scheduler import CosineAnnealingLR, MultiplicativeLR, OneCycleLR
 from torch.utils.data import DataLoader, RandomSampler, SequentialSampler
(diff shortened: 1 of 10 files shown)


felixdittrich92 (changes requested):
Thanks @odulcy-mindee lgtm :)
Only one thing if we are onces on it could you extend the progress bar also to each evaluation loop ?

references/classification/train_pytorch.py
```diff
@@ -329,9 +330,9 @@ def main(args):
     # Create loss queue
     min_loss = np.inf
     # Training loop
-    mb = master_bar(range(args.epochs))
+    mb = trange(args.epochs)
```
felixT2K, line 333: Can be removed ?

references/classification/train_tensorflow.py
```diff
@@ -293,9 +294,9 @@ def main(args):
     min_loss = np.inf
 
     # Training loop
-    mb = master_bar(range(args.epochs))
+    mb = trange(args.epochs)
```
felixT2K, line 297: same and so on :)

references/classification/train_pytorch.py
odulcy-mindee: Yeah, it's not that useful

felixT2K (approved):
Thanks :)

felixdittrich92 (approved):

PR #1390, opened 2023-11-23
[demo] remove limitation and update demo

felixdittrich92:
This PR:

- removes the limitation so that all models can also be used for possible rotated inputs
- updates the demo and add some more options

### DIFF
--- a/doctr/models/detection/linknet/tensorflow.py
+++ b/doctr/models/detection/linknet/tensorflow.py
@@ -27,12 +27,6 @@ default_cfgs: Dict[str, Dict[str, Any]] = {
         "mean": (0.798, 0.785, 0.772),
         "std": (0.264, 0.2749, 0.287),
         "input_shape": (1024, 1024, 3),
-        "url": "https://doctr-static.mindee.com/models?id=v0.6.0/linknet_resnet18-611b50f2.zip&src=0",
-    },
-    "linknet_resnet18_rotation": {
-        "mean": (0.798, 0.785, 0.772),
-        "std": (0.264, 0.2749, 0.287),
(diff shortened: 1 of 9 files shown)


doctr/models/detection/linknet/tensorflow.py
```diff
@@ -20,16 +20,10 @@
...
 
-__all__ = ["LinkNet", "linknet_resnet18", "linknet_resnet34", "linknet_resnet50", "linknet_resnet18_rotation"]
+__all__ = ["LinkNet", "linknet_resnet18", "linknet_resnet34", "linknet_resnet50"]
 
 default_cfgs: Dict[str, Dict[str, Any]] = {
     "linknet_resnet18": {
-        "mean": (0.798, 0.785, 0.772),
-        "std": (0.264, 0.2749, 0.287),
-        "input_shape": (1024, 1024, 3),
-        "url": "https://doctr-static.mindee.com/models?id=v0.6.0/linknet_resnet18-611b50f2.zip&src=0",
```
odulcy-mindee, line 30: This one should be kept and remove the other one ?

felixT2K: Mh it's the same arch so i changed the chekpoint because the sec one was already trained on rotated samples so while finetuning it should converge faster as the checkpoint which wasn't trained with rotated samples :)

Makes only sense for TF because the change to antialias=True was much less changing as compared to pytorch

felixT2K: In the end we'll train it anyway ^^

odulcy-mindee (approved):
Thanks !

PR #1399, opened 2023-12-03
Update README.md

ooguz:
Hello,

Since we are trying docTR for the first time, while following the README tutorial, got an error on the `result.show()` line as:
```
>>> result.show()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: show() missing 1 required positional argument: 'pages'
```

After reading the documentation, I've realized that it must be `result.show(doc)`, so there is a fix for that. So many people may got this error as well.

Sincerely,

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ If both options are set to False, the predictor will always fit and return rotat
 To interpret your model's predictions, you can visualize them interactively as follows:
 
 ```python
-result.show()
+result.show(doc)
 ```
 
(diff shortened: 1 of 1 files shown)


felixdittrich92:
Hi @ooguz Thanks for opening, you are right with version 0.7.0 it is `.show(doc)` but the readme is already updated for the new release 0.8.0 and here we have a change to `.show()` so everything is correct after the next release :)

PR #1412, opened 2023-12-21
[Fixes / docs] Add more vocabs / Fix Style / HF hub / API Dep

felixdittrich92:
This PR:

- adds some more vocabs and updated doc
- fix HF hub upload (increased to >=0.20)
- fix style
- remove onnx constraint
- fix api dep (TF needs py 3.9)
- Add two new custom models to the docs list

Any feedback is welcome :hugs:

PR #1414, opened 2023-12-22
feat: Adds conda recipe & corresponding CI jobs

### LINKED ISSUE #113: [conda] Unable to make a conda build

frgfm:
- adds a conda recipe
- adds a CI job to build & upload to anaconda cloud upon release
- adds another to check that it can then be installed

Note: for now the recipe is without TF or PyTorch. Conditional builds seem tricky for now :sweat_smile: 

cc @mara004 

Closes #113


.conda/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - shapely >=1.6.0, <3.0.0
+    - langdetect >=1.0.9, <2.0.0
+    - rapidfuzz >=3.0.0, <4.0.0
+    - matplotlib >=3.1.0
+    - weasyprint >=55.0
+    - defusedxml >=0.7.0
+    - mplcursors >=0.3
+    - unidecode >=1.0.0
+    - tqdm >=4.30.0
+    - huggingface_hub >=0.5.0
```
felixdittrich92, line 41: >=0.20.0 (See #1412 )

.github/workflows/builds.yml
```diff
@@ -49,3 +49,23 @@ jobs:
...
       - name: Import package
         run: python -c "import doctr; print(doctr.__version__)"
+
+  conda:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: conda-incubator/setup-miniconda@v3
+        with:
+          auto-update-conda: true
```
felixdittrich92, line 59: Python 3.9 because TF minimum is 3.9 ?

felixdittrich92:
Overall fine but i would suggest to update the docs accordingly :)

.github/workflows/builds.yml
```diff
@@ -49,3 +49,23 @@ jobs:
...
+          python-version: 3.9
+      - name: Install dependencies
+        shell: bash -el {0}
+        run: conda install -y conda-build conda-verify anaconda-client
+      - name: Build and publish
+        shell: bash -el {0}
```
felixdittrich92, line 65: Mh i think we can get the version dynamic from the above `doctr__version__` ?

.conda/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+  run:
+    - importlib_metadata
+    - numpy >=1.16.0, <2.0.0
+    - scipy >=1.4.0, <2.0.0
+    - pillow >=9.2.0
+    - matplotlib >=3.1.0
+    - h5py >=3.1.0, <4.0.0
+    - opencv >=4.5.0, <5.0.0
+    - pypdfium2_helpers >=4.0.0, <5.0.0
```
mara004, line 30: `pypdfium2-team::pypdfium2_helpers` ?

.github/workflows/builds.yml
```diff
@@ -49,3 +49,23 @@ jobs:
...
+        run: |
+          echo "BUILD_VERSION=0.8.0a0" | cut -c 2- >> $GITHUB_ENV
+          python setup.py sdist
+          mkdir conda-dist
+          conda build .conda/ -c conda-forge -c pypdfium2-team -c bblanchon --output-folder conda-dist
```
mara004, line 70: ... these channels will also have to be added on the end users' install command, which is tedious - sadly custom channels are second class with conda ...

.conda/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+build:
+  noarch: python
+  script: python setup.py install --single-version-externally-managed --record=record.txt
+
+requirements:
+  host:
+    - python>=3.8, <4.0
+    - setuptools
+
+  run:
```
felixdittrich92, line 22: `{{ project.get('dependencies') }}` does not work because of `pypdfium2_helpers` instead of `pypdfium2` ?

.github/workflows/builds.yml
felixdittrich92: an entry for the installation docs section please then it should be fine :sweat_smile:

felixdittrich92 (changes requested):
Thanks again :+1: 
Last two comments from my side ^^

.conda/meta.yaml
frgfm: assumably correct I didn't even try since there is no link from pypi to the conda channel & package

.conda/meta.yaml
mara004: Yes, but I believe it ties the package to the right channel.
e.g. if a pypdfium2_helpers package appeared in some other higher-priority channel, the above should presumably override that

.conda/meta.yaml
mara004: Hmm, indeed the name incongruence turns out to be unfortunate here.

Is there any chance of replacing `pypdfium2` by `pypdfium2_helpers` after the `project.get(...)` via jinja dynamics?

mara004: Perhaps just like below with the summary ?
`{{ project.get('dependencies') | replace("pypdfium2", "pypdfium2_helpers") }}`
But I don't know if that works with lists...

.github/workflows/publish.yml
```diff
@@ -64,3 +64,45 @@ jobs:
...
+    needs: conda
+    steps:
+      - uses: conda-incubator/setup-miniconda@v3
+        with:
+          auto-update-conda: true
+          python-version: 3.8
+      - name: Install package
+        shell: bash -el {0}
+        run: |
+          conda install -c mindee doctr
```
felixdittrich92, line 108: @odulcy-mindee can you create a (public) conda channel for mindee and afterwards add the token as `ANACONDA_TOKEN` to the github secrets please ? :sweat_smile: 
@frgfm That's the way how we can create a custom channel (correct me if i m wrong) : https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/create-custom-channels.html by replacing `channel` with `mindee`

.github/workflows/builds.yml
mara004: still it's a bit sad end users can't just do `conda install doctr`
but in any case, they'll have to specify the channel you'll be publishing to, so hopefully two more custom channels shouldn't be that much of a problem

.github/workflows/publish.yml
```diff
@@ -64,3 +64,45 @@ jobs:
...
+        id: release_tag
+        run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
+      - name: Build and publish
+        shell: bash -el {0}
+        env:
+          ANACONDA_API_TOKEN: ${{ secrets.ANACONDA_TOKEN }}
+          VERSION: ${{ steps.release_tag.outputs.VERSION }}
+        run: |
+          echo "BUILD_VERSION=${VERSION}" >> $GITHUB_ENV
+          python setup.py sdist
```
mara004, line 90: calling `setup.py` directly is deprecated, you'll want something like `python -m build --sdist` instead
see also https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

felixdittrich92: @frgfm  ? :)

.conda/meta.yaml
felixdittrich92: @frgfm  ? :)

.github/workflows/publish.yml
frgfm: Oh I didn't know, thanks mara!
But for this PR, I don't think we need the edit:
- we'd have to do this for the other occurrences in the repo (feels like a standalone PR to me, and we should check backward compatibility)
- it will be deprecated but it's still working to the best of my understanding right?

felixdittrich92 (approved):
Thanks @frgfm :hugs:

.github/workflows/publish.yml
mara004: > it will be deprecated but it's still working to the best of my understanding right?

Yes, for the time being. They might want to remove it after some deprecation period, but I don't know if they'll ever do, as this somehow doesn't seem to have reached the crowd at large.

.github/workflows/publish.yml
felixdittrich92: @odulcy-mindee we need only the token to add :) No first manual upload or something else that's done about the CI with the next release :)

felixdittrich92:
@odulcy-mindee How does it look do we have a token now ? :sweat_smile:

odulcy-mindee (approved):
Token has been added !

mara004:
I suppose you're aware the installation instructions are yet TBD

RoyiAvital:
Any chance to update the recipe (https://anaconda.org/channels/techmindee/packages/python-doctr/overview) and add Windows support?

PR #1419, opened 2024-01-03
[FIX] Fix mypy errors

felixdittrich92:
This PR:

- fix / ignore opencv related mypy errors
- seems to be a known issue with the latest OpenCV version: https://github.com/opencv/opencv/issues/24803

@odulcy-mindee I checked the function calls and everything is fine (only switch from int to float in `fillPoly`)
Any feedback is welcome :) 

PS: I think we can remove the ignores after the next opencv / mypy patch release

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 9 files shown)


odulcy-mindee (approved):
Thanks Felix for this fix !

PR #1424, opened 2024-01-08
feat: :sparkles: new TF Linknet Resnet checkpoints

odulcy-mindee:
Update existing TF Linknet Resnet checkpoints. They have been trained using new augmentations from reference scripts.


felixdittrich92 (approved):
Thanks ☺️

PR #1432, opened 2024-01-17
[references] TF / PT crop & document orientation classifier train scripts

felixdittrich92:
This PR:

- Add training scripts (TF & PT) for the crop orientation classifier and additional for general document orientation classification (needed for 0.9.0 (similar to the crop_orientation_predictor) to generally recognize the documents rotation angle correctly before we compute the angle from the bin_map - which works only for rotated documents in the range ~ -60-60 well)
- Add a simple dataset (prefilled with 0 degree targets - to match the AbstractDataset checks)

Any feedback is welcome :)

- [x] Tests
- [x] Optimization  (input_size tests left)  - for crop orientation (32x32 is enough and works well) / for document orientation (256x256 seems to be enough)
- [x] Clean up
- [x] Readme

Dummy runs (mobilenet_v3_small)

```
crop orientation (800k train / 200k val)
Validation loss decreased 0.372947 --> 0.355291: saving state...
Epoch 5/30 - Validation loss: 0.355291 (Acc: 81.51%)

doc orientation (only 4k train / 1k val)
Validation loss decreased 0.559973 --> 0.481535: saving state...
Epoch 14/30 - Validation loss: 0.481535 (Acc: 79.22%)
```

### DIFF
--- a/.github/workflows/references.yml
+++ b/.github/workflows/references.yml
@@ -109,10 +109,10 @@ jobs:
           unzip toy_detection_set-bbbb4243.zip -d det_set
       - if: matrix.framework == 'tensorflow'
         name: Train for a short epoch (TF)
-        run: python references/classification/train_pytorch_orientation.py ./det_set ./det_set mobilenet_v3_small -b 2 --epochs 1
+        run: python references/classification/train_tensorflow_orientation.py ./det_set ./det_set mobilenet_v3_small -b 2 --epochs 1
       - if: matrix.framework == 'pytorch'
         name: Train for a short epoch (PT)
-        run: python references/classification/train_tensorflow_orientation.py ./det_set ./det_set mobilenet_v3_small -b 2 --epochs 1
+        run: python references/classification/train_pytorch_orientation.py ./det_set ./det_set mobilenet_v3_small -b 2 --epochs 1
 
   train-text-recognition:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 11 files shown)


odulcy-mindee (commented):
Code seems good :+1:

felixdittrich92:
@odulcy-mindee So this part could be merged into 0.8.0 and the rest of this task is for 0.9.0 :)
Are you fine with the additional `OrientationDataset`  if was thinking on putting it into references/classification/utils but i think the dataset folder is a better fit :sweat_smile:

odulcy-mindee (commented):
> Are you fine with the additional OrientationDataset if was thinking on putting it into references/classification/utils but i think the dataset folder is a better fit 😅
@felixdittrich92 yeah, that makes more sense :+1:

I added one more suggestion.
For another PR: is it possible to have only one `train_{framework}.py` for both purposes ? Otherwise, maybe it makes more sense to rename the other scripts `train_{framework}_character.py` or something else ?

odulcy-mindee (approved):
Thank you @felixdittrich92 ! :rocket:

felixdittrich92:
> > Are you fine with the additional OrientationDataset if was thinking on putting it into references/classification/utils but i think the dataset folder is a better fit 😅
> > @felixdittrich92 yeah, that makes more sense 👍
> 
> I added one more suggestion. For another PR: is it possible to have only one `train_{framework}.py` for both purposes ? Otherwise, maybe it makes more sense to rename the other scripts `train_{framework}_character.py` or something else ?

I totally missed this :sweat_smile: 
Merging both into one makes no sense in my mind i think this would be confusing 
But renaming the other one with suffix character would be fine :+1:

PR #1444, opened 2024-01-30
[references] Add poly scheduler for detection training

felixdittrich92:
This PR:

- Add Polynomial Scheduler for detection training (TF & PT)
- It's often mentioned in text detection papers like DB & FAST & .. so i think it is a good idea to add it :)

### DIFF
--- a/references/detection/train_tensorflow.py
+++ b/references/detection/train_tensorflow.py
@@ -421,7 +421,7 @@ def parse_args():
         help="metrics evaluation with straight boxes instead of polygons to save time + memory",
     )
     parser.add_argument(
-        "--sched", type=str, default="exponential", choices=["exponential", "poly"], help="scheduler to use"
+        "--sched", type=str, default="poly", choices=["exponential", "poly"], help="scheduler to use"
     )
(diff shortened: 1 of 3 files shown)


references/detection/train_tensorflow.py
```diff
@@ -268,14 +268,25 @@ def main(args):
...
+        )
+    elif args.sched == "poly":
+        scheduler = tf.keras.optimizers.schedules.PolynomialDecay(
+            args.lr,
+            decay_steps=args.epochs * len(train_loader),
+            end_learning_rate=1e-7,
+            power=1.0,
+            cycle=False,
+            name="PolynomialDecay",
+        )
```
odulcy-mindee, line 288: It should raise an error if `args.sched` is not matched with previous values

felixdittrich92:
@odulcy-mindee I think with choices it's more clean instead of raising an exception :)
Switched the default to poly -> (After testing LinkNet / DB & FAST on 2 different detection datasets x Poly / Cosine / Onecycle -> Poly has reached the best results with mostly no drops)

odulcy-mindee (approved):
Yeah, it's better with choices, thanks !

PR #1447, opened 2024-02-01
[FIX] Fix streamlit demo

felixdittrich92:
This PR:

- fix for demo with latest streamlit version

### DIFF
--- a/demo/app.py
+++ b/demo/app.py
@@ -114,7 +114,7 @@ def main(det_archs, reco_archs):
 
                 # Display JSON
                 st.markdown("\nHere are your analysis results in JSON format:")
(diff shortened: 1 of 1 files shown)


odulcy-mindee (dismissed):
Thank you !

FYI, I set `st.json(..., expanded=False)` on HF Space, otherwise when you enlarge one of the images above with JSON expanded, you got a blank page and you need to scroll to see your image. This way, it does not look like a bug "from our side" but a "bug" on Steamlit

felixdittrich92:
@odulcy-mindee sounds good i have applied it also to the local demo

odulcy-mindee (approved):

PR #1449, opened 2024-02-02
[prototype] loc pred callbacks

felixdittrich92:


demo/app.py
odulcy-mindee, line 1: We need to do the same on HF Space

docs/source/using_doctr/using_models.rst
```diff
@@ -398,3 +398,48 @@ For reference, here is a sample XML byte string output:
...
+            # The outpout structure needs to be the same as the input location predictions
+            return loc_preds
+
+    my_hook = CustomHook()
+
+    predictor = ocr_predictor(pretrained=True)
+    # Add a hook in the middle of the pipeline
+    predictor.add_hook(my_hook)
+    # You can also add multiple hooks which will be executed sequentially
+    [predictor.add_hook(hook) for hook in [my_hook, my_hook, my_hook]]
```
odulcy-mindee, line 445: Can we write this part using:
```python
for hook in [my_hook, my_hook, my_hook]:
    predictor.add_hook(hook)
```
please ? :see_no_evil:

docs/source/using_doctr/using_models.rst
```diff
@@ -398,3 +398,48 @@ For reference, here is a sample XML byte string output:
...
+
+
+* Add a hook to the `ocr_predictor` to manipulate the location predictions before the crops are passed to the recognition model.
+
+.. code:: python3
+
+    from doctr.model import ocr_predictor
+
+    class CustomHook:
+        def __call__(self, loc_preds):
```
odulcy-mindee, line 434: Has `loc_preds` a specific format ? I saw it can be a dict but can it be something else ?

doctr/models/preprocessor/pytorch.py
```diff
@@ -35,7 +35,6 @@ def __init__(
         batch_size: int,
         mean: Tuple[float, float, float] = (0.5, 0.5, 0.5),
         std: Tuple[float, float, float] = (1.0, 1.0, 1.0),
-        fp16: bool = False,
```
odulcy-mindee, line 38: No more used or was it here only for signature consistency for something else ?

odulcy-mindee (changes requested):

doctr/models/preprocessor/pytorch.py
felixT2K: it's unused from an old try now it's simply `.cuda().half()`  to the predictor or the TF equivalent :)

docs/source/using_doctr/using_models.rst
felixT2K: yeah we can do this 😅 ^^

demo/app.py
felixT2K: yep this would be good afterwards :)

docs/source/using_doctr/using_models.rst
felixT2K: For KIE it's a dict with the labels and np.ndarrys with the coords otherwise it's only an np.ndarray (coords are quads or rects - depends on `assume_straight_pages` True or False)

doctr/models/preprocessor/pytorch.py
odulcy-mindee: ok, thanks

docs/source/using_doctr/using_models.rst
felixdittrich92: ok to check on every hook that the values are still between 0 and 1 produces some overhead ... as a advanced option i think that's something the user needs to be aware

felixdittrich92:
@odulcy-mindee I added two points in the docs - as an advanced functionality for manipulation - i think it's better to mention it in the docs instead of producing some overhead to check after each hook that the coords are still relative in the range between 0 and 1 :sweat_smile:

felixdittrich92:
> @odulcy-mindee I added two points in the docs - as an advanced functionality for manipulation - i think it's better to mention it in the docs instead of producing some overhead to check after each hook that the coords are still relative in the range between 0 and 1 😅

@dchaplinsky What do you think from a user perspective ? Is it clear enough from the docs that input/output structure needs to be the same and the coords are relative so it needs also to be returned as relative (0-1) or should we add a check for the value range (which would produce a small overhead)

odulcy-mindee:
@felixdittrich92 I'm fine with it !

odulcy-mindee (approved):
@felixdittrich92 Thnak you Felix, nice to have these hooks !

dchaplinsky:
@felixdittrich92 well documented interface is enough. And the output format is dictated by the input format, so it will be strange if someone will receive relative coordinates and return absolute instead.

felixdittrich92:
> @felixdittrich92 well documented interface is enough. And the output format is dictated by the input format, so it will be strange if someone will receive relative coordinates and return absolute instead.

Oh believe me i saw a lot 😂

PR #1453, opened 2024-02-07
test

felixdittrich92:


.github/workflows/publish.yml
```diff
@@ -105,4 +107,4 @@ jobs:
       - name: Install package
         shell: bash -el {0}
         run: |
-          conda install -c mindee doctr
+          conda install -c mindee -c conda-forge -c pypdfium2-team -c bblanchon doctr
```
felixdittrich92, line 110: @frgfm @mara004 the channels was missing before correct ? Or does it use it from the config and we can keep only the mindee channel ?

felixdittrich92:
Thanks @mara004 your conda section from pypdfium2 was really helpful :+1:

.github/workflows/publish.yml
felixdittrich92: changed to:

```
conda config --set channel_priority strict
conda install -c mindee -c pypdfium2-team -c bblanchon -c conda-forge python-doctr
```

odulcy-mindee (approved):

.github/workflows/publish.yml
```diff
@@ -105,4 +107,5 @@ jobs:
       - name: Install package
         shell: bash -el {0}
         run: |
-          conda install -c mindee doctr
+          conda config --set channel_priority strict
+          conda install -c mindee -c pypdfium2-team -c bblanchon -c conda-forge python-doctr
```
felixT2K, line 111: @frgfm the channel is created automatically the first time it's uploaded correct ? 😅

.github/workflows/builds.yml
```diff
@@ -53,18 +53,20 @@ jobs:
   conda:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - uses: conda-incubator/setup-miniconda@v3
         with:
           auto-update-conda: true
           python-version: 3.8
+          channels: pypdfium2-team,bblanchon,conda-forge
```
frgfm, line 61: The nice thing with putting the channels directly in the command is that it will be the same for installation & we don't rely too much on the evolution of this Github action.

.github/workflows/publish.yml
frgfm: Nope, the channel has to be created along with the token so that we can upload it :)

frgfm (approved):
Thanks for the PR felix! A few gardening comments (the conda channel & token needs to be set for the job to work)

.github/workflows/builds.yml
felixT2K: Mh yeah that's correct but now we have both ways with the action and in publish the same thing without (for installation) :)

PR #1454, opened 2024-02-07
[references] Add interval saving for detection trainings

felixdittrich92:
This PR:

- Add the option to save after each epoch (@odulcy-mindee as we saw this would be a very useful option ^^)


odulcy-mindee (approved):
Thanks !

PR #1457, opened 2024-02-08
[Misc] drop py 3.8 support

felixdittrich92:
On hold - after release 0.8.0

This PR:

- drop py3.8 support (with the next release this should be fine -> https://devguide.python.org/versions/)
- tf still keeped with lower boundary at >=2.11 
- fixed to <2.16.0 because keras-core (keras3) would destroy the whole TF part - needs to be updated in another PR - breaking change
- modified the references jobs a bit for speed up

TF:
```
Support for Python 3.8 has been removed starting with TF 2.14. The TensorFlow 2.13.1 patch release will still have Python 3.8 support.
```
https://github.com/tensorflow/tensorflow/releases/tag/v2.14.0

### DIFF
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -10,7 +10,7 @@ authors = ["Mindee <contact@mindee.com>"]
 license = "Apache-2.0"
 
 [tool.poetry.dependencies]
-python = ">=3.9,<4.0"  # pypdfium2 needs a python version above 3.8.2
+python = ">=3.9,<3.12"
 tensorflow = ">=2.14.0,<2.16.0"  # cf. https://github.com/mindee/doctr/pull/1461
 python-doctr = {git = "https://github.com/mindee/doctr.git", extras = ['tf'], branch = "main" }
 # Fastapi: minimum version required to avoid pydantic error
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -11,7 +11,7 @@ license = "Apache-2.0"
 
 [tool.poetry.dependencies]
 python = ">=3.9,<3.12"
-tensorflow = ">=2.14.0,<2.16.0"  # cf. https://github.com/mindee/doctr/pull/1461
+tensorflow = ">=2.11.0,<2.16.0"  # cf. https://github.com/mindee/doctr/pull/1461
 python-doctr = {git = "https://github.com/mindee/doctr.git", extras = ['tf'], branch = "main" }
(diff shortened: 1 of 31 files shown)


felixdittrich92:
@odulcy-mindee Let's talk about this next week again i don't think that it raises any impact on your side if you use a pinned doctr version.
For development it's crucical because in our CI with py3.8 we are bound to TF <= 2.13 / which makes it also not possible to work on the breaking keras changes in >= TF 2.16 :)

felixdittrich92:
@odulcy-mindee 
If your internal usage of doctr is pinned (what i thing) we could merge this also and update the github ci jobs :)
Before this isn't merged it makes less sense to work on support for TF >2.14 (>=2.16 with keras-v3) ^^

pyproject.toml
```diff
@@ -13,7 +13,7 @@ maintainers = [
     {name = "Felix Dittrich"},
 ]
 readme = "README.md"
-requires-python = ">=3.8.0,<4"
+requires-python = ">=3.9.0,<4"
```
felixdittrich92, line 16: @odulcy-mindee as alternative we could keep 3.8 here for the moment until you have internal also the 3.9 upgrade done wdyt ?
But all the tests should switch already to 3.9 as new "main" version

pyproject.toml
```diff
@@ -13,7 +13,7 @@ maintainers = [
     {name = "Felix Dittrich"},
 ]
 readme = "README.md"
-requires-python = ">=3.8.0,<4"
+requires-python = ">=3.8.0,<4"  # TODO: bump to >=3.9 before release
```
felixdittrich92, line 16: @odulcy-mindee as discussed

api/pyproject.toml
```diff
@@ -10,8 +10,8 @@ authors = ["Mindee <contact@mindee.com>"]
 license = "Apache-2.0"
 
 [tool.poetry.dependencies]
-python = ">=3.8.2,<3.11"  # pypdfium2 needs a python version above 3.8.2
-tensorflow = ">=2.11.0,<2.16.0"  # cf. https://github.com/mindee/doctr/pull/1461
+python = ">=3.9,<3.12"
+tensorflow = ">=2.14.0,<2.16.0"  # cf. https://github.com/mindee/doctr/pull/1461
```
odulcy-mindee, line 14: Do we need to bump min version of TF ?

odulcy-mindee (approved):

PR #1479, opened 2024-02-23
read labels in utf-8 and log input string on vocab error

eikaramba:
fixes issues raised in https://github.com/mindee/doctr/discussions/1366 where vocab issues arrise if labels.json file is not processed in correct encoding


felixdittrich92:
@eikaramba Thanks for the PR 👍 
Could you run `make style` please to fix formating afterwards we are fine

felixdittrich92 (approved):
Thanks for the quick improvements 🤗

PR #1499, opened 2024-03-04
feat: :sparkles: torch fast_tiny checkpoint

odulcy-mindee:


kuraga:
Hi @felixdittrich92 ,

> You can fine tune the model initialized from this checkpoint if that's what you mean :)

sure can :)

More precise: is the training code closed source?

felixdittrich92:
It's also open see: https://mindee.github.io/doctr/using_doctr/custom_models_training.html

Only the mindee internal data which was used to train the models can't be shared because they contain sensible data :)

kuraga:
Ah, thanks!

P.S. a universal training framework,  wow

ajkdrag:
I tried running the FAST detection model. As per Papers-with-code stats, FAST is supposed to run in realtime, but it doesn't seem to be the case. It's almost as slow as DB_RESNET

felixdittrich92:
Hi @ajkdrag :wave:,

That's correct because you run atm the normal non-reparameterized model

function to reparameterize the model for inference is already added: https://github.com/mindee/doctr/blob/058b5db5393702fe93e02cda730346e6928c1558/doctr/models/detection/fast/pytorch.py#L282

but not enabled by default yet first we need to train the models and do some evaluations :)

ajkdrag:
Thanks for the quick response. 
So, at the moment, we can't use reparameterize even with the fast_tiny checkpoint, right?

felixdittrich92:
> Thanks for the quick response. So, at the moment, we can't use reparameterize even with the fast_tiny checkpoint, right?

```
base_model = fast_tiny(pretrained=True).eval()
rep_model = reparameterize(base_model)

model = ocr_predictor(det_arch=rep_model, pretrained=True)
```

But we haven't tested it yet the reparameterization logic is mostly borrowed from the original implementation :)

Some note about the checkpoint we have faiced that the provided checkpoint doesn't performs as expected so we start retraining it today with a higher pooling size and extended evaluation, so in short the checkpoint will change :)

PR #1500, opened 2024-03-05
[models] Tune FAST pooling size

felixdittrich92:
This PR:

- change back to paper mentioned default pooling size

### DIFF
--- a/doctr/models/detection/fast/pytorch.py
+++ b/doctr/models/detection/fast/pytorch.py
@@ -58,9 +58,9 @@ class FastNeck(nn.Module):
         out_channels: int = 128,
     ) -> None:
         super().__init__()
-        self.reduction = nn.ModuleList(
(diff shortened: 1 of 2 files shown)


odulcy-mindee (approved):

ajkdrag:
Any estimate on when can we have the retrained checkpoint? The earlier checkpoint performs fine on few internal documents intrested. But I will still wait for the new one.

felixdittrich92:
Hi @ajkdrag We experiment currently to find the best params because the first checkpoint works super well on detection (recall) but the seg masks was a bit to shrunken (which leads also to a low precision)

First try was to increase the pooling size: failed
Currently we moved back to a pooling size of 4 but increased the shrink ratio by 0.3 (which will result in larger text kernel target masks) so we will see next week how this performs :)

PR #1503, opened 2024-03-06
[Fix] Pin py3.11 for MacOS latest / update publish version checks

felixdittrich92:
MacOS: 
https://github.com/actions/setup-python/issues/855 
https://github.com/actions/setup-python/issues/865


felixdittrich92:
print(doctr.__version__) instead

felixdittrich92:
@odulcy-mindee We could apply this as temp fix because i think the fix in setup-python takes longer as expected :)

odulcy-mindee (commented):
What about pin macos version to `macos-13` instead, as suggested here ?
https://github.com/actions/setup-python/issues/855

felixdittrich92:
> What about pin macos version to `macos-13` instead, as suggested here ? [actions/setup-python#855](https://github.com/actions/setup-python/issues/855)

@odulcy-mindee In this case we would need to update the required CI jobs 

The repo says for python:

```
5 most popular major.minor versions
```

So i expect a fix for this in a future release that we can easily change this back, without updating our required CI jobs.

That's also the reason why i would prefer the py3.11 pin over the maco-13 pin :)

odulcy-mindee:
> @felixdittrich92 In this case we would need to update the required CI jobs

Aaaaaah, that's true :-(

Ok, let's keep it like that

odulcy-mindee (approved):

PR #1514, opened 2024-03-15
[docs] documentation for changing predictors batch sizes

felixdittrich92:
This RP:

- Add missing documentation for batch sizes
- change kwargs get to defaults and update for crop orientation also to 128 (like reco predictor default)


doctr/models/recognition/zoo.py
```diff
@@ -45,7 +45,7 @@ def _predictor(arch: Any, pretrained: bool, **kwargs: Any) -> RecognitionPredict
 
     kwargs["mean"] = kwargs.get("mean", _model.cfg["mean"])
     kwargs["std"] = kwargs.get("std", _model.cfg["std"])
-    kwargs["batch_size"] = kwargs.get("batch_size", 32)
+    kwargs["batch_size"] = kwargs.get("batch_size", 128)
```
odulcy-mindee, line 48: 128? Isn't that big ?

felixdittrich92: Not really (keep in mind that's only 32x128 crops) :) 
I'm not sure why we wrote here as default 32 because the general default is already 128 (so only a correction) see:
https://github.com/mindee/doctr/blob/c2b197deed27147d62e67a47208797c22a908c7e/doctr/models/zoo.py#L25

odulcy-mindee (approved):
:pencil2:

PR #1534, opened 2024-03-28
[prototype] object det replacement / init contrib modules

felixdittrich92:
NOTE: Only a dirty first draft for discussions :)

The idea behind a contrib module:
- easy to extend with further functionalites like: [doctr++](https://github.com/fh2019ustc/DocTr-Plus) | [UVDoc](https://github.com/felixdittrich92/UVDoc) / table Transformer / super resolution / etc. for example
- 3 main requirements: 
-- modules needs to work with the `DocumentBuilder` output structure (List[np.ndarray])
-- framework independent (no torch or tf allowed)
-- if model loading is required -->  onnxruntime

Further advantages on our side:
- Less code to maintain 
- Full focus on OCR quality 
- Maybe also a cool new place for contributors without the need to step deep into doctr's core

A first service for object detection which replaces the current (not really maintained and robust) one :sweat_smile: 

- trained with yolov8 (artefact detection dataset) --> training outsourced to ultralytics -> easy to replace with any yolov8 (no OBB atm) trained model (exported to onnx)

example:
```
import os

from doctr.contrib import ArtefactDetector

root = "/home/felix/Desktop/doctr_test_data"

doc = DocumentFile.from_images([os.path.join(root, "6.jpg"), os.path.join(root, "7.jpg"), "/home/felix/Desktop/5a89cd6d989803e2.jpg"])

detector = ArtefactDetector(batch_size=2, conf_threshold=0.9, iou_threshold=0.5)

res = detector(doc)
detector.show()
print(res)
```

out:
```
[[], [], [{'label': 'photo', 'confidence': 0.9760028, 'box': [665, 194, 767, 319]}, {'label': 'logo', 'confidence': 0.97400737, 'box': [158, 745, 275, 862]}, {'label': 'photo', 'confidence': 0.9690048, 'box': [13, 790, 131, 933]}, {'label': 'bar_code', 'confidence': 0.94711626, 'box': [647, 425, 795, 469]}, {'label': 'logo', 'confidence': 0.94670904, 'box': [314, 9, 372, 67]}, {'label': 'qr_code', 'confidence': 0.94380593, 'box': [368, 691, 417, 740]}, {'label': 'bar_code', 'confidence': 0.9333756, 'box': [291, 813, 339, 850]}, {'label': 'bar_code', 'confidence': 0.93306136, 'box': [391, 108, 683, 151]}]]
```

visualized:
Screenshot from 2024-03-28 11-39-11

TODOS would be:
- refactor code (each contrib module should inerhit from base.py - BasePredictor (call, (_process _preprocess needs every contrib module to implement))
- docs
- tests


doctr/contrib/artefacts.py
```diff
@@ -0,0 +1,119 @@
...
+        model_path: the path to the model to use
+        labels: the labels to use
+        mask_labels: the mask labels to use
+        conf_threshold: the confidence threshold to use
+        iou_threshold: the intersection over union threshold to use
+        **kwargs: additional arguments to be passed to `download_from_url`
+    """
+
+    def __init__(
+        self,
```
felixdittrich92, line 43: @odulcy-mindee If the general idea is fine should we make the config more dynamic with self.cfg ? (default=default_cfgs) ? Especially for custom trained models which was trained on a different input_shape ?

felixdittrich92:
About the `Artefact`-DocumentBuilder element i'm not sure should we keep it for maybe later usage ? (As a kind of "add to the OCR results") Or should we remove it (currently always empty and unused) ?
In general it makes more sense to encapsulate the object detection (and other possible pre-proc modules as mentioned) from the OCR pipeline to interact directly with the `DocumentFile` output.
As an example: Now the object detection pre step could be used for example to mask some detected parts of the images before it's passed to the OCR pipeline to ignore irrelevant fields like images which contains text.

.github/workflows/references.yml
odulcy-mindee, line 1: why is it removed ?

odulcy-mindee:
> In general it makes more sense to encapsulate the object detection (and other possible pre-proc modules as mentioned) from the OCR pipeline to interact directly with the DocumentFile output.

So, purpose of `contrib` is to always wrap a `model` and doing some predictions ? Or can it be only some `postprocessing` function ?

doctr/contrib/base.py
```diff
@@ -0,0 +1,105 @@
...
+        model_path: the path to the model to use
+        **kwargs: additional arguments to be passed to `download_from_url`
+    """
+
+    def __init__(self, batch_size: int, url: Optional[str] = None, model_path: Optional[str] = None, **kwargs) -> None:
+        self.batch_size = batch_size
+
+        self.onnx_model_path = self._init_model(url, model_path, **kwargs)
+        self.session = ort.InferenceSession(
+            self.onnx_model_path, providers=["CUDAExecutionProvider", "CPUExecutionProvider"]
```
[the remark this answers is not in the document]
felixdittrich92: Then it will default to cpu :)

docs/source/using_doctr/using_contrib_modules.rst
```diff
@@ -0,0 +1,49 @@
...
+    detector = ArtefactDetection(batch_size=2, conf_threshold=0.5, iou_threshold=0.5)
+    artefacts = detector(doc)
+
+    # Visualize the detected artefacts
+    detector.show()
+
+You can also use your custom trained YOLOv8 model to detect artefacts or anything else you need.
+Reference: `YOLOv8 <https://github.com/ultralytics/ultralytics>`_
+
+**NOTE:** The YOLOv8 model (no OBB supported yet) needs to be provided as onnx exported model with a dynamic batch size.
```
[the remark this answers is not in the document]
felixdittrich92: correct but OBB in this case means (Oriented Bounding Box) which is supported by Yolov8 but we support atm only the default converted models which returns predictions of straight pages without an rotation angle :)

.github/workflows/references.yml
felixdittrich92: this contrib module replaces the current obj_detection so train scripts, latency script are removed also which includes the CI test :)

felixdittrich92:
> > In general it makes more sense to encapsulate the object detection (and other possible pre-proc modules as mentioned) from the OCR pipeline to interact directly with the DocumentFile output.
> 
> So, purpose of `contrib` is to always wrap a `model` and doing some predictions ? Or can it be only some `postprocessing` function ?

The idea was to make it compatible with the `DocumentFile` output so in this case it's only `preprocessing`.
And yes with the actual state each contrib module would need a onnx model do you have something in mind where it would make sense to have something without a model ? (It's a prototype so open for every idea :D)

doctr/contrib/base.py
```diff
@@ -0,0 +1,105 @@
...
+
+        Args:
+        ----
+            inputs: the inputs to use
+
+        Returns:
+        -------
+            Any: the postprocessed output
+        """
+        self._inputs = inputs
```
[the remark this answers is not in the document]
felixdittrich92: correct or `.show()` like in this example :)

docs/source/using_doctr/using_contrib_modules.rst
```diff
@@ -0,0 +1,49 @@
...
+Available contribution modules
+------------------------------
+
+**NOTE:** To use the contrib module, you need to install the `onnxruntime` package. You can install it using the following command:
+
+.. code:: bash
+
+    pip install onnxruntime
+
+Here are all contribution modules that are available through docTR:
```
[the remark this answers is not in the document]
felixdittrich92: As mentioned currently it's "designed" that each contrib module requires an onnx model :)

doctr/contrib/__init__.py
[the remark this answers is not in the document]
felixdittrich92: yep it's changed :+1:

odulcy-mindee:
> The idea was to make it compatible with the DocumentFile output so in this case it's only preprocessing.
And yes with the actual state each contrib module would need a onnx model do you have something in mind where it would make sense to have something without a model ? (It's a prototype so open for every idea :D)

Ok, for a first version, we can always for a model. Code can be improved later to set onxxmodel as optional.

pyproject.toml
```diff
@@ -155,6 +158,7 @@ module = [
 	"anyascii.*",
 	"tensorflow.*",
 	"torchvision.*",
+    "onnxruntime.*",
```
odulcy-mindee, line 161: not aligned

felixdittrich92:
> > The idea was to make it compatible with the DocumentFile output so in this case it's only preprocessing.
> > And yes with the actual state each contrib module would need a onnx model do you have something in mind where it would make sense to have something without a model ? (It's a prototype so open for every idea :D)
> 
> Ok, for a first version, model is required. Code can be improved later to set onxxmodel as optional.

Correct :)

pyproject.toml
felixdittrich92: that's a github view bug see:
Screenshot from 2024-04-24 05-12-27

doctr/contrib/artefacts.py
```diff
@@ -0,0 +1,131 @@
...
+
+from .base import _BasePredictor
+
+__all__ = ["ArtefactDetector"]
+
+default_cfgs: Dict[str, Dict[str, Any]] = {
+    "yolov8_artefact": {
+        "input_shape": (3, 1024, 1024),
+        "labels": ["bar_code", "qr_code", "logo", "photo"],
+        "url": "https://github.com/mindee/doctr/releases/download/v0.8.1/yolo_artefact-f9d66f14.onnx",
```
odulcy-mindee, line 21:
```suggestion
        "url": "https://doctr-static.mindee.com/models?id=v0.8.1/yolo_artefact-f9d66f14.onnx&src=0",
```

odulcy-mindee (commented):
model uploaded

Issue #1052, opened 2022-09-08
Have different install options to skip installing UI packages (matplotlib & mplcursors) for a production environment  


frallain:
### 🚀 The feature

On a production machine that runs ML algorithms, the UI oriented packages are not necessary to make the algorithms run.

It would be nice to be able to skip these UI oriented packages installation by grouping these UI packages into a `ui` group in the ppyproject.ml file so that we can then perform a `poetry install --without ui`.

### Motivation, pitch

The motivation here is to avoid UI oriented packages dependencies problem on production machines.

frgfm:
Hi @frallain :wave:

I agree that this is a topic, and we've discussed this several times. Here is my point of view:
- yes we should separate non-critical deps. For a maintenance perspective (to prioritize core features/issues against optional ones), they should be moved to optional extras
- however, for production, that would be extremely useful to have your feedback: what are your specific use cases in production? (I'm asking because we try to move everything to ONNX for production to avoid heavy deps and unoptimized DL operations)

frallain:
My specific usecase in production is the following:
```python
from doctr.models.detection import db_resnet50
from doctr.models.detection.predictor import DetectionPredictor
from doctr.models.predictor import OCRPredictor
from doctr.models.preprocessor import PreProcessor
from doctr.models.recognition import crnn_vgg16_bn
from doctr.models.recognition.predictor import RecognitionPredictor

...

def get_ocr_model():
    if MachineLearningModelType.OCR_MODEL not in MACHINE_LEARNING_MODELS:

        detection_model = db_resnet50(
            pretrained_backbone=False  # needed to not download default weights
        )

        detection_predictor = DetectionPredictor(
            PreProcessor(
                detection_model.cfg["input_shape"][:-1],
                mean=detection_model.cfg["mean"],
                std=detection_model.cfg["std"],
                batch_size=1,
            ),
            detection_model,
        )
        download_model_files("db_resnet50")

        detection_predictor.model.load_weights(
            os.path.join(settings.MODELS_DIR, "ocr_model/db_resnet50/weights")
        )

        recognition_model = crnn_vgg16_bn(
            pretrained_backbone=False  # needed to not download default weights
        )

        recognition_predictor = RecognitionPredictor(
            PreProcessor(
                recognition_model.cfg["input_shape"][:2],
                preserve_aspect_ratio=True,
                mean=recognition_model.cfg["mean"],
                std=recognition_model.cfg["std"],
                batch_size=32,
            ),
[...]

Issue #1545, opened 2024-04-10
Cannot run library, issues with gobject-2.0-0


dvalletj:
### Bug description

I'm on Mac OS 13.5.1, running Python 3.9 and I cannot run the library in my python script, been stuck on this for a few days, have tried installing gobject-2.0-0 with no success:

`OSError: cannot load library 'gobject-2.0-0': dlopen(gobject-2.0-0, 0x0002): tried: 'gobject-2.0-0' (no such file), '/System/Volumes/Preboot/Cryptexes/OSgobject-2.0-0' (no such file), '/Users/dv/opt/anaconda3/lib/gobject-2.0-0' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/dv/opt/anaconda3/lib/gobject-2.0-0' (no such file), '/Users/dv/opt/anaconda3/lib/gobject-2.0-0' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/dv/opt/anaconda3/lib/gobject-2.0-0' (no such file), '/Users/dv/opt/anaconda3/lib/python3.9/site-packages/../../gobject-2.0-0' (no such file), '/Users/dv/opt/anaconda3/lib/gobject-2.0-0' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/dv/opt/anaconda3/lib/gobject-2.0-0' (no such file), '/Users/dv/opt/anaconda3/bin/../lib/gobject-2.0-0' (no such file), '/usr/lib/gobject-2.0-0' (no such file, not in dyld cache), 'gobject-2.0-0' (no such file), '/usr/local/lib/gobject-2.0-0' (no such file), '/usr/lib/gobject-2.0-0' (no such file, not in dyld cache). Additionally, ctypes.util.find_library() did not manage to locate a library called 'gobject-2.0-0'
`

```
[...]

Issue #1540, opened 2024-04-06
Update Readme to include common errors installing WeasyPrint


nikitakor7:
### 🚀 The feature

Please link to this https://github.com/Kozea/WeasyPrint/issues/1556#issuecomment-1097977671

### Motivation, pitch

This error is guaranteed

PR #1522, opened 2024-03-25
[API] update api for multi file and pdf support

felixdittrich92:
This PR:

- update API template to support multiple images and/or pdfs
- add config options
- extend tests

Any feedback is welcome :hugs:

PR #1551, opened 2024-04-14
Exclude deps & split into optional parts

felixdittrich92:
This PR:

- exclude weasyprint / matplotlib & mplcursors from base package -> optional-deps
- split synth and visualization into different modules for easier handling

Closes: #1052 #1540 #1545


doctr/file_utils.py
```diff
@@ -82,6 +78,26 @@
...
+
+    Args:
+    ----
+        name: name of the package
+        extra_message: additional message to display if the package is not found
+    """
+    try:
+        _ = importlib.metadata.version(name)
+    except importlib.metadata.PackageNotFoundError:
+        raise ValueError(
```
felixdittrich92, line 93: Should be ImportError

.github/workflows/public_docker_images.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       matrix:
         # Must match version at https://www.python.org/ftp/python/
         python: ["3.9.18", "3.10.13", "3.11.8"]
-        framework: ["tf", "torch"]
+        framework: ["tf", "torch", "tf-extras", "torch-extras"]
```
felixdittrich92, line 26: @frgfm @odulcy-mindee Not sure do we want this as "prod" "non prod" (with -extras) !? :sweat_smile:

pyproject.toml
```diff
@@ -66,6 +62,23 @@ torch = [
     "torchvision>=0.13.0",
     "onnx>=1.12.0,<3.0.0",
 ]
+tf-extras = [
```
felixdittrich92, line 65: @frgfm Maybe another idea in mind i really really don't like all these dups :sweat_smile: (Yeah i know it's handled by dependabot but yeah ^^)

felixdittrich92: Something like
```
tf-extras = [
     **python-doctr["tf"] 
     **python-doctr["extras"]
]
```
would be awesome from a deps dict

pyproject.toml
```diff
@@ -32,24 +32,20 @@ classifiers=[
...
 dynamic = ["version"]
 dependencies = [
-    "importlib_metadata",
     # For proper typing, mypy needs numpy>=1.20.0 (cf. https://github.com/numpy/numpy/pull/16515)
     # Additional typing support is brought by numpy>=1.22.4, but core build sticks to >=1.16.0
     "numpy>=1.16.0,<2.0.0",
     "scipy>=1.4.0,<2.0.0",
     "h5py>=3.1.0,<4.0.0",
-    "opencv-python>=4.5.0,<5.0.0",
+    "opencv-python-headless>=4.5.0,<5.0.0",
```
frgfm, line 40: I think we should have split that PR because this change on its own could have been separated from the rest, and simply speeds up installation :+1: 

But let's see if we can make it

pyproject.toml
frgfm: Yup, we can't afford to have all of those. Otherwise it complexifies the installation for users and I bet we'll get more GH issue about resulting unexpected behaviours :/

frgfm: But why don't we simply do a separate extra just for weasyprint, mpl & mplcursors? because we can install multiple extras at once, and we can easily adjust installation instructions

frgfm (commented):
Thanks for the PR! I haven't reviewed all files, only the extra def part, which I think we should handle before the rest. Probably doing separate orthogonal extras, and then figuring out a good logic to combine those or add proper installation instructions

pyproject.toml
felixdittrich92: Yeah i agree moved to #1558

felixdittrich92:
> Thanks for the PR! I haven't reviewed all files, only the extra def part, which I think we should handle before the rest. Probably doing separate orthogonal extras, and then figuring out a good logic to combine those or add proper installation instructions

Mh i think we should keep the dependency split and handling the optional dependencies in the same PR (at the end it's not much new logic) because it's in context and so it's easier to get back to it.
About the different install options i was also thinking about but at the end i would say it's better to reduce the different options to base and full (with optional deps) because otherwise the user can install optional dependencies on it's own if required/needed (the exception messages are clear enough and easy to understand additional this impacts only `.show()` and `.from_url()` which is also updated in the readme with a comment `# requires weasyprint to be installed`).
For the future i don't think that we will add much more optional dependencies so i would say it's better to keep the installation jungle to a minimum :sweat_smile: 
@odulcy-mindee wdyt ?

pyproject.toml
odulcy-mindee: > But why don't we simply do a separate extra just for weasyprint, mpl & mplcursors? because we can install multiple extras at once, and we can easily adjust installation instructions

:+1:

odulcy-mindee (changes requested):

felixdittrich92:
moved to `viz` and `html` as options :) 

i would like to keep the "full" installations for the CI jobs instead of cherry picking the install versions depending on the jobs 

api tests would fail if there is an export/import problem because the api installs only python-doctr[tf] without weasyprint / matplotlip & mplcursors

odulcy-mindee (approved):
LGTM. Waiting for @frgfm's comment

felixdittrich92:
@frgfm feel free to comment later on 😊 @odulcy-mindee as discussed merge today

README.md
```diff
@@ -350,9 +341,15 @@ Your API should now be running locally on your port 8002. Access your automatica
...
+headers = {"accept": "application/json"}
+params = {"det_arch": "db_resnet50", "reco_arch": "crnn_vgg16_bn"}
+
 with open('/path/to/your/doc.jpg', 'rb') as f:
-    data = f.read()
-response = requests.post("http://localhost:8002/ocr", files={'file': data}).json()
+    files = [  # application/pdf, image/jpeg, image/png supported
+        ("files", ("doc.jpg", f.read(), "image/jpeg")),
+    ]
+print(requests.post("http://localhost:8080/ocr", headers=headers, params=params, files=files).json())
```
frgfm, line 352: In my experience, it's better to have the shortest snippet possible for users to try it out. We went from 3 LoC to 5 excluding formatting, and this PR doesn't change the minimal backend API so I'm confused as to why there is a change in the "files" passed :thinking: (if that was changed in another PR, I think we should avoid putting it in this PR :sweat_smile: )

A few suggestions:
- either we identify the N°1 thing people want to modulate in our snippet (say model or file type) and we pass explicitly without changing much (the path to the file in my opinion)
- or we stick to the original version

Questions/notes:
- there is a typo above where we pass the port as env var (doesn't work actually)
- is the accept header really necessary here? params aren't JSON data (it formats the URL) and I don't think files are as well
- are the params necessary? Let's perhaps leave the default?
- assuming some people will copy paste, I'd suggest storing the result in `response` and optionally print it so that people don't have to run it again to get the response

This is a minor detail in the scope of the PR, but maintaining accessibility & readibility of snippet is important for adoption :)

docs/source/getting_started/installing.rst
```diff
@@ -19,17 +19,6 @@ For MacBooks with M1 chip, you will need some additional packages or specific ve
...
-
-For MacOS users, you can install them using `Homebrew <https://brew.sh/>`_ as follows:
-
-.. code:: shell
-
-    brew install cairo pango gdk-pixbuf libffi
-
-For Windows users, those dependencies are included in GTK. You can find the latest installer over `here <https://github.com/tschoonj/GTK-for-Windows-Runtime-Environment-Installer/releases>`_.
-
-
```
frgfm, line 32: cd previous comment

frgfm (commented):
Thanks for the PR! Late comments, nothing major :)

README.md
```diff
@@ -136,16 +137,6 @@ The KIE predictor results per page are in a dictionary format with each key repr
...
-Since we use [weasyprint](https://weasyprint.org/), you will need extra dependencies if you are not running Linux.
-
-For MacOS users, you can install them as follows:
-
-```shell
-brew install cairo pango gdk-pixbuf libffi
-```
-
-For Windows users, those dependencies are included in GTK. You can find the latest installer over [here](https://github.com/tschoonj/GTK-for-Windows-Runtime-Environment-Installer/releases).
-
```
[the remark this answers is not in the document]
felixdittrich92: It's still needed (mac & windows) but now the error will directly provide a link to the weasyprint installation docs :)
https://github.com/mindee/doctr/blob/bcd521463b9631c1fbf12b1aa0d123e5f1a5a7d7/doctr/io/reader.py#L61

Maybe in the special case of weasyprint we should also catch the OSError while importing weasyprint #1545 in 
https://github.com/mindee/doctr/blob/bcd521463b9631c1fbf12b1aa0d123e5f1a5a7d7/doctr/file_utils.py#L80

docs/source/getting_started/installing.rst
felixdittrich92: same as above :)

README.md
felixdittrich92: Correct the API has changed in #1522 (with multi file and params support)

Yeah i agree that has nothing to do with the PR (it's something i missed with the API change PR) 

1. Let me check about the port :)
2. and 3.  params & header are optional but about the params i think this shows well that a user can paramterize the request !?
4. Yeah sure :)

frgfm: Alright, the new syntax is a bit complex for file handling :sweat_smile: 

Only one final thing: I did use the right port but when uploading a file using this snippet, I'm getting errors on the backend side saying it cannot access the `.content_type` of the `UploadFile`. I haven't the other PR, so I don't know whether this is some fastapi/starlette version issue or a backend issue but I think there is a problem (if you managed to run it locally, it's quite likely this is a version constraint issue)

Suvoo:
> This PR:
> 
> * exclude weasyprint / matplotlib & mplcursors from base package -> optional-deps
> * split synth and visualization into different modules for easier handling
> 
> Any feedback is welcome 🤗
> 
> Closes: #1052 #1540 #1545

Hey, Are there any plans to release this update soon? As the current "pip versions" have the weasyprint issue or else are you planning to keep this only a part of building from source (cloning the repo)?

felixdittrich92:
> > This PR:
> > 
> > * exclude weasyprint / matplotlib & mplcursors from base package -> optional-deps
> > * split synth and visualization into different modules for easier handling
> > 
> > Any feedback is welcome 🤗
> > Closes: #1052 #1540 #1545
> 
> Hey, Are there any plans to release this update soon? As the current "pip versions" have the weasyprint issue or else are you planning to keep this only a part of building from source (cloning the repo)?

Hey @Suvoo 👋,
We have planned to do the next release end of July / beginning August :)

Suvoo:
Thanks, will use source build till then👍

Issue #1460, opened 2024-02-09
[orientation] Add page orientation interface & integration


felixdittrich92:
### 🚀 The feature

To finalize the orientation detection / straighten pages we need a general page orientation model/interface in combination with the current orientation detection from the models bin_map (works in range ~ -90 - 90 degrees)

- Add interface & integration

### Motivation, pitch

x

PR #1553, opened 2024-04-15
[orientation] page orientation improvements

felixdittrich92:
This PR:

- refactor `OCRPredictor` logic (limit duplication)
- extend estimate angle function to handle also upside-down images
- correct return value from estimate angle (negative - left side rot / positive - right side rot) -> update tests
- integrate `page_orientation_predictor` 

NOTE: Draft until TF orientation models are merged

Closes: #1460

### DIFF
--- a/doctr/models/_utils.py
+++ b/doctr/models/_utils.py
@@ -75,7 +75,7 @@ def estimate_orientation(
         k_x = max(1, (floor(w / 100)))
         k_y = max(1, (floor(h / 100)))
         kernel = cv2.getStructuringElement(cv2.MORPH_RECT, (k_x, k_y))
-        thresh = cv2.dilate(thresh, kernel, iterations=1)  # type: ignore[assignment]
+        thresh = cv2.dilate(thresh, kernel, iterations=1)
 
     # extract contours
     contours, _ = cv2.findContours(thresh, cv2.RETR_LIST, cv2.CHAIN_APPROX_SIMPLE)
--- a/doctr/models/_utils.py
+++ b/doctr/models/_utils.py
@@ -98,7 +98,7 @@ def estimate_orientation(
     if general_page_orientation and general_page_orientation[1] >= min_confidence:
         estimated_angle = estimated_angle + general_page_orientation[0]
 
-    return estimated_angle
+    return -estimated_angle  # return the clockwise angle
 
 
 def rectify_crops(
@@ -107,12 +107,10 @@ def rectify_crops(
 ) -> List[np.ndarray]:
     """Rotate each crop of the list according to the predicted orientation:
     0: already straight, no rotation
-    1: 90 ccw, rotate 3 times ccw
-    2: 180, rotate 2 times ccw
-    3: 270 ccw, rotate 1 time ccw
+    1: 90 cw, rotate 3 times ccw
+    2: 180 cw, rotate 2 times ccw
+    3: 270 cw, rotate 1 times ccw
     """
-    # Inverse predictions (if angle of +90 is detected, rotate by -90)
-    orientations = [4 - pred if pred != 0 else 0 for pred in orientations]
     return (
         [crop if orientation == 0 else np.rot90(crop, orientation) for orientation, crop in zip(orientations, crops)]
         if len(orientations) > 0
--- a/doctr/models/_utils.py
+++ b/doctr/models/_utils.py
@@ -62,13 +62,13 @@ def estimate_orientation(
         gray_img = cv2.medianBlur(gray_img, 5)
         thresh = cv2.threshold(gray_img, thresh=0, maxval=255, type=cv2.THRESH_BINARY_INV + cv2.THRESH_OTSU)[1]
     else:
-        thresh = img.astype(np.uint8)
+        thresh = img.astype(np.uint8)  # type: ignore[assignment]
 
     page_orientation, orientation_confidence = general_page_orientation or (None, 0.0)
     if page_orientation and orientation_confidence >= min_confidence:
         # We rotate the image to the general orientation which improves the detection
         # No expand needed bitmap is already padded
-        thresh = rotate_image(thresh, -page_orientation)
+        thresh = rotate_image(thresh, -page_orientation)  # type: ignore
     else:  # That's only required if we do not work on the detection models bin map
         # try to merge words in lines
         (h, w) = img.shape[:2]
--- a/doctr/models/_utils.py
+++ b/doctr/models/_utils.py
@@ -128,7 +128,8 @@ def rectify_loc_preds(
     return (
         np.stack(
             [
-                np.roll(page_loc_pred, orientation, axis=0)
+                # shift clockwise
+                np.roll(page_loc_pred, -orientation, axis=0)
                 for orientation, page_loc_pred in zip(orientations, page_loc_preds)
             ],
             axis=0,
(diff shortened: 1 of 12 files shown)


felixdittrich92:
Maybe for better understanding:

To test I took some different images and rotated randomly in range (-180 - 180)
Example:  (The test data contains much more dirty samples as the attached one :sweat_smile:)
Screenshot from 2024-06-06 12-48-09


some results:
```
Page 0
orientation: 10, rnd: 11
Page 1
orientation: -106, rnd: -106
Page 2
orientation: -95, rnd: -96
Page 3
orientation: -132, rnd: -133
Page 4
orientation: 148, rnd: 147
Page 5
orientation: -99, rnd: -100
Page 6
orientation: 138, rnd: -42
Page 7
orientation: -118, rnd: -117
```

doctr/models/_utils.py
```diff
@@ -29,42 +31,63 @@ def get_max_width_length_ratio(contour: np.ndarray) -> float:
...
+        the estimated angle of the page (clockwise, negative for left side rotation, positive for right side rotation)
     """
     assert len(img.shape) == 3 and img.shape[-1] in [1, 3], f"Image shape {img.shape} not supported"
-    max_value = np.max(img)
-    min_value = np.min(img)
-    if max_value <= 1 and min_value >= 0 or (max_value <= 255 and min_value >= 0 and img.shape[-1] == 1):
-        thresh = img.astype(np.uint8)
-    if max_value <= 255 and min_value >= 0 and img.shape[-1] == 3:
+    thresh = None
+    # Convert image to grayscale if necessary
```
odulcy-mindee, line 61: ok, so here, no more test on values to guess if it's grayscale or not

doctr/models/_utils.py
```diff
@@ -29,42 +31,63 @@ def get_max_width_length_ratio(contour: np.ndarray) -> float:
...
-    # try to merge words in lines
-    (h, w) = img.shape[:2]
-    k_x = max(1, (floor(w / 100)))
-    k_y = max(1, (floor(h / 100)))
-    kernel = cv2.getStructuringElement(cv2.MORPH_RECT, (k_x, k_y))
-    thresh = cv2.dilate(thresh, kernel, iterations=1)
+    else:
+        thresh = img.astype(np.uint8)  # type: ignore[assignment]
+
+    page_orientation, orientation_confidence = general_page_orientation or (None, 0.0)
```
odulcy-mindee, line 69: Here, `None` can also be `0` for `page_orientation` ?

doctr/models/predictor/base.py
```diff
@@ -29,27 +29,83 @@ class _OCRPredictor:
...
         self.doc_builder = DocumentBuilder(**kwargs)
         self.preserve_aspect_ratio = preserve_aspect_ratio
         self.symmetric_pad = symmetric_pad
         self.hooks: List[Callable] = []
 
+    def _general_page_orientations(
+        self,
+        pages: List[np.ndarray],
+    ) -> List[Tuple[int, float]]:
+        _, classes, probs = zip(self.page_orientation_predictor(pages))  # type: ignore[misc]
```
odulcy-mindee, line 66: I don't why `Codacy` complains about this line. Here, `page_orientation_predictor` is not `None` no ?

doctr/models/predictor/base.py
```diff
@@ -29,27 +29,83 @@ class _OCRPredictor:
...
+            origin_pages_orientations
+            if origin_pages_orientations
+            else [
+                estimate_orientation(seq_map, general_orientation)
+                for seq_map, general_orientation in zip(seg_maps, general_pages_orientations)
+            ]
+        )
+        return [
+            # We exapnd if the page is wider than tall and the angle is 90 or -90
+            rotate_image(page, angle, expand=page.shape[1] > page.shape[0] and abs(angle) == 90)
```
odulcy-mindee, line 105: Do you need to apply `round` on `angle` if, let's say `angle = 87`, or should it stay like that ?

doctr/models/_utils.py
felixT2K: yep because the `general_page_orientation` is optional to keep it backward compatible

doctr/models/predictor/base.py
felixdittrich92: Yeah i saw this also not sure why it is not None :)
I extended also the test case in this PR to check that everything is correct :)

doctr/models/predictor/base.py
felixdittrich92: `estimate_orientation` returns an int and for `straighten_pages` we want to rotate by the exact angle - so `round` would distort the behavior :)

doctr/models/_utils.py
felixdittrich92: Mh a few points about `estimate_orientation`:

- i have tried to simplify some logic / by keeping it backward compatible best as possible - from my view
- so it will still work with bitmaps / RGB images also without the results from the page orientation classifier


If you see anything else or have ideas for optimizations, I would be grateful for any recommendations :hugs:

felixdittrich92: this part does the same as before also for grayscale images but with easier logic to check :sweat_smile:

doctr/models/_utils.py
```diff
@@ -75,10 +98,20 @@ def estimate_orientation(img: np.ndarray, n_ct: int = 50, ratio_threshold_for_li
...
         median = -median_low(angles)
-        return round(median) if abs(median) != 0 else 0
+        estimated_angle = -round(median) if abs(median) != 0 else 0
+
+    # combine with the general orientation and the estimated angle
+    if page_orientation and orientation_confidence >= min_confidence:
+        # special case where the estimated angle is mostly wrong:
+        # when estimated angle is + or - 90 degree, we should consider the general orientation
+        if abs(estimated_angle) == 90 and abs(page_orientation) in [0, 90]:
+            return page_orientation
```
odulcy-mindee, line 111: I don't understand entirely this section. Here, `estimated_angle == 90`, so according to your comment, it's wrong, so `page_orientation` is returned but it's also `90`. It means there is no evident fallback to fix this issue ?

felixdittrich92: yeah the problem case is if estimated_angle is exact 90 or -90 it's mostly not correct (swapping - and + or complete wrong)

So we should check the page orientation also..

I think we can simplify this:

```
if abs(estimated_angle) == abs(page_orientation):
    return page_orientation
```

wdyt ?

odulcy-mindee: But what "mostly wrong" means here ? Based on your experimentations, if it's 90 or -90, it's completly wrong ?

felixdittrich92: In the most cases yes

felixdittrich92: @odulcy-mindee 

pushed a small update 

tested again with a complex set of documents (mobile phone, receipts, etc) 30 samples

randomly rotated in range -180 - 180

it mostly varies between 22 - 25 correct from 30 

before this integration / changes on the same set we have had 7 - 11 correct so as mentioned still some space for further improvements it's not perfect but already much better as before :)

felixdittrich92: ah.. if i lower the min_confidence we get ~27 - 29 correct xD trust the page orientation model xD

odulcy-mindee (approved):

doctr/models/_utils.py
odulcy-mindee: ok, it sounds ok for a "first" iteration. We can improve later in another PR

felixdittrich92: yep i think the same

PR #1558, opened 2024-04-18
[CI] update CI actions

felixdittrich92:
This PR:

- includes CI action updates from deprecated node 16 to node 20


felixdittrich92:
@odulcy-mindee for the PR it's fine but could you check afterwards that the `CODECOV_TOKEN` is set in the secrets ?

.github/workflows/main.yml
```diff
@@ -113,9 +113,10 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/download-artifact@v3
+      - uses: actions/download-artifact@v4
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
         with:
           flags: unittests
           fail_ci_if_error: true
+          token: ${{ secrets.CODECOV_TOKEN }}
```
odulcy-mindee, line 122: This secret does not exist. What should I do to get one ?

pyproject.toml
```diff
@@ -38,7 +38,7 @@ dependencies = [
     "numpy>=1.16.0,<2.0.0",
     "scipy>=1.4.0,<2.0.0",
     "h5py>=3.1.0,<4.0.0",
-    "opencv-python>=4.5.0,<5.0.0",
+    "opencv-python-headless>=4.5.0,<5.0.0",
```
odulcy-mindee, line 41: This change would require a little more thought. Here, it means that we can no longer use the view methods otherwise it will generate an error for users and they will have to reinstall/overwrite the current installation with `opencv-python`.

I'm more in favour of keeping the non-headless package for now and find a way to have headless by default and have an optional dependency so it's not `opencv-python-headless` but `opencv-python`.
It's harder than having just an optional dependancy: it should replace `opencv-python-headless` by the other one and I'm not sure we can do that with `pyproject.toml`.

odulcy-mindee (changes requested):

.github/workflows/main.yml
felixT2K: @odulcy-mindee login at https://about.codecov.io/ and copy the token  (repos -> doctr -> settings)  the sec token is it which needs to be added to the repo secrets :)

pyproject.toml
felixT2K: > This change would require a little more thought. Here, it means that we can no longer use the view methods otherwise it will generate an error for users and they will have to reinstall/overwrite the current installation with `opencv-python`.
> 
> I'm more in favour of keeping the non-headless package for now and find a way to have headless by default and have an optional dependency so it's not `opencv-python-headless` but `opencv-python`. It's harder than having just an optional dependancy: it should replace `opencv-python-headless` by the other one and I'm not sure we can do that with `pyproject.toml`.

you mean the visualization we use at doctr ? That's done with matplotlib so doctr itself doesn't requires opencv GUI support :)

felixT2K: ok tested and the behaviour is strange so yeah you are right would require something like `python-doctr[headless]`

Switch PR to only CI updates :)

.github/workflows/main.yml
odulcy-mindee: done

odulcy-mindee (approved):

PR #1567, opened 2024-04-26
[orientation] augment angle while training

felixdittrich92:
This PR:

- augment randomly the angle while training (+-25 degree) (expand if not directly 0, 90, 180 or -90)
- the idea: the current orientation models works pretty well if the real angle of the page or crop is close to [0, 90, 180, -90] but has some lack with larger deviations like >+- 5 degree

@odulcy-mindee We should test it on the TF orientation model training runs i think this will improve the robustness a lot and then do the same again with the PT models (or we test it directly with the PT models)

### DIFF
--- a/references/classification/train_pytorch_orientation.py
+++ b/references/classification/train_pytorch_orientation.py
@@ -43,7 +43,7 @@ def rnd_rotate(img: torch.Tensor, target):
     idx = CLASSES.index(angle)
     # augment the angle randomly with a probability of 0.5
     if np.random.rand() < 0.5:
-        angle += float(np.random.choice(np.arange(-20, 20, 5)))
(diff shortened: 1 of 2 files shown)


odulcy-mindee (commented):
Will the `expand` arg not mess with the input size of the image ? Or will it be correctly resized before entering the training pipeline ?

felixT2K:
> Will the `expand` arg not mess with the input size of the image ? Or will it be correctly resized before entering the training pipeline ?

It's correctly resized before :)
Screenshot from 2024-04-26 15-38-49
Screenshot from 2024-04-26 15-38-15

odulcy-mindee (approved):

PR #1569, opened 2024-04-29
[tests/onnx] Add onnx and model out check

felixdittrich92:
This PR:

- extend the test with missing allclose check for onnx export - non critical - only warn
- rs will output also why tests was skipped (only as information for us)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,7 +32,7 @@ jobs:
           pip install -e .[testing]
       - name: Run unittests
         run: |
-          coverage run -m pytest tests/common/
+          coverage run -m pytest tests/common/ -rs
           coverage xml -o coverage-common.xml
       - uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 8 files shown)


felixdittrich92:
CC @llFireHawkll

felixdittrich92:
> I find the output misleading: https://github.com/mindee/doctr/actions/runs/8875628926/job/24365456870?pr=1569
> 
> ```
> =========================== short test summary info ============================
> SKIPPED [1] tests/pytorch/test_models_detection_pt.py:183: Output of db_mobilenet_v3_large:
> Difference is 9.999999747378752e-06
> SKIPPED [2] tests/pytorch/test_models_recognition_pt.py:116: too less memory
> =========== 235 passed, 3 skipped, 30 warnings in 302.91s (0:05:02) ============
> ```
> 
> Value seems really tiny but it's a mean and not the maximum difference observed, as computed before. Maybe you can write as message that the outputs are not element-wise equal within a tolerance and then provide this mean difference or return the max diff you observe between two arrays.
> 
> BTW, why change the results of this test as `skipped` ?

Yeah we can update the message :)
skipped is only in this special case (because we don't want that it fails here) otherwise the test will still fail or pass as before :)

odulcy-mindee (approved):

felixdittrich92:
@odulcy-mindee missed to add 1 line xD

odulcy-mindee (approved):

llFireHawkll:
Great thanks @felixdittrich92 for this update. Really appreciate your help.

PR #1571, opened 2024-04-29
feat: :sparkles: torch mobilenet_v3_small_crop_orientation

odulcy-mindee:


felixdittrich92 (approved):
If the segmentation is fine the crop orientation prediction was on my test samples ~95% correct :+1: 
Before ~92%

PR #1584, opened 2024-05-05
[transforms] small random resize improvement

felixdittrich92:
This PR:

- choose preserve aspect ratio and symmetric_pad also randomly to cover more cases


felixdittrich92:
> It's a bit curious to have only one parameter to control two behaviors. I mean, why not a parameter for `preserve_aspect_ratio` and `symmetric_pad` which can be a `Union[bool, float]`, with `float` as a parameter to control the probability to set this param ?

Yeah that's a good idea 👍

odulcy-mindee (approved):
Definitively better, thank you !

PR #1588, opened 2024-05-07
[models] Change default model to `fast_base` - soft breaking change

felixdittrich92:
This PR:

- changes the default detection model from `db_resnet50` to `fast_base`
- reason: 
   - more accure (actually much much better also on rotated samples) / needs less memory / small bit faster 
   
NOTE: marked as breaking change because with the next release we should mention this change and to revert to the previous behavior that it only requires to pass `db_resnet50` as det_arch


odulcy-mindee (approved):
Thanks !

PR #1598, opened 2024-05-15
[Fix] close PIL images when loading images to tensor/numpy

helpmefindaname:
The previous behavior opened PIL images, but didn't release their memory, causing a huge memory leak on our training pipeline.
With this PR, the PIL images will be closed immediately after the copy as numpy array or torch tensor is created.


felixdittrich92:
Hi @helpmefindaname :wave:,

Tanks for the fix good catch :+1:

felixdittrich92:
failing build / script jobs are expected atm so don't wonder :sweat_smile:

felixdittrich92 (approved):
Thanks for the fix lgtm :+1:

PR #1625, opened 2024-05-31
[builder] Add objectness scores

felixdittrich92:
This PR:

- unify detection predictor output (N, 5) / (N, 5, 2)  --> last is objectness_score
- add objectness score to Document output
- update API template
- update docs
- add corresponding tests

Use case: enables the end user to filter out some detected artefacts for example

NOTE: API tests will fail until merged :)

Related to: #1612

### DIFF
--- a/doctr/io/elements.py
+++ b/doctr/io/elements.py
@@ -72,10 +72,11 @@ class Word(Element):
         confidence: the confidence associated with the text prediction
         geometry: bounding box of the word in format ((xmin, ymin), (xmax, ymax)) where coordinates are relative to
         the page's size
+        objectness_score: the objectness score of the detection
         crop_orientation: the general orientation of the crop in degrees and its confidence
     """
 
-    _exported_keys: List[str] = ["value", "confidence", "geometry", "crop_orientation"]
+    _exported_keys: List[str] = ["value", "confidence", "geometry", "objectness_score", "crop_orientation"]
     _children_names: List[str] = []
 
     def __init__(
@@ -83,12 +84,14 @@ class Word(Element):
         value: str,
         confidence: float,
         geometry: Union[BoundingBox, np.ndarray],
+        objectness_score: float,
         crop_orientation: Dict[str, Any],
     ) -> None:
         super().__init__()
         self.value = value
         self.confidence = confidence
         self.geometry = geometry
+        self.objectness_score = objectness_score
         self.crop_orientation = crop_orientation
 
     def render(self) -> str:
@@ -148,7 +151,7 @@ class Line(Element):
             all words in it.
     """
 
-    _exported_keys: List[str] = ["geometry"]
+    _exported_keys: List[str] = ["geometry", "objectness_score"]
     _children_names: List[str] = ["words"]
     words: List[Word] = []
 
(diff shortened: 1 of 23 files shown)


doctr/models/builder.py
```diff
@@ -444,14 +470,16 @@ def __call__(  # type: ignore[override]
...
+        objectness_scores: np.ndarray,
         word_preds: List[Tuple[str, float]],
         crop_orientations: List[Dict[str, Any]],
     ) -> List[Prediction]:
         """Gather independent words in structured blocks
 
         Args:
         ----
-            boxes: bounding boxes of all detected words of the page, of shape (N, 5) or (N, 4, 2)
+            boxes: bounding boxes of all detected words of the page, of shape (N, 4) or (N, 4, 2)
```
odulcy-mindee, line 481: Shape was wrong in the docstring ?

doctr/models/detection/differentiable_binarization/base.py
```diff
@@ -150,10 +150,11 @@ def bitmap_to_boxes(
                     raise AssertionError("When assume straight pages is false a box is a (4, 2) array (polygon)")
                 _box[:, 0] /= width
                 _box[:, 1] /= height
-                boxes.append(_box)
+                # Add score to box as (0, score)
+                boxes.append(np.vstack([_box, np.array([0.0, score])]))
```
odulcy-mindee, line 154: What is 0 ?

doctr/io/elements.py
```diff
@@ -148,15 +151,19 @@ class Line(Element):
...
 
     def __init__(
         self,
         words: List[Word],
         geometry: Optional[Union[BoundingBox, np.ndarray]] = None,
+        objectness_score: Optional[float] = None,
     ) -> None:
+        # Compute the objectness score of the line
+        if objectness_score is None:
+            objectness_score = float(np.mean([w.objectness_score for w in words]))
```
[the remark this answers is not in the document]
felixdittrich92: We need a float here and `np.mean` returns a `ndarray` :)
ref.: https://numpy.org/doc/stable/reference/generated/numpy.mean.html

doctr/models/builder.py
felixdittrich92: We detach the scores in the OCRPredictor so afterwards it is (N, 4) / (N, 4, 2) instead of (N, 5) / (N, 4, 2) afterwards .. so before the score was only "dragged along unused"

As mentioned it's unified now with `assume_straight_pages=False | True` :)

doctr/models/detection/differentiable_binarization/base.py
felixdittrich92: That's only a placeholder otherwise it wouldn't be possible to transport the objectness score here because with `assume_straight_pages=True` we have (N, 5) last is score and with `assume_straight_pages=False` we now have also (N, 5, 2) where score is [0.0, score]  up to the OCRPredictor where we detach it and go further with the real `loc_preds` (N, 4) / (N, 4, 2)

doctr/io/elements.py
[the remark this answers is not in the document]
odulcy-mindee: @felixdittrich92 Well, in the examples, `float` is not used to get the mean on the arrays. I don't mind to keep it like that, I thought there was an issue

doctr/models/builder.py
odulcy-mindee: Yeah, indeed, I understand that after a closer reading of this PR

doctr/models/detection/differentiable_binarization/base.py
odulcy-mindee: Ah ok, so it's "useless" here

odulcy-mindee (approved):

doctr/models/detection/differentiable_binarization/base.py
felixT2K: correct ^^
see it as split between the loc_preds and the score :)

PR #1635, opened 2024-06-06
[bug] remove TF multiprocessing workers

felixdittrich92:
This PR:

- Remove garbage TF workers -> crashes on GPU

@odulcy-mindee as discussed ^^

### DIFF
--- a/doctr/models/preprocessor/tensorflow.py
+++ b/doctr/models/preprocessor/tensorflow.py
@@ -10,7 +10,6 @@ import numpy as np
 import tensorflow as tf
 
 from doctr.transforms import Normalize, Resize
-from doctr.utils.multithreading import multithread_exec
 from doctr.utils.repr import NestedObject
 
 __all__ = ["PreProcessor"]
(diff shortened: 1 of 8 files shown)


doctr/models/preprocessor/tensorflow.py
```diff
@@ -113,13 +114,13 @@ def __call__(self, x: Union[tf.Tensor, np.ndarray, List[Union[tf.Tensor, np.ndar
 
         elif isinstance(x, list) and all(isinstance(sample, (np.ndarray, tf.Tensor)) for sample in x):
             # Sample transform (to tensor, resize)
-            samples = list(multithread_exec(self.sample_transforms, x))
+            samples = list(multithread_exec(self.sample_transforms, x, threads=1 if self._runs_on_cuda else None))
```
felixdittrich92, line 117: For the inference pipeline we can use threading because on CPU it works without issues :) 
So only disable if it runs on GPU

odulcy-mindee (approved):
Thank you @felixdittrich92 ! Hopefully, TF will work as expected

felixT2K:
> Thank you @felixdittrich92 ! Hopefully, TF will work as expected

Yeah that's not optimal (for training) but for the moment better as crashing all the time ^^

PR #1657, opened 2024-06-24
feat:Alignment during recontruction of image

SkaarFacee:


felixdittrich92:
Hi @SkaarFacee :wave:,

Thanks for the PR.

doctr/utils/reconstitution.py
```diff
@@ -38,14 +38,18 @@ def synthesize_page(
     # Draw each word
     for block in page["blocks"]:
         for line in block["lines"]:
+            line_ymin = min(int(round(h * word["geometry"][0][1])) for word in line["words"])
```
felixdittrich92, line 41: I would suggest the following:

```
def synthesize_page(
    page: Dict[str, Any],
    draw_proba: bool = False,
    font_family: Optional[str] = None,
    adjust_to_line: bool = False,
) -> np.ndarray:
    """Draw a the content of the element page (OCR response) on a blank page.

    Args:
    ----
        page: exported Page object to represent
        draw_proba: if True, draw words in colors to represent confidence. Blue: p=1, red: p=0
        font_size: size of the font, default font = 13
        font_family: family of the font
        adjust_to_line: if True, adjust y coordinates to line geometry

    Returns:
    -------
        the synthesized page
    """
    # Draw template
    h, w = page["dimensions"]
    response = 255 * np.ones((h, w, 3), dtype=np.int32)

    # Draw each word
    for block in page["blocks"]:
        multiline = len(block["lines"]) > 1
        for line in block["lines"]:
            for word in line["words"]:
                # Get absolute word geometry
                (xmin, ymin), (xmax, ymax) = word["geometry"]
                xmin, xmax = int(round(w * xmin)), int(round(w * xmax))

                if multiline and adjust_to_line:
                    ymin = int(round(h * line["geometry"][0][1]))
                    ymax = int(round(h * line["geometry"][1][1]))
                else:
                    ymin, ymax = int(round(h * ymin)), int(round(h * ymax))
```
In this case the user can still decide and adjusting makes only sense if we have lines (so `resolve_lines=True`)

doctr/utils/reconstitution.py
```diff
@@ -38,14 +38,18 @@ def synthesize_page(
...
             for word in line["words"]:
                 # Get absolute word geometry
                 (xmin, ymin), (xmax, ymax) = word["geometry"]
                 xmin, xmax = int(round(w * xmin)), int(round(w * xmax))
                 ymin, ymax = int(round(h * ymin)), int(round(h * ymax))
-
                 # White drawing context adapted to font size, 0.75 factor to convert pts --> pix
-                font = get_font(font_family, int(0.75 * (ymax - ymin)))
+                ymin, ymax = line_ymin, line_ymax
+                calculate_font_size = int(0.75 * (ymax - ymin))
```
felixdittrich92, line 50: This does still not work well see:

- does still overlap 


Screenshot from 2024-06-25 09-29-07
test_page
Screenshot from 2024-06-25 09-30-29

felixdittrich92 (changes requested):

doctr/utils/reconstitution.py
SkaarFacee: Okay, I will do that right away

doctr/utils/reconstitution.py
SkaarFacee: That is interesting. I shall debug this and see why the issue exists

SkaarFacee: Hey there, what models did you use for these?

felixdittrich92: Hey :)

`fast_base` and `parseq` and `db_mobilenet_v3_large` and `crnn_mobilenet_v3_large`

felixdittrich92:
#1750

PR #1676, opened 2024-07-22
[datasets] Added Bangla characters to VOCABS

bliss22:
This PR:

- Add new vocab (bangla)
- Add corresponding docs entry

### DIFF
--- a/doctr/datasets/vocabs.py
+++ b/doctr/datasets/vocabs.py
@@ -20,6 +20,8 @@ VOCABS: Dict[str, str] = {
     "hindi_digits": "٠١٢٣٤٥٦٧٨٩",
     "arabic_diacritics": "ًٌٍَُِّْ",
     "arabic_punctuation": "؟؛«»—",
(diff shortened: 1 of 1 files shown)


doctr/datasets/vocabs.py
```diff
@@ -66,6 +69,7 @@
         + VOCABS["danish"]
         + VOCABS["finnish"]
         + VOCABS["swedish"]
+        + VOCABS["bangla"]
```
felixdittrich92, line 72: Let's remove this ftm, because we don't have bangla data atm we could use to train a multilingual model which includes this vocab :+1:

felixdittrich92 (changes requested):
Hi @bliss22 :wave:,

Thanks for the PR :+1: 

One small change and please add the new vocab also to the docs:

https://github.com/mindee/doctr/blob/main/docs/source/modules/datasets.rst

felixdittrich92:
Closed by #1687

PR #1690, opened 2024-08-08
Add torch.compile support for pytorch 2.4

Fabioomega:
Added support for torch.compile only for version 2.4 or higher of pytorch. Included support for all the detection models and a recognition model (parseq).

Unfortunately, triton support is only available on linux plataforms. WSL seems to work fine tough, so it may be used that way.

Example use:

1. Enable the feature by setting the enviroment variable: `USE_TRITON = YES`.
2. Try the following code:

```python
from doctr.io import Document
from cv2 import imread
from time import time

t1 = time()
reader = ocr_predictor(det_arch='db_resnet50', pretrained=True)
print('Loading time:', time() - t1)

img = imread("<img>")
t2 = time()
d: Document = reader([img])
print('Document of the first try ', d)
print('Processing time of the first try', time() - t2)
t3 = time()
d: Document = reader([img])
print('Processing time of the second try:', time() - t3)
```


doctr/models/recognition/parseq/pytorch.py
```diff
@@ -266,7 +266,8 @@ def decode_autoregressive(self, features: torch.Tensor, max_len: Optional[int] =
         ).int()
 
         pos_logits = []
-        for i in range(max_length):
+        i = 0
+        while i < max_length:
```
felixdittrich92, line 270: I remember there was a issue with `while` loops by exporting to onnx so we have to be careful here (needs to be checked)

doctr/models/detection/_utils/pytorch_compile.py
felixdittrich92, line 1: Optimal would be if we let it as is (so without modifying the post proc) and compile only the models without the post proc.
For example:
```
det_model = fast_tiny(pretrained=True, compile=True, compile_kwargs={..}) # Could be handled with the kwargs
```

In this case the user would have the full control.
This would require only minimal changes to the interface and we can document it well.
Wdyt ?

doctr/models/recognition/parseq/pytorch.py
Fabioomega: I changed it because it was some unecessary complication related to breaks in torch.compile. Changing to a while loop and changing the logic a bit helped. Hopefully it works for the onnx also

doctr/models/detection/_utils/pytorch_compile.py
Fabioomega: I like the ideia. I think it shouldn't be too hard to implement. Will try my hand at it today.

doctr/file_utils.py
```diff
@@ -76,6 +75,18 @@
         " is installed and that either USE_TF or USE_TORCH is enabled."
     )
 
+if _torch_available:
```
felixdittrich92, line 78: @Fabioomega We can remove this.

2 options:

We pin the lower boundary to >= 2.0.0 here https://github.com/mindee/doctr/blob/9045dcfc9c5c837b06fcda8e802f7cf1d95bd18c/pyproject.toml#L61
https://github.com/mindee/doctr/blob/9045dcfc9c5c837b06fcda8e802f7cf1d95bd18c/pyproject.toml#L105
and `torchvision>=0.15.0`

or we mention in the docs that this requires >= 2.0.0 for compile and >=2.4.0 for compile + fullgraph

@odulcy-mindee wdyt ? 
We are already at 2.4.0 so i would prefer the >=2.0.0 pin

doctr/file_utils.py
```diff
@@ -104,3 +115,11 @@ def is_torch_available():
 def is_tf_available():
     """Whether TensorFlow is installed."""
     return _tf_available
+
+def does_torch_have_compile_capability():
```
felixdittrich92, line 119: remove :)

doctr/models/classification/zoo.py
felixdittrich92, line 1: can be reverted complete

doctr/models/detection/_utils/__init__.py
felixdittrich92, line 1: revert :)

doctr/models/detection/fast/base.py
felixdittrich92, line 1: revert :)

doctr/models/detection/linknet/base.py
felixdittrich92, line 1: revert :)

doctr/models/detection/zoo.py
felixdittrich92, line 1: revert :)

doctr/models/recognition/parseq/pytorch.py
felixdittrich92, line 1: revert :)

doctr/models/recognition/zoo.py
felixdittrich92, line 1: revert :)

doctr/models/zoo.py
felixdittrich92, line 1: revert :)

tests/pytorch/test_models_detection_pt.py
```diff
@@ -186,3 +186,46 @@ def test_models_onnx_export(arch_name, input_shape, output_size):
         assert np.allclose(pt_logits, ort_outs[0], atol=1e-4)
     except AssertionError:
         pytest.skip(f"Output of {arch_name}:\nMax element-wise difference: {np.max(np.abs(pt_logits - ort_outs[0]))}")
+
+@pytest.mark.skipif(not does_torch_have_compile_capability(), reason="requires pytorch >= 2.0.0")
+@pytest.mark.skipif(not is_pytorch_backend_available(), reason="requires pytorch backend to be available")
```
felixdittrich92, line 191: remove the first two `skipif` Ci runs always on latest pytorch - same for the other tests :+1:

felixdittrich92 (changes requested):
Hi @Fabioomega :wave: 

Tests looks already good to me :+1: 
Docs section missing here: https://github.com/mindee/doctr/blob/main/docs/source/using_doctr/using_model_export.rst

As mentioned a table would be great :)
For the `classification` models it would be enough to add both `orientation` models to the table (i don't think we should blow up the table by adding all the backbone models)


Todo's:

- [ ] unittests
- [ ] docs

Left some comments to revert unrequired parts :)

As mentioned for follow up PR's we can focus on fixes for the models which does not work yet out of the box :+1:

doctr/models/zoo.py
Fabioomega: I hame some questions about that! Wasn't the original ideia to add a new argument to enable compilation? Did I misunderstood?

felixT2K: That was the first thought as your code looked like changes to the pipeline/models were needed. However, we then saw that these were not needed.
Which is why we only add tests here and a section on how to use it. The compilation therefore remains on the user side, which is at the same time much more flexible. :)
Additional this avoids to add a arg which at the end only does -> `model = torch.compile(model, ..)`  and is backend depending (PyyTorch).

felixT2K: A full sample would look like then for example:

```
import requests
import torch
from doctr.models import ocr_predictor, parseq, fast_base

bytes_data = requests.get(
    "https://i1.rgstatic.net/publication/231831562_Another_Boring_Day_in_Paradise_Rock_and_Roll_and_the_Empowerment_of_Everyday_Life/links/57d02a2408ae601b39a05636/largepreview.png"
).content

doc = DocumentFile.from_images([bytes_data])

rec_model = torch.compile(parseq(pretrained=True))
det_model = torch.compile(fast_base(pretrained=True))
predictor = ocr_predictor(det_arch=det_model, reco_arch=rec_model, pretrained=True)

res = predictor(doc)
res.show()
```

The only required change here would be to allow also:
`torch._dynamo.eval_frame.OptimizedModule` in https://github.com/mindee/doctr/blob/9045dcfc9c5c837b06fcda8e802f7cf1d95bd18c/doctr/models/recognition/zoo.py#L39 and https://github.com/mindee/doctr/blob/9045dcfc9c5c837b06fcda8e802f7cf1d95bd18c/doctr/models/detection/zoo.py#L59 and https://github.com/mindee/doctr/blob/9045dcfc9c5c837b06fcda8e802f7cf1d95bd18c/doctr/models/classification/zoo.py#L45

felixdittrich92:
#1791

PR #1700, opened 2024-08-21
feat: added ukrainian vocab

holyCowMp3:
Added ukrainian vocab to allow use handwritten UA datasets


doctr/datasets/vocabs.py
```diff
@@ -13,7 +13,7 @@
     "digits": string.digits,
     "ascii_letters": string.ascii_letters,
     "punctuation": string.punctuation,
-    "currency": "£€¥¢฿",
+    "currency": "£€¥¢฿₴",
```
felixdittrich92, line 16: Unfortunately this would break our current models so we need to add it directly to the ukrainian vocab

doctr/datasets/vocabs.py
```diff
@@ -73,6 +75,7 @@
         + VOCABS["danish"]
         + VOCABS["finnish"]
         + VOCABS["swedish"]
+        + VOCABS["ukrainian"]
```
felixdittrich92, line 78: Please remove from the multilingual entry because we don't have a multilingual dataset which includes cyrillic letters atm :)

felixdittrich92 (changes requested):
Hi @holyCowMp3 :wave:,
Thanks for the PR :+1: 

Could you please also add the entry in the documentation see as reference: https://github.com/mindee/doctr/pull/1687

felixdittrich92:
@holyCowMp3 entry in docs missing see: https://github.com/mindee/doctr/pull/1687/files
:)

felixdittrich92:
> Знімок екрана 2024-08-21 о 18 17 39 Do we need to add the generic cyrrilic letters as well?

Yes please :)

felixdittrich92 (approved):
Thanks 🤗

PR #1708, opened 2024-08-28
[orientation] Enable usage of custom trained orientation models

felixdittrich92:
This PR:

- Add ability that users can train and use there own orientation models
- Open interface to load custom orient models
- Add corresponding doc section and tests

NOTE: 
Overwriting is i think a temp solution to avoid blowing up the `ocr_predictor` args until we have a solution for a better interface.

Request: https://github.com/mindee/doctr/discussions/1698

### DIFF
--- a/doctr/models/classification/mobilenet/pytorch.py
+++ b/doctr/models/classification/mobilenet/pytorch.py
@@ -9,12 +9,14 @@ from copy import deepcopy
 from typing import Any, Dict, List, Optional
 
 from torchvision.models import mobilenetv3
+from torchvision.models.mobilenetv3 import MobileNetV3
 
 from doctr.datasets import VOCABS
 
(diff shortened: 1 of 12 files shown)


doctr/models/classification/mobilenet/pytorch.py
```diff
@@ -9,12 +9,14 @@
...
 
 from torchvision.models import mobilenetv3
+from torchvision.models.mobilenetv3 import MobileNetV3
 
 from doctr.datasets import VOCABS
 
 from ...utils import load_pretrained_params
 
 __all__ = [
+    "MobileNetV3",
```
felixdittrich92, line 19: That's required for the `isinstance` check in `classification.zoo`

odulcy-mindee (approved):
Nice PR, looks good to me !

PR #1709, opened 2024-08-29
[build] NumPy 2.0 support

felixdittrich92:
This PR:

- Add numpy 2.0 support

- For TF this will take full effect after we can upgrade to keras v3 (in the meanwhile it falls back to numpy ~1.26.4 handled by pip)


odulcy-mindee (approved):
New numpy support ! Thanks @felixdittrich92 !

PR #1715, opened 2024-09-03
[Bug] Fix eval scripts + possible overflow in Resize

felixdittrich92:
This PR:

- fix for eval scripts with keeping aspect ratio / symmetric pad - optional
- tf: fix deprecated 
- fix small overflow

### DIFF
--- a/doctr/transforms/modules/pytorch.py
+++ b/doctr/transforms/modules/pytorch.py
@@ -77,12 +77,13 @@ class Resize(T.Resize):
                 # Pad image
                 img = pad(img, _pad)
 
-                if self.symmetric_pad:
-                    offset = half_pad[0] / img.shape[-1], half_pad[1] / img.shape[-2]
-
             # In case boxes are provided, resize boxes if needed (for detection task if preserve aspect ratio)
             if target is not None:
                 target = np.clip(target, 0, 1)
+
+                if self.symmetric_pad:
+                    offset = half_pad[0] / img.shape[-1], half_pad[1] / img.shape[-2]
+
                 if self.preserve_aspect_ratio:
                     # Get absolute coords
                     if target.shape[1:] == (4,):
--- a/doctr/transforms/modules/pytorch.py
+++ b/doctr/transforms/modules/pytorch.py
@@ -264,7 +264,11 @@ class RandomResize(torch.nn.Module):
         self.p = p
         self._resize = Resize
 
-    def forward(self, img: torch.Tensor, target: np.ndarray) -> Tuple[torch.Tensor, np.ndarray]:
+    def forward(
+        self,
+        img: torch.Tensor,
+        target: Union[np.ndarray, Dict[str, Union[np.ndarray, List[str]]]],
+    ) -> Tuple[torch.Tensor, Union[np.ndarray, Dict[str, Union[np.ndarray, List[str]]]]]:
         if torch.rand(1) < self.p:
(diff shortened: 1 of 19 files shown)


doctr/transforms/modules/tensorflow.py
```diff
@@ -104,49 +104,66 @@ def extra_repr(self) -> str:
...
+        # In case boxes are provided, resize boxes if needed (for detection task if preserve aspect ratio)
+        if target is not None:
+            # Possible formats:
+            # KIE: Dict[str, np.ndarray]
+            # Built-in datasets: Dict[str, Union[np.ndarray, List[str]]]
+            # Custom datasets: np.ndarray
+
+            if isinstance(target, dict):
+                # Built-in datasets
+                if "boxes" and "labels" in target.keys():
```
felixdittrich92, line 159: Not really happy with this but otherwise it would require a larger refactoring with breaking changes..

doctr/transforms/modules/pytorch.py
```diff
@@ -74,33 +74,42 @@ def forward(
...
                             target[..., 1] *= raw_shape[-2] / img.shape[-2]
                     else:
-                        raise AssertionError
-                return img, target
+                        raise AssertionError("Boxes should be in the format (n_boxes, 4, 2) or (n_boxes, 4)")
+                return np.clip(target, 0, 1)
+
+            # In case boxes are provided, resize boxes if needed (for detection task if preserve aspect ratio)
+            if target is not None:
+                if isinstance(target, dict) and "boxes" in target.keys():
```
felixdittrich92, line 106: Not really happy with ..alternative would be to modify the `sample_transforms` in `scripts/evaluate.py` and `scripts/evaluate_kie.py` @odulcy-mindee wdyt ?

felixdittrich92: For the detection stuff we could in another PR introduce something like `detection_task=True` same way we use already for built-in `recognition_task=..`
If we prefer to modify the the sample_transform instead of the resize call

felixdittrich92: ok updated :sweat_smile:

felixdittrich92:
failing test depends on #1717

doctr/transforms/modules/pytorch.py
```diff
@@ -74,33 +74,34 @@ def forward(
...
                             target[..., 1] = offset[1] + target[..., 1] * raw_shape[-2] / img.shape[-2]
                         else:
                             target[..., 0] *= raw_shape[-1] / img.shape[-1]
                             target[..., 1] *= raw_shape[-2] / img.shape[-2]
                     else:
-                        raise AssertionError
-                return img, target
+                        raise AssertionError("Boxes should be in the format (n_boxes, 4, 2) or (n_boxes, 4)")
+
+                return img, np.clip(target, 0, 1)
```
odulcy-mindee, line 104: Do you remember why you added this `np.clip` on target ?

felixdittrich92: @odulcy-mindee Yeah because several people mentioned that there training on a custom dataset failed ..the reason was a small overflow after modifying the targets

felixdittrich92: Here for example: 
https://github.com/mindee/doctr/discussions/1714

odulcy-mindee: Ok, thank !

odulcy-mindee (approved):

Issue #1455, opened 2024-02-07
Flipped text recognition prediction.


decadance-dance:
### Bug description

When I set the option assume_straight_pages=False, some of the predictions may be turned upside down.
I tried db_resnet34, db_resnet50 and master, parseg. For each pair I observed this bug.

### Code snippet to reproduce the bug

```python
from doctr.models import ocr_predictor
from doctr.io import DocumentFile


input = DocumentFile.from_images("./gh.png")

model = ocr_predictor(
    'db_resnet50', 
    'parseq', 
    pretrained=True,
    assume_straight_pages=False,
).cuda().half()

result = model(input)
print(result)
```

### Error traceback

```
...
Line(
  (words): [
              Word(value='ster', confidence=1.0),
              Word(value='and', confidence=1.0),
              Word(value='Graham', confidence=1.0),
              Word(value='6661]', confidence=0.95),         <-- Should be '[1999'
              Word(value='and', confidence=1.0),
              Word(value='2012],', confidence=1.0),
              Word(value='Gamba', confidence=1.0),
              Word(value='and', confidence=1.0),
              Word(value='Graham', confidence=1.0),
              Word(value='[2018]', confidence=1.0),
              Word(value='and', confidence=1.0),
              Word(value='Axelrod', confidence=1.0),
              Word(value='[2018).', confidence=0.99),
            ]
),
...
```
![gh_mark](https://github.com/mindee/doctr/assets/86170544/2decd0ff-cfea-4b8d-bd91-8237f817a640)

### Environment

Collecting environment information...

DocTR version: 0.8.0a0
TensorFlow version: N/A
PyTorch version: 2.1.0a0+4136153 (torchvision 0.16.0a0)
OpenCV version: 4.9.0
OS: Ubuntu 22.04.2 LTS
Python version: 3.10.6
Is CUDA available (TensorFlow): N/A
Is CUDA available (PyTorch): Yes
CUDA runtime version: 12.1.105
GPU models and configuration: GPU 0: NVIDIA A30
Nvidia driver version: 525.147.05
cuDNN version: Probably one of the following:
/usr/lib/x86_64-linux-gnu/libcudnn.so.8.9.2
/usr/lib/x86_64-linux-gnu/libcudnn_adv_infer.so.8.9.2
/usr/lib/x86_64-linux-gnu/libcudnn_adv_train.so.8.9.2
/usr/lib/x86_64-linux-gnu/libcudnn_cnn_infer.so.8.9.2
[...]

PR #1723, opened 2024-09-12
Feature/assume straight text

milosacimovic:
Modifying the ocr_predictor API to support assume_straight_text as an argument.
When used with assume_straight_pages=False this reduces the reliance on an unreliable crop orientation model when the text is almost straight and additionally reduces speed of execution. It should alleviate the issue mentioned in https://github.com/mindee/doctr/issues/1455 if the use-case is one where the text is straight i.e. no rotations of 90, 180 and 270 degrees.

The main contribution to the pipeline is the logic around a new geometry function which extracts the crops while dewarping the images based on the corners of the text detection, which returns polygons (when assume_straight_pages=False and assume_straight_text=True).

### DIFF
--- a/doctr/models/kie_predictor/base.py
+++ b/doctr/models/kie_predictor/base.py
@@ -36,6 +36,7 @@ class _KIEPredictor(_OCRPredictor):
     def __init__(
         self,
         assume_straight_pages: bool = True,
+        assume_straight_text: bool = False,
         straighten_pages: bool = False,
         preserve_aspect_ratio: bool = True,
         symmetric_pad: bool = True,
@@ -43,7 +44,13 @@ class _KIEPredictor(_OCRPredictor):
         **kwargs: Any,
     ) -> None:
         super().__init__(
-            assume_straight_pages, straighten_pages, preserve_aspect_ratio, symmetric_pad, detect_orientation, **kwargs
+            assume_straight_pages,
+            assume_straight_text,
+            straighten_pages,
+            preserve_aspect_ratio,
+            symmetric_pad,
+            detect_orientation,
+            **kwargs,
         )
 
         self.doc_builder: KIEDocumentBuilder = KIEDocumentBuilder(**kwargs)
--- a/doctr/models/kie_predictor/pytorch.py
+++ b/doctr/models/kie_predictor/pytorch.py
@@ -29,6 +29,8 @@ class KIEPredictor(nn.Module, _KIEPredictor):
         reco_predictor: recognition module
         assume_straight_pages: if True, speeds up the inference by assuming you only pass straight pages
             without rotated textual elements.
+        assume_straight_text: if True, speeds up the inference by assuming you only pass straight text
+            without rotated textual elements.
         straighten_pages: if True, estimates the page general orientation based on the median line orientation.
             Then, rotates page before passing it to the deep learning modules. The final predictions will be remapped
             accordingly. Doing so will improve performances for documents with page-uniform rotations.
@@ -44,6 +46,7 @@ class KIEPredictor(nn.Module, _KIEPredictor):
         det_predictor: DetectionPredictor,
         reco_predictor: RecognitionPredictor,
         assume_straight_pages: bool = True,
+        assume_straight_text: bool = False,
         straighten_pages: bool = False,
         preserve_aspect_ratio: bool = True,
         symmetric_pad: bool = True,
@@ -57,6 +60,7 @@ class KIEPredictor(nn.Module, _KIEPredictor):
         _KIEPredictor.__init__(
             self,
             assume_straight_pages,
+            assume_straight_text,
             straighten_pages,
             preserve_aspect_ratio,
             symmetric_pad,
@@ -129,10 +133,11 @@ class KIEPredictor(nn.Module, _KIEPredictor):
                 dict_loc_preds[class_name],
                 channels_last=channels_last,
                 assume_straight_pages=self.assume_straight_pages,
+                assume_straight_text=self.assume_straight_text,
             )
         # Rectify crop orientation
         crop_orientations: Any = {}
-        if not self.assume_straight_pages:
+        if not self.assume_straight_pages and not self.assume_straight_text:
             for class_name in dict_loc_preds.keys():
                 crops[class_name], dict_loc_preds[class_name], word_orientations = self._rectify_crops(
                     crops[class_name], dict_loc_preds[class_name]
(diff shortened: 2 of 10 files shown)


felixdittrich92 (changes requested):
Hi @milosacimovic :wave: 
Thanks a lot for the quick PR :+1: 

A problem i see is the increasing complexity with `ocr_predictor` / `kie_predictor` .
From my experience adding another `assume_` kwarg would users more and more confuse.
Additional it makes it only possible to disable the `crop_orientation_predictor` actually.

So two suggestions from my view:

Option 1:
Advantages: 
- We can avoid that the orientation models needs to be initialized
- Clear about what it does

Disadvantages:
- Needs also modifications in `demo` and `api` 
- 2 additional `ocr_predictor` / `kie_predictor` args (could also be passed as kwargs maybe !?)

In this case `disable_page_orientation` has only an effect in combination with `assume_straight_pages=False` and or `detect_orientation=True` and or `straighten_pages=True` where it then can handle only small rotations in the range between ~ -45 and 45 degrees
And `disable_crop_orientation`would have only an effect with `assume_straight_pages=False` so that it everytime results in a "prediction" of 0 and 1.0 as probability (or None)


```
model = ocr_predictor(
    pretrained=True,
    assume_straight_pages=False,
    straighten_pages=True,
    detect_orientation=True,
    disable_page_orientation=True, # maybe as kwarg ? Then can handle only small rotations
    disable_crop_orientation=True, # maybe as kwarg ? Then returns always 0 and as prob 1.0 or None (prefered None ?)
)
```

Option 2:
Advantages:
- Encapsulated from `ocr_predictor` and cleaner handling by specific `predictor`

Disadvantages:
- Needs also modifications in `demo` and `api`
- Possible no way that the orientation models are loaded once into RAM before removal

```
predictor = ocr_predictor(
    pretrained=True,
    assume_straight_pages=False,
    straighten_pages=True,
    detect_orientation=True,
)

# Overwrite the orientation models - disable
predictor.crop_orientation_predictor = crop_orientation_predictor(disabled=True)
predictor.page_orientation_predictor = page_orientation_predictor(disabled=True)
```

pseudo code:
https://github.com/mindee/doctr/blob/9045dcfc9c5c837b06fcda8e802f7cf1d95bd18c/doctr/models/classification/zoo.py#L37
```
if kwargs.get("disabled", False):
     return OrientationPredictor(None, None)
```

https://github.com/mindee/doctr/blob/9045dcfc9c5c837b06fcda8e802f7cf1d95bd18c/doctr/models/classification/predictor/pytorch.py#L18
```
class OrientationPredictor(nn.Module):

    def __init__(
        self,
        pre_processor: Optional[PreProcessor] = None,
        model: Optional[nn.Module] = None,
    ) -> None:
        super().__init__()
        self.pre_processor = pre_processor
        self.model = model.eval() if model else model

    @torch.inference_mode()
    def forward(
        self,
        inputs: List[Union[np.ndarray, torch.Tensor]],
    ) -> List[Union[List[int], List[float]]]:
        # Dimension check
        if any(input.ndim != 3 for input in inputs):
            raise ValueError("incorrect input shape: all inputs are expected to be multi-channel 2D images.")

        if model is None:
            in_length = len(inputs)
            return [[0] * in_length, [0] * in_length, [1.0] in_length]

        processed_batches = self.pre_processor(inputs)
        _params = next(self.model.parameters())
        self.model, processed_batches = set_device_and_dtype(
            self.model, processed_batches, _params.device, _params.dtype
        )
        predicted_batches = [self.model(batch) for batch in processed_batches]
        # confidence
        probs = [
            torch.max(torch.softmax(batch, dim=1), dim=1).values.cpu().detach().numpy() for batch in predicted_batches
        ]
        # Postprocess predictions
        predicted_batches = [out_batch.argmax(dim=1).cpu().detach().numpy() for out_batch in predicted_batches]

        class_idxs = [int(pred) for batch in predicted_batches for pred in batch]
        classes = [int(self.model.cfg["classes"][idx]) for idx in class_idxs]
        confs = [round(float(p), 2) for prob in probs for p in prob]

        return [class_idxs, classes, confs]
```

Only quick and dirty to "visualize" the idea i have in mind :sweat_smile: 

If we could realize option 1 and use kwargs to pass these values i think i would orefer the first idea.

Additional in every case we need some entry in the documentation for the new logic.

@milosacimovic @odulcy-mindee wdyt ? :hugs:

felixdittrich92:
So i totally agree with the feature but we should take care of both the `crop` and `page` orientation predictors and we should take care not to miss stuff:

- Needs to work for both `ocr_predictor` and `kie_predictor`
- New logic needs also to be added in `demo` and `api`
- An well formed entry in the documentation to describe and showcase the new logic. (Should be part in https://github.com/mindee/doctr/blob/main/docs/source/using_doctr/using_models.rst)

:smiley:

felixdittrich92:
> Hi @felixdittrich92 , Thank you so much for considering my PR so quickly and for the immensely valuable feedback on the API changes. I will look into the first option. However, what I would still like to get from you is your thoughts on the subtle differences between
> 
> ```
> def extract_dewarped_crops(
>     img: np.ndarray, polys: np.ndarray, dtype=np.float32, channels_last: bool = True
> ) -> List[np.ndarray]:
>     """Created cropped images from list of skewed/warped bounding boxes,
>     but containing straight text
> ```
> 
> and currently used
> 
> ```
> def extract_rcrops(
>     img: np.ndarray, polys: np.ndarray, dtype=np.float32, channels_last: bool = True
> ) -> List[np.ndarray]:
>     """Created cropped images from list of rotated bounding boxes
> ```
> 
> From my experience the `extract_rcrops` has issues when extracting crops from slightly rotated documents (-45, 45) where it rotates the crops even though it should keep them straight.
> 
> This was actually my main complaint about the current implementation.

I will take a look into asap :+1: But all the stuff points to the same issue so we can combine both in your PR :+1:

felixdittrich92:
@milosacimovic Tested your function and yeah it works better for smaller rotated pages (between -45 and 45).
It's also a bit slower but not as much.
I think that's something to combine:

With `disable_page_orientation=True` (where we expect only small rotated pages) + your function
Otherwise: the current function

Wdyt ?

felixdittrich92:
Hi @milosacimovic :wave:,

I quickly prototyped this feature.
Wdyt about the changes: https://github.com/mindee/doctr/compare/main...felixdittrich92:doctr:disable-orient-prototype ?

Feel free to test and update your PR with my changes if everything works as expected then only the docs part and maybe additional tests + mypy/format fixes would be open :hugs:

felixdittrich92:
#1735

PR #1731, opened 2024-09-25
[Fix] Remove image padding after rotation correction with `straighten_pages=True`

felixdittrich92:
This PR:

- Fix padding removal after rotation correction
- Add function to remove the padding + corresponding test

Before:

Screenshot from 2024-09-25 08-45-05

After:

Screenshot from 2024-09-25 09-00-45

### DIFF
--- a/doctr/models/predictor/base.py
+++ b/doctr/models/predictor/base.py
@@ -8,7 +8,7 @@ from typing import Any, Callable, Dict, List, Optional, Tuple
 import numpy as np
 
 from doctr.models.builder import DocumentBuilder
-from doctr.utils.geometry import extract_crops, extract_rcrops, rotate_image
+from doctr.utils.geometry import extract_crops, extract_rcrops, remove_image_padding, rotate_image
 
 from .._utils import estimate_orientation, rectify_crops, rectify_loc_preds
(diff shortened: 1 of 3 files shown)


doctr/models/predictor/base.py
```diff
@@ -101,8 +101,8 @@ def _straighten_pages(
             ]
         )
         return [
-            # expand if height and width are not equal
-            rotate_image(page, angle, expand=page.shape[0] != page.shape[1])
+            # expand if height and width are not equal, afterwards remove padding
```
odulcy-mindee, line 104:
```suggestion
            # expand if height and width are not equal, then remove the padding
```

odulcy-mindee (approved):

felixdittrich92:
@odulcy-mindee btw the API fails actually (after the numpy 2.0 PR merge) 

poetry install tf == 2.14 besides numpy 2.0 - which can't work :sweat_smile: 
and pip installs tf 2.15 besides numpy < 2.0 which is correct and works 

Should we pin numpy < 2.0 in the api requirements or pyproject.toml for the moment ? 
If the keras v3 stuff is done we can upgrade tf also and remove the pin

PR #1744, opened 2024-10-07
[TF] Move model building & unify train scripts

felixdittrich92:
This PR:

- Move model building unified "under the hood"
- Unify train scripts
- Fix missing kwargs (training = True / False :exploding_head:)


doctr/models/factory/hub.py
felixdittrich92, line 1: @felixdittrich92 remove pass

odulcy-mindee (approved):

Issue #1692, opened 2024-08-13
page.synthesize() returns error when assume_straight_pages=False


hienphan161:
### Bug description

I've seen that when I use `assume_straight_pages=False`, it makes `result.pages[0].synthesize()` stop working. 

### Code snippet to reproduce the bug

```
predictor = ocr_predictor(
  pretrained=True,
  assume_straight_pages=False,
  straighten_pages=True,
  detect_orientation=True,
  resolve_blocks=True
)

file_name = "/content/vecteezy_modern-professional-business-card-creative-and-simple_9010481 - Copy.jpg"
image = Image.open(file_name)
image = ImageOps.exif_transpose(image)

result = predictor([np.asarray(image)])
result.pages[0].synthesize()
```

### Error traceback

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
[<ipython-input-7-7d5fc2bce17c>](https://localhost:8080/#) in <cell line: 1>()
----> 1 result.pages[0].synthesize()

1 frames
[/usr/local/lib/python3.10/dist-packages/doctr/io/elements.py](https://localhost:8080/#) in synthesize(self, **kwargs)
    315             synthesized page
    316         """
--> 317         return synthesize_page(self.export(), **kwargs)
    318 
    319     def export_as_xml(self, file_title: str = "docTR - XML export (hOCR)") -> Tuple[bytes, ET.ElementTree]:

[/usr/local/lib/python3.10/dist-packages/doctr/utils/reconstitution.py](https://localhost:8080/#) in synthesize_page(page, draw_proba, font_family)
     41             for word in line["words"]:
     42                 # Get absolute word geometry
---> 43                 (xmin, ymin), (xmax, ymax) = word["geometry"]
     44                 xmin, xmax = int(round(w * xmin)), int(round(w * xmax))
     45                 ymin, ymax = int(round(h * ymin)), int(round(h * ymax))

ValueError: too many values to unpack (expected 2)
```

### Environment

DocTR version: v0.9.0
TensorFlow version: 2.17.0
PyTorch version: 2.3.1+cu121 (torchvision 0.18.1+cu121)
OpenCV version: 4.10.0
OS: Ubuntu 22.04.3 LTS
Python version: 3.10.12
Is CUDA available (TensorFlow): Yes
Is CUDA available (PyTorch): Yes
CUDA runtime version: 12.2.140
GPU models and configuration: GPU 0: Tesla T4
Nvidia driver version: 535.104.05
[...]

PR #1750, opened 2024-10-10
[Reconstitution] Improve reconstitution

felixdittrich92:
This PR:

- Improve synthesize quality
- Allow also to render polygons but warn because to large rotations can't be rendered correctly yet
- Extend corresponding tests

Still not perfect but looks much better as before :)

Old:  (Lots of overlapping / clipping)
Screenshot from 2024-10-10 10-40-13

New:
Screenshot from 2024-10-10 11-21-24

New without line:
Screenshot from 2024-10-10 10-50-56

### DIFF
--- a/doctr/utils/reconstitution.py
+++ b/doctr/utils/reconstitution.py
@@ -67,10 +67,6 @@ def _synthesize(
     except ValueError:
         font = get_font(font_family, min_font_size)
 
-    # Position the text left side of the bounding box
-    x_offset = 0
-    y_offset = 0
-
     # Create a mask for the word
     mask = Image.new("L", (w, h), 0)
     ImageDraw.Draw(mask).polygon([(int(round(w * x)), int(round(h * y))) for x, y in polygon], fill=255)
@@ -78,9 +74,9 @@ def _synthesize(
     # Draw the word text
     d = ImageDraw.Draw(response)
     try:
-        d.text((xmin + x_offset, ymin + y_offset), word_text, font=font, fill=(0, 0, 0), anchor="lt")
+        d.text((xmin, ymin), word_text, font=font, fill=(0, 0, 0), anchor="lt")
     except UnicodeEncodeError:
-        d.text((xmin + x_offset, ymin + y_offset), anyascii(word_text), font=font, fill=(0, 0, 0), anchor="lt")
+        d.text((xmin, ymin), anyascii(word_text), font=font, fill=(0, 0, 0), anchor="lt")
 
     if draw_proba:
         confidence = (
--- a/doctr/utils/reconstitution.py
+++ b/doctr/utils/reconstitution.py
@@ -14,7 +14,7 @@ from .fonts import get_font
 __all__ = ["synthesize_page", "synthesize_kie_page"]
 
 
-def _warn_rotation(entry: Dict[str, Any], already_warned: bool) -> bool:
+def _warn_rotation(entry: Dict[str, Any], already_warned: bool) -> bool:  # pragma: no cover
     if len(entry["geometry"]) == 4 and not already_warned:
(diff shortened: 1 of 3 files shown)


doctr/utils/reconstitution.py
```diff
@@ -13,61 +14,156 @@
 __all__ = ["synthesize_page", "synthesize_kie_page"]
 
 
+def _warn_rotation(entry: Dict[str, Any], already_warned: bool) -> bool:  # pragma: no cover
+    if len(entry["geometry"]) == 4 and not already_warned:
+        logging.warning("Polygons with larger rotations will lead to inaccurate rendering")
+        return True
+    return already_warned
```
odulcy-mindee, line 21: `already_warned` should be a global variable, not a parameter. It's not the responsability of other callers to check if warning has been already raised

Something like this:
```suggestion
ROTATION_WARNING = False

def _warn_rotation(entry: Dict[str, Any]) -> None:
    if not ROTATION_WARNING and len(entry["geometry"]) == 4:
        logging.warning("Polygons with larger rotations will lead to inaccurate rendering")
        ROTATION_WARNING = True
```

doctr/utils/reconstitution.py
```diff
@@ -13,61 +14,156 @@
...
+
+    # Create a mask for the word
+    mask = Image.new("L", (w, h), 0)
+    ImageDraw.Draw(mask).polygon([(int(round(w * x)), int(round(h * y))) for x, y in polygon], fill=255)
+
+    # Draw the word text
+    d = ImageDraw.Draw(response)
+    try:
+        d.text((xmin, ymin), word_text, font=font, fill=(0, 0, 0), anchor="lt")
+    except UnicodeEncodeError:
```
odulcy-mindee, line 78: It's strange that you're always able to compute the font_size for this word_text, but it might still be possible to fail when writing it

felixdittrich92:
Closes: #1692

doctr/utils/reconstitution.py
felixdittrich92: Yeah good catch will be applied tomorrow :)

tests/common/test_utils_reconstitution.py
```diff
@@ -23,12 +23,6 @@ def test_synthesize_page():
     assert isinstance(render_one_line, np.ndarray)
     assert render_one_line.shape == (*pages[0].dimensions, 3)
 
-    # test with a smiley which can't be rendered by unidecode
-    pages_one_line["blocks"][0]["lines"][0]["words"][0]["text"] = "🤯"
-    render_one_line = reconstitution.synthesize_page(pages_one_line, draw_proba=True)
-    assert isinstance(render_one_line, np.ndarray)
-    assert render_one_line.shape == (*pages[0].dimensions, 3)
-
```
odulcy-mindee: Haha, it didn't work ?

felixdittrich92: It worked but i was not able to reach the line where it breaks d.draw because d.draw was also able to draw it xDDD

felixdittrich92: So the idea was to cover the unicode error also where it falls back to anyascii xD

felixdittrich92: but i would still keep it as fallback only for the case :sweat_smile:

felixdittrich92:
@odulcy-mindee Any further changes in mind or fine ? :)

PS: Please check your Slack 😅

felixdittrich92:
@odulcy-mindee ?

tests/common/test_utils_reconstitution.py
odulcy-mindee: Ok, got it, thanks

odulcy-mindee (approved):

PR #1758, opened 2024-10-24
Typo fix in train_pytorch.py

agarkovv:
"targe" -> "target"


felixdittrich92 (approved):
Thanks @agarkovv :hugs:

PR #1762, opened 2024-10-25
[tests/style] Fix tf test and formatting

felixdittrich92:
This PR:

- Fix failing TF test
- Formatting

### DIFF
--- a/doctr/transforms/functional/tensorflow.py
+++ b/doctr/transforms/functional/tensorflow.py
@@ -140,7 +140,7 @@ def rotate_sample(
     rotated_geoms[..., 0] = rotated_geoms[..., 0] / rotated_img.shape[1]
     rotated_geoms[..., 1] = rotated_geoms[..., 1] / rotated_img.shape[0]
 
-    return rotated_img, np.clip(rotated_geoms, 0, 1)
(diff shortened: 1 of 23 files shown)


doctr/transforms/functional/tensorflow.py
```diff
@@ -140,7 +140,7 @@ def rotate_sample(
     rotated_geoms[..., 0] = rotated_geoms[..., 0] / rotated_img.shape[1]
     rotated_geoms[..., 1] = rotated_geoms[..., 1] / rotated_img.shape[0]
 
-    return rotated_img, np.clip(rotated_geoms, 0, 1)
+    return rotated_img, np.clip(np.around(rotated_geoms, decimals=15), 0, 1)
```
felixdittrich92, line 143: @odulcy-mindee I added this already to the pytorch implementation (https://github.com/mindee/doctr/blob/e52538017dd80418c9d3f19decd6143b1a9d114e/doctr/transforms/functional/pytorch.py#L92) but missed to add it to TF side :sweat_smile: 

Now with TF 2.18 it installs with numpy 2.0 (before <2.0 with 2.0 compatibility) and that's the reason why the test was failing

odulcy-mindee (approved):

Issue #1743, opened 2024-10-07
[Docker] Update Dockerfile to CUDA 12.X


felixdittrich92:
### 🚀 The feature

Dockerfile is still pinned to CUDA 11.8 now with the upgraded versions we should switch also to CUDA 12.X

PR #1769, opened 2024-11-05
bump: :arrow_up: CUDA Version

odulcy-mindee:
Fix https://github.com/mindee/doctr/issues/1743

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,7 +98,8 @@ docs = [
 dev = [
     # Tensorflow
     # cf. https://github.com/mindee/doctr/pull/1461
-    "tensorflow[and-cuda]>=2.15.0,<3.0.0",
+    "tensorflow[and-cuda]>=2.15.0,<3.0.0; sys_platform != 'darwin'",
+    "tensorflow>=2.15.0,<3.0.0; sys_platform == 'darwin'", # TensorFlow compatible for macOS
     "tf-keras>=2.15.0,<3.0.0",  # Keep keras 2 compatibility
     "tf2onnx>=1.16.0,<2.0.0",  # cf. https://github.com/onnx/tensorflow-onnx/releases/tag/v1.16.0
     # PyTorch


Dockerfile
```diff
@@ -16,22 +16,22 @@ RUN if [ "$SYSTEM" = "gpu" ]; then \
...
+    libcufft-12-3 \
+    libcurand-12-3 \
+    libcusolver-12-3 \
+    libcusparse-12-3 \
+    libcublas-12-3 \
     # - CuDNN: https://docs.nvidia.com/deeplearning/sdk/cudnn-install/index.html#ubuntu-network-installation
-    libcudnn8=8.6.0.163-1+cuda11.8 \
-    libnvinfer-plugin8=8.6.1.6-1+cuda11.8 \
-    libnvinfer8=8.6.1.6-1+cuda11.8; \
+    libcudnn8=8.9.6.50-1+cuda12.2 \
```
felixdittrich92, line 32: I think this should be `libcudnn9` !? 
I remember this was the main issue i had also with OnnxTR running with cuda >= 12
ref.: https://hub.docker.com/layers/nvidia/cuda/12.3.2-cudnn9-runtime-ubuntu20.04/images/sha256-5bb3eeae860aa89abd6f133c68eb15eea08d784a8655aa79d55aa450b2961d63?context=explore

Would it be maybe easier to use a already working base image ?

I did something similar for our pipelines

```
ARG IMAGE=nvidia/cuda:12.6.2-base-ubuntu22.04

FROM ${IMAGE}
ARG SYSTEM=gpu
ARG PORT=80
ARG GITHUB_TOKEN

WORKDIR /app
ENV PATH="/root/miniconda3/bin:${PATH}"
ARG PATH="/root/miniconda3/bin:${PATH}"
ENV STORAGE_PATH ${STORAGE_PATH}
ENV PORT ${PORT}
ENV GITHUB_TOKEN ${GITHUB_TOKEN}

# if STORAGE_PATH is not empty set it as VOLUME otherwise set "/simplification-storage"
VOLUME ${STORAGE_PATH:-/simplification-storage}


COPY pyproject.toml /app/pyproject.toml
COPY app /app/app

RUN apt-get -y update
RUN apt-get -y install git
RUN apt-get install -y wget && rm -rf /var/lib/apt/lists/*

RUN wget \
    https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh \
    && mkdir /root/.conda \
    && bash Miniconda3-latest-Linux-x86_64.sh -b \
    && rm -f Miniconda3-latest-Linux-x86_64.sh

RUN if [ "$SYSTEM" = "gpu" ]; then \
    conda install pytorch torchvision torchaudio pytorch-cuda=11.8 -c pytorch -c nvidia ; \
    else \
    conda install pytorch torchvision torchaudio cpuonly -c pytorch ; \
    fi

RUN pip install . --extra-index-url=https://XYZ

# Echo some information
RUN echo ${PORT} \
    && echo ${SYSTEM} \
    && echo ${STORAGE_PATH}

EXPOSE ${PORT}

CMD uvicorn app.main:app --host 0.0.0.0 --port ${PORT}
```

In my case default is used for the kubernetes cluster and the following for the GH CI testing stuff

```
docker compose build --build-arg "IMAGE=python:3.10-slim" --build-arg "SYSTEM=cpu" --build-arg "PORT=8000"
```

odulcy-mindee: @felixdittrich92 You install `pytorch-cuda=11.8` in Dockerfile, I think it'll mess up the nvidia driver in the Docker container

odulcy-mindee: Yeah, indeed, it would be easier but I think these images are bloated. Do you know where I can find its Dockerfile, I can't find it.
With the updated version, PyTorch works but not TensorFlow :facepalm:
Moreover, both frameworks provides nvidia drivers now, so I think we need to revamp a bit the Dockerfile and the `pyproject.toml`.
I'll continue this PR, I'll not have too much time today to work on it

felixT2K: @odulcy-mindee Not sure i think it's this: https://gitlab.com/nvidia/container-images/cuda/-/blob/master/templates/ubuntu/base-dockerfile.j2

felixT2K: Ähhmm.. i think with tensorflow it changed a bit 

pip3 install tensorflow  does not anymore provide cuda support by default it should be `tensorflow[and-cuda]` now, if the problem is that it does not take the GPU

felixT2K: Yeah no stress --slow progress is better as no progress 🤣

felixdittrich92: And this one looks like the latest: https://gitlab.com/nvidia/container-images/cuda/-/blob/master/dist/12.6.1/ubuntu2204/base/Dockerfile?ref_type=heads

odulcy-mindee: Ah, thank you, the latter seems simpler. I wanted to have a lighter docker image for CPU. I guess simpler version would be to use `nvidia/cuda:12.4.0-base-ubuntu22.04`.

felixT2K:
```
- name: Build Docker image
        id: build
        uses: docker/build-push-action@f2a1d5e99d037542a71f64918e516c093c6f3fc4
        with:
          context: .
          build-args: |
            IMAGE=${{ matrix.base_image }}
            FRAMEWORK=${{ matrix.framework }}
            PYTHON_VERSION=${{ matrix.python }}
            SYSTEM=${{ matrix.system }}
            DOCTR_REPO=${{ github.repository }}
            DOCTR_VERSION=${{ github.sha }}
```

felixT2K: Ah i see this would conflict with the python version arg

felixT2K: And for the tensorflow issue i think the only option would be to add an additional specifier:

```
tf = [
    # cf. https://github.com/mindee/doctr/pull/1461
    "tensorflow>=2.15.0,<3.0.0",
    "tf-keras>=2.15.0,<3.0.0",  # Keep keras 2 compatibility
    "tf2onnx>=1.16.0,<2.0.0",  # cf. https://github.com/onnx/tensorflow-onnx/releases/tag/v1.16.0
]
tf-gpu = [
    # cf. https://github.com/mindee/doctr/pull/1461
    "tensorflow[and-cuda]>=2.15.0,<3.0.0",
    "tf-keras>=2.15.0,<3.0.0",  # Keep keras 2 compatibility
    "tf2onnx>=1.16.0,<2.0.0",  # cf. https://github.com/onnx/tensorflow-onnx/releases/tag/v1.16.0
]
```

Which would also require updates in the readme / docs / (CI / API ?) 😅

odulcy-mindee: > the only option would be to add an additional specifier
Yeah, I guess so

felixdittrich92: Pseudo: 

Something like the following could work:

```
jobs:
  build:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        base_image: ["nvidia/cuda:12.4.0-base-ubuntu22.04", "python:py-version-slim"]
        python: ["3.10.13", "3.11.8", "3.12.7"]
    
    steps:
      - name: Set IMAGE
        id: image
        run: |
          if [[ "${{ matrix.base_image }}" == *"py-version"* ]]; then
            echo "IMAGE=${{ matrix.base_image/py-version/${{ matrix.python }} }}" >> $GITHUB_ENV
          else
            echo "IMAGE=${{ matrix.base_image }}" >> $GITHUB_ENV

      - name: Build Docker image
        id: build
        uses: docker/build-push-action@f2a1d5e99d037542a71f64918e516c093c6f3fc4
        with:
          context: .
          build-args: |
            IMAGE=${{ env.IMAGE }}
            FRAMEWORK=${{ matrix.framework }}
            PYTHON_VERSION=${{ matrix.python }}   # Only required for the nvidia base image
            SYSTEM=${{ matrix.system }}  # I think can be dropped in this case
            DOCTR_REPO=${{ github.repository }}
            DOCTR_VERSION=${{ github.sha }}
 ```

**Comment by odulcy-mindee** [comment]
@felixdittrich92 I checked disk usage for both images, I don't think it's worthwhile to make a distinction for base image:
```
ubuntu                             22.04                                      97271d29cb79   2 months ago    77.9MB
nvidia/cuda                        12.4.0-base-ubuntu22.04                    adcf2ab568f0   7 months ago    244MB
```

So I only kept `nvidia/cuda:12.2.0-base-ubuntu22.04` as base image.
Moreover, with the current docTR setup, torch always install nvidia packages, so I decided to do the same with TensorFlow, which leads to simplification in the CI.
I also set CUDA version to `12.2`. It's an arbitrary choice. It should be compatible enough for most setup.

Tell me what you think about this setup, especially the suggested changes I made in `pyproject.toml`

**Comment by felixdittrich92** [comment]
> @felixdittrich92 I checked disk usage for both images, I don't think it's worthwhile to make a distinction for base image:
> 
> ```
> ubuntu                             22.04                                      97271d29cb79   2 months ago    77.9MB
> nvidia/cuda                        12.4.0-base-ubuntu22.04                    adcf2ab568f0   7 months ago    244MB
> ```
> 
> So I only kept `nvidia/cuda:12.2.0-base-ubuntu22.04` as base image. Moreover, with the current docTR setup, torch always install nvidia packages, so I decided to do the same with TensorFlow, which leads to simplification in the CI. I also set CUDA version to `12.2`. It's an arbitrary choice. It should be compatible enough for most setup.
> 
> Tell me what you think about this setup, especially the suggested changes I made in `pyproject.toml`

The `pyproject.toml` change is fine for me it restores the old behaviour before tf 2.18 and at the end people can still install the prefered tf version before installing doctr :+1: 

Yeah every CUDA 12.X should be fine :)

I'm also fine with the base image - (less CI jobs ^^) because it doesn't matter much (torch and tf are already really huge packages) - people can still build there own containers - or use OnnxTR if they want something more lightweight/hardware optimized

File: README.md [hunk 2]
```diff
@@ -254,37 +254,37 @@ Check out our [TensorFlow.js demo](https://github.com/mindee/doctr-tfjs-demo) to
...
 ### Docker container
 
-[We offer Docker container support for easy testing and deployment](https://github.com/mindee/doctr/pkgs/container/doctr).
+We offer Docker container support for easy testing and deployment. [Here are the available docker tags.](https://github.com/mindee/doctr/pkgs/container/doctr).
 
 #### Using GPU with docTR Docker Images
 
-The docTR Docker images are GPU-ready and based on CUDA `11.8`.
-However, to use GPU support with these Docker images, please ensure that Docker is configured to use your GPU.
+The docTR Docker images are GPU-ready and based on CUDA `12.2`. Make sure your host is **at least `12.2`**, otherwise Torch or TensorFlow won't be able to initialize the GPU.
```
felixdittrich92, line 261: I think we should also mention that the `nvidia-container-toolkit`
on the host system is required (https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html)
otherwise also with correct setup driver / CUDA / cudnn  it wouldn't detect your host GPU from the container correctly

felixdittrich92: Ah i was to fast xD is solved ^^

pyproject.toml
```diff
@@ -53,7 +53,7 @@ dependencies = [
 [project.optional-dependencies]
 tf = [
     # cf. https://github.com/mindee/doctr/pull/1461
-    "tensorflow>=2.15.0,<3.0.0",
+    "tensorflow[and-cuda]>=2.15.0,<3.0.0",
```
felixdittrich92, line 56: Mh it looks like that's conflicting with MacOS unfortunately :/  --> failing CI job

felixdittrich92: We could try to control it with sys_platform condition -> https://peps.python.org/pep-0631/#example

felixdittrich92: something like:
```
"tensorflow[and-cuda]>=2.15.0,<3.0.0; platform_system != 'Darwin' or tensorflow>=2.15.0,<3.0.0; platform_system == 'Darwin'"
```
Or (not sure that the `or` condition works correctly

```
"tensorflow[and-cuda]>=2.15.0,<3.0.0; platform_system != 'Darwin'",
# TensorFlow compatible for macOS
"tensorflow>=2.15.0,<3.0.0; platform_system == 'Darwin'",
```

felixT2K: works ^^ 👍🏼

pyproject.toml
```diff
@@ -97,7 +98,7 @@ docs = [
 dev = [
     # Tensorflow
     # cf. https://github.com/mindee/doctr/pull/1461
-    "tensorflow>=2.15.0,<3.0.0",
+    "tensorflow[and-cuda]>=2.15.0,<3.0.0",
```
felixT2K, line 101: @odulcy-mindee we should do the same here as above :)

odulcy-mindee: Oh yeah :+1:

felixdittrich92 (approved):
@odulcy-mindee 

Thanks LGTM :+1: 

But one thing the mypy CI job fails now due to:

```
ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device
```
:sweat_smile:

felixdittrich92:
So the problem now with the updated dev dependencies is that it needs to much space for the Github standard runners:

Option 1: Increase the runners to some larger ones (`ubuntu-latest-large` for example everywhere in the CI where we need to install .[dev] - I think this would need a configuration if you don't use enterprise on the mindee github account)
Option 2: ?? :sweat_smile:  (some hacky quick fixes would be possible - but personally i don't like one of them) :sweat_smile:

felixdittrich92:
Ok after cleaning the action caches it works again maybe we should add a CI job to auto clean the caches:

https://docs.github.com/de/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#managing-caches

Failing mypy can be ignored the issues are already solved on main

felixdittrich92:
For cache cleaning we could add a CI job (other PR)

For example:
172800

```
name: Clear Cache

on:
  workflow_dispatch:

on:
  schedule:
    - cron: '0 0 * * *'  # Runs once a day
  
jobs:
  clear:
    name: Clear cache
    runs-on: ubuntu-latest
    steps:
    - uses: MyAlbum/purge-cache@v1
      with:
          max-age: 172800 # Caches older then 2 days are delete
```

odulcy-mindee:
@felixdittrich92 Ok, seems a good idea to add a CI Job in another PR

felixdittrich92:
@odulcy-mindee Looks like anything with the naming tag is going wrong xD See last:
https://github.com/mindee/doctr/pkgs/container/doctr

odulcy-mindee:
@felixdittrich92 oh, thank you for pointing this issue. I'll have a look at it today

PR #1770, opened 2024-11-06
[build] Upgrade py 3.10

felixdittrich92:
This PR:

- Move to minimum py 3.10 for next release

- reason: 
   - numpy 2.0 requires py >= 3.10 
   - same for TF >=2.18 because it depends on numpy >= 2.0 
   - same for onnxruntime >= 1.19 which dropped support for py 3.9 because of dep  numpy >= 2.0

- Fix mypy
- minor numpy 2.0 compatibility fix int0 -> intp (does also exist in <2.0)

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python: ["3.9", "3.10"]
+        python: ["3.10", "3.11"]
         framework: [tensorflow, pytorch]
     steps:
       - uses: actions/checkout@v4
@@ -24,7 +24,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           # MacOS issue ref.: https://github.com/actions/setup-python/issues/855 & https://github.com/actions/setup-python/issues/865
-          python-version: ${{ matrix.os == 'macos-latest' && '3.11' || matrix.python }}
+          python-version: ${{ matrix.os == 'macos-latest' && matrix.python == '3.10' && '3.11' || matrix.python }}
           architecture: x64
       - if: matrix.framework == 'tensorflow'
         name: Cache python modules (TF)
@@ -58,7 +58,7 @@ jobs:
       - uses: conda-incubator/setup-miniconda@v3
         with:
           auto-update-conda: true
-          python-version: 3.9
+          python-version: "3.10"
           channels: pypdfium2-team,bblanchon,defaults,conda-forge
           channel-priority: strict
       - name: Install dependencies
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ maintainers = [
     {name = "Felix Dittrich"},
 ]
 readme = "README.md"
-requires-python = ">=3.9.0,<4"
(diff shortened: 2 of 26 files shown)


felixdittrich92:
@odulcy-mindee py3.9 end of life is Oct 25 but if already so many move to 3.10 i think we should follow that people will benefit from the improvements - Additional this would make it possible to move to python's built in types and away from the >= py3.11 deprecated `typing` types

.github/workflows/builds.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           # MacOS issue ref.: https://github.com/actions/setup-python/issues/855 & https://github.com/actions/setup-python/issues/865
-          python-version: ${{ matrix.os == 'macos-latest' && '3.11' || matrix.python }}
+          python-version: ${{ matrix.os == 'macos-latest' && matrix.python == '3.10' && '3.11' || matrix.python }}
```
odulcy-mindee, line 27: Why is there `3.11` written in this condition ?

odulcy-mindee (approved):

.github/workflows/builds.yml
felixdittrich92: It's still the action bug with macOS for the python matrix == 3.10 uses python 3.11 to avoid the bug do you remember ? :)

felixdittrich92:
@odulcy-mindee you need to merge this and adjust the required test rules in the GitHub repo settings :)

.github/workflows/builds.yml
odulcy-mindee: Yeah, but why `matrix.python == '3.10' && '3.11' || matrix.python` ? Shouldn't `matrix.os == 'macos-latest' && matrix.python == '3.10'` be enough ?

felixdittrich92: @odulcy-mindee syntax is a bit confusing the last && defines which value to use so

matrix.os == 'macos-latest' && matrix.python == '3.10' -> if both True use 3.11 (&& '3.11')  else use the matrix.python version

We could also used the version before so if macos-latest use always python 3.11 the change is a bit more explicit (but can be fully removed later on if we change to py 3.11 & py 3.12)

odulcy-mindee: Ah ok, thanks for the clarification

odulcy-mindee:
> you have to merge this (i can't ^^) and adjust the required test rules in the GitHub repo settings :)

@felixdittrich92 done

felixdittrich92:
Thanks 😊

pyproject.toml
```diff
@@ -183,7 +183,7 @@ ignore_missing_imports = true
 [tool.ruff]
 exclude = [".git", "venv*", "build", "**/__init__.py"]
```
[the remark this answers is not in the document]
felixdittrich92: Hey @kuraga :)

Yeah because it does strange things here:

__init__.py example:

```
from doctr.file_utils import is_tf_available, is_torch_available

if is_tf_available():
    from .tensorflow import *
elif is_torch_available():
    from .pytorch import *  # type: ignore[assignment]
```

and if i remember it has raised circular import issues on other places :sweat_smile:

PR #1779, opened 2024-11-16
[misc] Change prefered backend from tf to torch

felixdittrich92:
This PR:

- Change docTR's "prefered" backend from TF to PyTorch if both frameworks are installed
- Change the API template backend to PyTorch
- Documentation is build from code base with PyTorch as prefered backend now
- Some documentation fixes


api/Makefile
```diff
@@ -3,7 +3,7 @@
 .PHONY: lock run stop test
 # Pin the dependencies
 lock:
-	pip install poetry>=1.0
+	pip install poetry>=1.0 poetry-plugin-export
```
felixdittrich92, line 6: @odulcy-mindee That's required because `poetry-plugin-export` will be dropped from future poetry releases

api/app/utils.py
```diff
@@ -24,11 +24,9 @@ async def get_documents(files: List[UploadFile]) -> Tuple[List[np.ndarray], List
     """Convert a list of UploadFile objects to lists of numpy arrays and their corresponding filenames
 
     Args:
-    ----
```
felixdittrich92, line 27: This was not rendered correctly in our sphinx docs so i removed it

doctr/contrib/__init__.py
```diff
@@ -0,0 +1 @@
+from .artefacts import ArtefactDetector
```
felixdittrich92, line 1: without we can't build the docs from the module

odulcy-mindee (approved):

PR #1786, opened 2024-11-19
[CI] Fix PR labeler job & Rollback doc deploy deps

felixdittrich92:
This PR:

- Fix PR labeler
- Rollback docs deploy deps

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -25,11 +25,10 @@ jobs:
           path: ~/.cache/pip
           key: ${{ runner.os }}-pkg-deps-${{ matrix.python }}-${{ hashFiles('pyproject.toml') }}-docs
       - name: Install dependencies
-        # Earlier documentation (before v10) requires TensorFlow - later it's built from the PyTorch code - so we need to install both (dev)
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -25,11 +25,10 @@ jobs:
           path: ~/.cache/pip
           key: ${{ runner.os }}-pkg-deps-${{ matrix.python }}-${{ hashFiles('pyproject.toml') }}-docs
       - name: Install dependencies
-        # Earlier documentation (before v10) requires TensorFlow - later it's built from the PyTorch code - so we need to install both (dev)
-        # Additional the dev specifier includes all [docs] dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -e .[dev]
+          pip install -e .[tf,viz,html] --upgrade
+          pip install -e .[docs]
```
odulcy-mindee, line 31: No more PyTorch ?

felixdittrich92: Normally yes but it can be that this is required after the next succesful docs deployment - because it works with tf - it's a bit strange

PR #1800, opened 2024-11-29
docs: fix faulty code for prediction and recognition demos

khanfarhan10:
fix faulty code for prediction and recognition demos

### DIFF
--- a/docs/source/using_doctr/using_models.rst
+++ b/docs/source/using_doctr/using_models.rst
@@ -92,12 +92,12 @@ Detection predictors
 
     import numpy as np
     from doctr.models import detection_predictor
-    predictor = detection_predictor('db_resnet50')
+    model = detection_predictor('db_resnet50')
     dummy_img = (255 * np.random.rand(800, 600, 3)).astype(np.uint8)
     out = model([dummy_img])
 
 You can pass specific boolean arguments to the predictor:
-
+* `pretrained`: if you want to use a model that has been pretrained on a specific dataset, this will load the corresponding weights to enhance the model's performance.
 * `assume_straight_pages`: if you work with straight documents only, it will fit straight bounding boxes to the text areas.
 * `preserve_aspect_ratio`: if you want to preserve the aspect ratio of your documents while resizing before sending them to the model.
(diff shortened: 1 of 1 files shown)


khanfarhan10:
Looks like a minor bug @felixdittrich92 , kindly merge if looks okay to you!

Amazing OCR guys, Open Source all the way!

docs/source/using_doctr/using_models.rst
```diff
@@ -92,12 +92,12 @@ Detection predictors
...
     import numpy as np
     from doctr.models import detection_predictor
-    predictor = detection_predictor('db_resnet50')
+    model = detection_predictor('db_resnet50')
     dummy_img = (255 * np.random.rand(800, 600, 3)).astype(np.uint8)
     out = model([dummy_img])
 
 You can pass specific boolean arguments to the predictor:
-
+* `pretrained`: if you want to use a model that has been pretrained on a specific dataset, this will load the corresponding weights to enhance the model's performance.
```
felixdittrich92, line 100: @khanfarhan10 I think we should drop `to enhance the model's performance ` sounds a bit missleading because otherwise it's random init and will lead to no or useless results ^^ wdyt ?

felixdittrich92 (changes requested):
Hi @khanfarhan10 👋🏼,

Thanks for the PR :+1: 
Only a small comment to clearify before merging :)

felixdittrich92:
Failing api test can be ignored i missed to replace the deprecated `app=..`  in the conftest 😅 will be fixed next week so can be ignored atm

docs/source/using_doctr/using_models.rst
khanfarhan10: Modified to the following:

`pretrained`: if you want to use a model that has been pretrained on a specific dataset, setting `pretrained=True` this will load the corresponding weights. If `pretrained=False`, which is the default, would otherwise lead to a random initialization and would lead to no/useless results.

khanfarhan10: @felixdittrich92 kindly have a re-look.

felixdittrich92 (approved):
Thanks @khanfarhan10 🤗

PR #1807, opened 2024-12-03
[build] Fix tensorflow build dep

felixdittrich92:
This PR:

- the extra "[and-cuda]" is only available on linux
- for windows only WSL2 (linux) supports direct GPU support - where our logic would match - for native windows the extra does not exist same for MacOS

ref.: https://www.tensorflow.org/install/pip#windows-native


odulcy-mindee (approved):
Thanks

PR #1837, opened 2025-01-13
feat: :sparkles: tqdm slack

odulcy-mindee:
Usage of `tqdm slack` is explained [here](https://tqdm.github.io/docs/contrib.slack/#tqdm_slack-objects). You only need to set environment variables, no need to pass token and channel in args.

### DIFF
--- a/references/detection/train_tensorflow.py
+++ b/references/detection/train_tensorflow.py
@@ -14,13 +14,18 @@ os.environ["TF_CPP_MIN_LOG_LEVEL"] = "2"
 
 import datetime
 import hashlib
+import os
 import time
 from pathlib import Path
 
 import numpy as np
 import tensorflow as tf
 from tensorflow.keras import Model, mixed_precision, optimizers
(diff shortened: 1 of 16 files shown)


references/detection/train_tensorflow.py
```diff
@@ -14,13 +14,18 @@
 
 import datetime
 import hashlib
+import os
```
felixdittrich92, line 17: Already imported line 6 :)

felixdittrich92 (changes requested):
Thanks @odulcy-mindee 🤗 
Could we add a quick note to each README in the references how to setup (maybe slack reference link) and that both values needs to be set as env ?

references/detection/train_tensorflow.py
odulcy-mindee: Good catch !

felixdittrich92 (approved):
LGTM thanks :)

PR #1844, opened 2025-01-16
feat: :sparkles: ClearML training loss logging

odulcy-mindee:


references/detection/train_pytorch.py
```diff
@@ -377,32 +404,26 @@ def main(args):
...
+
     # Create loss queue
     min_loss = np.inf
     if args.early_stop:
         early_stopper = EarlyStopper(patience=args.early_stop_epochs, min_delta=args.early_stop_delta)
 
     # Training loop
     for epoch in range(args.epochs):
-        fit_one_epoch(model, train_loader, batch_transforms, optimizer, scheduler, amp=args.amp)
+        fit_one_epoch(model, train_loader, batch_transforms, optimizer, scheduler, amp=args.amp, clearml=args.clearml)
```
felixdittrich92, line 426: clearml_log instead :)

felixdittrich92 (changes requested):

references/detection/train_pytorch.py
odulcy-mindee: Good catch, thanks

felixdittrich92 (approved):
Thanks :)

kuraga:
Sorry, weren't there plans to make logging more universal?
(`clearml_log` argument name doesn't have a place for other loggers...)

felixdittrich92:
Hi @kuraga 👋🏼,

What do you mean with other loggers ? 
Sry I'm a bit out of scope with the Clearml topic (still using W&B on my own 😅)
CC @odulcy-mindee

felixdittrich92:
> @felixdittrich92 Hm... Is `W&B` exactly the answer? 😄
> 
> I like a lightweight [lazyscribe](https://github.com/lazyscribe/lazyscribe) BTW...
> 
> I don't ask "to implement other loggers" but about some abstract layer...

Ah got it I was thinking the question is related to something with ClearML :sweat_smile: For me personally W&B is still the answer xD But yeah everyone what they prefer

Mh yes I understood the request ..the thing is that the training scripts / references folder is more a "collection" of scripts everyone can modify to there own needs which has some advantages but on the other side also disadvantages (for example adding globally avaialble features shared between the different training tasks like logging, early stopping, and so on)

with an `Trainer` object like you know from the `transformers` lib for example it would be much easier to implement such features but on the other hand it would wrap everything, reduce the transparency and make it much harder for (experienced) users to transform the training process to there own needs :sweat_smile:

kuraga:
> the training scripts / references folder is more a "collection" of scripts

Aaahh, didn't catch it. My fault. Training of docTR is Mindee's inner job at most. Thanks, @felixdittrich92 :)

P.S. Add READMEs to `/references`, etc. :)

felixdittrich92:
Mh not directly we use the same scripts for pre-training all the models but users can use it also for fine-tuning on there own needs :)

The inner task specific folders have already README's :)
For example: https://github.com/mindee/doctr/tree/main/references/recognition

felixdittrich92:
And the corresponding doc: https://mindee.github.io/doctr/latest/using_doctr/custom_models_training.html

kuraga:
Well, if the mechanism is public I'll illustrate my

> I don't ask "to implement other loggers" but about some abstract layer...

: the `clearml_log: bool` argument could be `logger: logging.Logger` :)

kuraga:
Update: hhm, now [I see](https://github.com/clearml/clearml/blob/v1.17.0/clearml/logger.py#L37) `clearml.Logger` **does not** inherit `logging.Logger`...

ThanhNX0401:
Is there any reason why docTR don't output the Training loss log to WB? How can i tell if the model is overfit or underfit?

felixdittrich92:
> Is there any reason why docTR don't output the Training loss log to WB? How can i tell if the model is overfit or underfit?

Oh you are right we should do this for every train script the same way it's done with the `clearml_log` passed to the train function. Would you like to open a PR ?

felixdittrich92:
@ThanhNX0401 Fixed now on main branch 😊

PR #1847, opened 2025-01-21
[references] Update Logging

felixdittrich92:
This PR:

- Small logging updates
- Log train loss and lr also for each epoch
- Progress bar show step val loss
- Progress bar show lr changes on each step
- Instead of print use tqdm to write / comes together with the tqdm logging


odulcy-mindee (approved):
Awesome, thank you very much ! :partying_face: 
I saw you spotted few mistakes and typo as well, thanks !!

PR #1857, opened 2025-01-28
Switch to opencv-python-headless

fabiozappo:
## Description
Replace `opencv-python` with `opencv-python-headless` in project dependencies. The headless version provides the same OpenCV functionality but without GUI and display capabilities, making it more suitable for server environments and reducing unnecessary dependencies.

## Changes
- Updated dependency from `opencv-python>=4.5.0,<5.0.0` to `opencv-python-headless>=4.5.0,<5.0.0` in pyproject.toml

## Benefits
- Reduced package size and installation footprint
- Eliminates unnecessary GUI dependencies
- Better suited for server/cloud environments
- Same functionality for core OpenCV operations

## Version Constraints
Maintained the same version constraints (>=4.5.0,<5.0.0) to ensure compatibility


felixdittrich92:
Hi @fabiozappo :wave:;

Thanks for the PR :+1: 
Unfortunately that's not possible this way, because many other packages depends on python-opencv which would be conflicting then with docTR.
If you are looking for something prod optimized take a look here (which provides also `headless` install options): https://github.com/felixdittrich92/OnnxTR

felixdittrich92:
CC @odulcy-mindee

PR #1903, opened 2025-03-21
[misc] Avoid tf onnx test to fail the CI & mypy fixes

felixdittrich92:
This PR:

- Apply mypy latest version fixes & upgrade pre-commit
- Wrap the TensorFlow export test into xfail to avoid our CI to fail here 

Additional information:

Unfortunately tf2onnx is not well maintained - so I don't expect a quick fix (patch release) here in the next 2-3 months - pinning any deps because of this one failing TF model export / load is not an option


SiddhantBahuguna (approved):
Thanks!

PR #1909, opened 2025-03-26
[FIXED] Add proper Hebrew Unicode chars vocab

johnlockejrr:
I hope I fixed the issue.


felixdittrich92 (approved):
Yep looks better now thanks for the fix :+1:  :smiley:

Issue #1911, opened 2025-03-27
Unused parameters in parseq.


IvanBarabanau:
Hi! Have you tested the ddp recognition training with parseq?
The training fails due to the unused parameters, declared right[here](https://github.com/mindee/doctr/blob/main/doctr/models/recognition/parseq/pytorch.py#L75-L76).
This parameters are also unused for regular training, obviously.

felixdittrich92:
Hi @IvanBarabanau 👋🏼,
Thanks for reporting 👍🏼
You are right I will push a fix tomorrow which is compatible for already trained models.

felixdittrich92:
After double checking I can't fix this right now because it would break already trained models which are loaded from local.
But that's a good starting point to unify the model loading between from url (our pretrained), custom from local and from hf hub.
Afterwards we can fix the mentioned issue without breaking any already trained models.

So for the moment I would suggest to go with the non ddp training script for parseq training.

PR #1912, opened 2025-03-27
[Feat] PyTorch - VIP backbone and VIPTR recognition module

lkosh:
The code has been tested by running training+evaluation for 8 epochs on a single gpu using an 800k multilingual  dataset for training and a 200k multilingual dataset for evaluation. While training loss decreased during training, I didn't observe any substantial increase in precision on the eval dataset. I'm not sure if that's the expected behavior, but in any case I didn't find any errors that would cause that. Let me know if there's something I missed 🙂

PR #1915, opened 2025-03-28
[Feat] Simplify and unify model loading - from_pretrained

felixdittrich92:
## soft breaking change ## 

- next release notes: `from_pretrained` should be used instead of manual loading
- especially with `parseq` it would fail to load "old" weights after the next release by loading manual instead of `from_pretrained`

## This PR: ##

- Unify model loading between TF & PT (more control on our end)
- Simplify custom model loading & update docs
- The following allows users to load from local file or new also from url

```python
# NEW
model = vitstr(pretrained=False)
model.from_pretrained(...)  # local path or url to .pt or .h5

# Instead of depending on the backend
reco_params = torch.load('<path_to_pt>', map_location="cpu")
reco_model.load_state_dict(reco_params)
# Or
reco_model.load_weights(..)
```

- Fix a issue in `parseq` by keeping it backward compatible - for already trained models  ref.: #1911 
- Lightweight dep (only python > 3.8) was added for url validation / https://github.com/python-validators/validators/blob/master/pyproject.toml#L38

design: 
- Normally I would prefer a base class with such functionality where each model inerhits from - but this would require a larger refactoring especially for the torchvision / keras imported backbone (classification) models
- From now every model needs to have the `from_pretrained` method -> unittest


doctr/models/recognition/parseq/tensorflow.py
```diff
@@ -76,8 +76,6 @@ def __init__(
             d_model, ffd * ffd_ratio, dropout, layers.Activation(tf.nn.gelu)
         )
 
-        self.attention_norm = layers.LayerNormalization(epsilon=1e-5)
```
felixdittrich92, line 79: unused layers

doctr/models/classification/magc_resnet/pytorch.py
felixdittrich92, line 1: MAGC doesn't needs it's own `from_pretrained` method because it's only a Layer plugged into the ResNet implementation

.github/workflows/references.yml
```diff
@@ -219,10 +219,10 @@ jobs:
           pip install -e .[torch,viz,html] --upgrade
       - if: matrix.framework == 'tensorflow'
         name: Evaluate text recognition (TF)
-        run: python references/recognition/evaluate_tensorflow.py crnn_mobilenet_v3_small --dataset IIIT5K -b 32
+        run: python references/recognition/evaluate_tensorflow.py crnn_mobilenet_v3_small --dataset SVT -b 32
```
felixdittrich92, line 222: IIIT5K download url is down ..again .. wrote already an mail .. use another dataset for the CI test

sebastianMindee:
@felixdittrich92 Since most of the implementations are the exact same, wouldn't it be better to just have a common definition & just import it when needed (ofc keep the different ones in their respective classes)? I feel like having this many duplicated bits is a sure way to forget about it when it matters down the road :sweat_smile:

felixdittrich92:
> @felixdittrich92 ~Since most of the implementations are the exact same, wouldn't it be better to just have a common definition & just import it when needed (ofc keep the different ones in their respective classes)? I feel like having this many duplicated bits is a sure way to forget about it when it matters down the road 😅~
> 
> Edit: misread the implementation. But it does feel confusing that `from_pretrained` implementations are all made into methods just to relay params to `load_pretrained_params`? Why not just add the non-standard implementations to `load_pretrained_params` instead?

@sebastianMindee 
Yeah in general the cleanest way would be to have a base class which implements already the functionality and all models inerhit from them but this would require a larger refactoring as mentioned and I think would raise an circular import issue.

So the idea of the wrapper method was that we can use this place for custom model logic (like `parseq` in this case).
But I agree we could also modify `load_pretrained_params` slighty make it private, import from `models.utils` and bind it afterwards `model.load_pretrained_params = types.MethodTypes(model, '_load_pretrained_params')` in this case we would reduce the most boilerplate code ..I tried to avoid this case as much as possible because anyway it feels a bit hacky so I used it only at places where a larger refactoring would be required :sweat_smile: 

wdyt ?

btw we wouldn't miss to implement this method because if we add the corresponding test entries the CI would fail (Added a unittest to check that all models have the `from_pretrained` method)

felixdittrich92:
@sebastianMindee 

After double checking I have to revert the above case because it would result with the same issue as modifying `load_pretrained_params` directly - results in circular import 

And I'm not open to switch to full lazy loading by skipping anything non matching :sweat_smile: 

```
from ..recognition import PARSeq  # circular

def load_pretrained_params(
    model: nn.Module,
    path_or_url: str | None = None,
    hash_prefix: str | None = None,
    ignore_keys: list[str] | None = None,
    **kwargs: Any,
) -> None:
    """Load a set of parameters onto a model

    >>> from doctr.models import load_pretrained_params
    >>> load_pretrained_params(model, "https://yoursource.com/yourcheckpoint-yourhash.pt")

    Args:
        model: the PyTorch model to be loaded
        path_or_url: the path or URL to the model parameters (checkpoint)
        hash_prefix: first characters of SHA256 expected hash
        ignore_keys: list of weights to be ignored from the state_dict
        **kwargs: additional arguments to be passed to `doctr.utils.data.download_from_url`
    """
    if path_or_url is None:
        logging.warning("No model URL or Path provided, using default initialization.")
        return

    archive_path = (
        download_from_url(path_or_url, hash_prefix=hash_prefix, cache_subdir="models", **kwargs)
        if validators.url(path_or_url)
        else path_or_url
    )

    if isinstance(model, PARSeq):
        # NOTE: This is required to make the model backward compatible with already trained models docTR version <0.11.1
        # ref.: https://github.com/mindee/doctr/issues/1911
        if ignore_keys is None:
            ignore_keys = []
        ignore_keys.extend([
            "decoder.attention_norm.weight",
            "decoder.attention_norm.bias",
            "decoder.cross_attention_norm.weight",
            "decoder.cross_attention_norm.bias",
        ])

...
```


An option would be to check only for the name :see_no_evil: 

`if type(model).__name__ == "PARSeq":`

felixdittrich92:
Ok finally:

Option 1: 

We keep it as provided in the PR 

Advantage: 
- Clear structure for each model & class bound logic (no condition required about model type / name)
- Already known interface from a lot of other libs like `transformers` / `sentence-transformers` and so on -> class method `.from_pretrained`

Disadvantage:
- Produces some boilerplate code & a wrapper method
- Uses method binding at places where a ("to class wrapper") refactoring would be required otherwise


Option 2:

We handle anything in `load_pretrained_params` 

on user side then it would be:

```python
model = vitstr()
load_pretrained_params(model, "path/to/weights.X")
```

Advantage:
- Avoids all the boilerplate
- Minimal changes required

Disadvantage:
- Model name condition bound in `load_pretrained_params`
- Feels less familiar to well known libs
- Provides additional options which could be a bit confusing for a simple loading function (hash_prefix, ignore_keys, ..) which are "masked" as kwargs in option 1


@sebastianMindee @SiddhantBahuguna Now we should decide which option would you prefer ? :smile: 

My vote goes to option 1 because from user view it feels a bit smarter :sweat_smile: AND in this case we have well unittested that each model provides a defined method to load weights - this case provides several checks - `pretrained=True` would fail if the model does not implement the method (because of the `model.from_pretrained` call) and the added `assert hasattr` checks also

sebastianMindee (commented):
LGTM but I'd like to wait for @SiddhantBahuguna's approval before giving it the official go :slightly_smiling_face:

felixdittrich92:
> LGTM but I'd like to wait for @SiddhantBahuguna's approval before giving it the official go 🙂

Yeah that's fine ☺️

felixdittrich92:
Will update the missing part for #1912 next week :)

felixdittrich92:
Done :)

sebastianMindee (approved):

PR #1921, opened 2025-04-18
Enable profiling with yappi

lachesis:


felixdittrich92:
Hi @lachesis 👋,

About the cuda addition I'm fine but I think profiling would fit better in https://github.com/mindee/doctr/blob/main/scripts/evaluate.py

Wdyt ?

lachesis:
Yes that would probably be better. I just found that script to be somewhat overwhelming and I understood this one better. The main reason to add profiling was to collect data to support my other PR #1922.

lachesis:
FWIW I probably will not update this to profile the evaluation script. I would not be offended if you close this PR. I created it primarily to document the process that I used. I'm not sure if it would be generally applicable to other doctr users.

felixdittrich92:
@lachesis ok no problem I added something to OnnxTR so if we need it at any point we can copy&paste it from there (https://github.com/felixdittrich92/OnnxTR/blob/main/scripts/evaluate.py) 👍

PR #1923, opened 2025-04-20
[datasets] feat: add croatian vocabulary

cyanic-selkie:
I have two questions:

1. How do you want to handle languages that use a dual writing system (e.g., Serbian)? Do you want to be able to address them individually so they can be blacklisted/whitelisted or just put them in the same bucket?
2. Can I contribute wordlists? I don't know how you're generating them at the moment.


felixdittrich92 (approved):
Thanks for the PR @cyanic-selkie :+1: 

1. Correct in this case it would be `serbian_latin` and `serbian_cyrillic` :+1: 
2. I will create a HF space which provides a way to upload resources like workdlists and fonts - because I would like to avoid adding such resource files to the repo directly

PR #1926, opened 2025-04-28
[datasets] Extend latin based vocabs

felixdittrich92:
This PR:

- Add several latin based vocabs (this should cover a really large range already)


felixdittrich92:
#1928

PR #1933, opened 2025-05-08
Migrating training scripts to torchrun

lkosh:
This PR:

- Removes standalone DDP training scripts for detection & recognition
- Migrates DDP logic into the default training scripts - by using torchrun
- Updates necessary doc / readme sections


references/detection/train_pytorch_ddp.py
```diff
@@ -567,14 +572,4 @@ def parse_args():
...
-        args.devices = list(range(torch.cuda.device_count()))
-    # no of process per gpu
-    nprocs = len(args.devices)
-    # Environment variables which need to be
-    # set when using c10d's default "env"
-    # initialization mode.
-    os.environ["MASTER_ADDR"] = "localhost"
-    os.environ["MASTER_PORT"] = "29500"
-    mp.spawn(main, args=(nprocs, args), nprocs=nprocs, join=True)
+    main(args)
```
felixdittrich92, line 575: update main also : current -> `def main(rank: int, world_size: int, args):` -> `def main(rank: int, world_size: int, args):`

references/recognition/README.md
```diff
@@ -25,27 +25,36 @@ or PyTorch:
 python references/recognition/train_pytorch.py crnn_vgg16_bn --train_path path/to/your/train_set --val_path path/to/your/val_set --epochs 5
 ```
 
-### Multi-GPU support (PyTorch only - Experimental)
+### Multi-GPU support (PyTorch only )
```
felixdittrich92, line 28: `(PyTorch only )` -> `(PyTorch only)`

references/recognition/train_pytorch_ddp.py
```diff
@@ -117,13 +116,20 @@ def evaluate(model, device, val_loader, batch_transforms, val_metric, amp=False)
...
         rank (int): device id to put the model on
         world_size (int): number of processes participating in the job
         args: other arguments passed through the CLI
     """
+    # Setup device and distributed
+    rank = int(os.environ.get("LOCAL_RANK", 0))
+    torch.cuda.set_device(rank)
+    dist.init_process_group(backend=args.backend)
+
+    world_size = dist.get_world_size()
```
felixdittrich92, line 131: that's not required anymore right ?

felixdittrich92 (changes requested):
Thanks :+1: 

Mh Normally in this case I think we can merge both scripts into one (DDP & the normal train script) ? - because the logic is the same - anyway what we should test is that the logging does still work with torchrun (W&B for example)


```
if args.backend:
       torch.cuda.set_device(rank)
       dist.init_process_group(backend=args.backend)
```

references/detection/train_pytorch_ddp.py
lkosh: already noticed that myself :)

references/recognition/train_pytorch_ddp.py
lkosh: world_size is only used in DistributedSampler, but it seems to work fine without this parameter

references/recognition/README.md
```diff
@@ -25,27 +25,36 @@ or PyTorch:
...
+- `--rdzv_backend`, `--rdzv_endpoint`, `--rdzv_id`
+  (Optional) Rendezvous settings for multi-node jobs. See the [torchrun docs](https://pytorch.org/docs/stable/elastic/run.html) for details.
+
+#### GPU selection:
+By default all visible GPUs will be used. To limit which GPUs participate, set the `CUDA_VISIBLE_DEVICES` environment variable **before** running `torchrun`. For example, to use only CUDA devices 0 and 2:
+
 ```shell
-python references/recognition/train_pytorch_ddp.py crnn_vgg16_bn --train_path path/to/your/train_set --val_path path/to/your/val_set --epochs 5 --devices 0 1 --backend nccl
+CUDA_VISIBLE_DEVICES=0,2 \
+torchrun --nproc_per_node=2 references/recognition/train_pytorch_ddp.py \
```
felixdittrich92, line 48: `train_pytorch.py` instead

references/recognition/train_pytorch.py
```diff
@@ -36,91 +40,20 @@
 from doctr.datasets import VOCABS, RecognitionDataset, WordGenerator
 from doctr.models import login_to_hub, push_to_hf_hub, recognition
 from doctr.utils.metrics import TextMatch
-from utils import EarlyStopper, plot_recorder, plot_samples
```
felixdittrich92, line 39: we should keep this `record_lr` function please revert the removal

references/recognition/train_pytorch.py
```diff
@@ -190,88 +122,123 @@ def evaluate(model, val_loader, batch_transforms, val_metric, amp=False, log=Non
 
 
 def main(args):
+    """
```
felixdittrich92, line 125: remove docstring

references/recognition/train_pytorch.py
```diff
@@ -190,88 +122,123 @@ def evaluate(model, val_loader, batch_transforms, val_metric, amp=False, log=Non
...
+    Args:
+        rank (int): device id to put the model on
+        world_size (int): number of processes participating in the job
+        args: other arguments passed through the CLI
+    """
+    # Detect distributed setup
+    world_size = int(os.environ.get("WORLD_SIZE", 1))
```
felixdittrich92, line 132: world_size = dist.get_world_size() ?

references/recognition/train_pytorch.py
```diff
@@ -190,88 +122,123 @@ def evaluate(model, val_loader, batch_transforms, val_metric, amp=False, log=Non
...
+    distributed = world_size > 1
+
+    # GPU setup
+    if distributed:
+        rank = int(os.environ.get("LOCAL_RANK", 0))
```
felixdittrich92, line 137: rank = dist.get_rank()

this should be only != 0 if multi gpus are available right ?

references/recognition/train_pytorch.py
```diff
@@ -425,12 +395,6 @@ def main(args):
             weight_decay=args.weight_decay or 1e-4,
         )
 
-    # LR Finder
```
felixdittrich92, line 428: restore this please with if rank == 0

references/recognition/train_pytorch.py
```diff
@@ -443,30 +407,32 @@ def main(args):
...
+            "learning_rate": args.lr,
+            "epochs": args.epochs,
+            "weight_decay": args.weight_decay,
+            "batch_size": args.batch_size,
+            "architecture": args.arch,
+            "input_size": args.input_size,
+            "optimizer": args.optim,
+            "framework": "pytorch",
+            "scheduler": args.sched,
+            "train_hash": train_hash,
```
felixdittrich92, line 421: "vocab": args.vocab  missing here

references/recognition/train_pytorch.py
felixdittrich92: Additional I think a env var `DOCTR_DISABLE_DDP` would be good here to allow users with multi gpu system to run on a single gpu otherwise it will always default to ddp -> the env var needs to be documented in the readme section

references/recognition/train_pytorch.py
```diff
@@ -443,30 +407,32 @@ def main(args):
...
     global global_step
     global_step = 0  # Shared global step counter
-
     # W&B
-    if args.wb:
-        import wandb
-
-        run = wandb.init(name=exp_name, project="text-recognition", config=config)
+    if rank == 0 and args.wb:
+        run = wandb.init(
```
felixdittrich92, line 431: `import wandb` should be here and not on the top of the script

references/recognition/train_pytorch.py
```diff
@@ -506,7 +472,8 @@ def clearml_log_at_step(train_loss=None, val_loss=None, lr=None):
                     value=lr,
                 )
 
-    # Unified logger
+        # Unified logger
```
felixdittrich92, line 475: remove comment

references/recognition/train_pytorch.py
```diff
@@ -517,83 +484,92 @@ def log_at_step(train_loss=None, val_loss=None, lr=None):
...
+        if args.push_to_hub:
+            push_to_hf_hub(model, exp_name, task="recognition", run_config=args)
 
 
 def parse_args():
     import argparse
 
     parser = argparse.ArgumentParser(
-        description="DocTR training script for text recognition (PyTorch)",
+        description="DocTR torchrun training script for text recognition (PyTorch)",
```
felixdittrich92, line 562: -torchrun

references/recognition/train_pytorch.py
```diff
@@ -646,7 +640,6 @@ def parse_args():
         "--sched", type=str, default="cosine", choices=["cosine", "onecycle", "poly"], help="scheduler to use"
     )
     parser.add_argument("--amp", dest="amp", help="Use Automatic Mixed Precision", action="store_true")
-    parser.add_argument("--find-lr", action="store_true", help="Gridsearch the optimal LR")
```
felixdittrich92, line 649: should be restored

references/recognition/train_pytorch.py
```diff
@@ -180,7 +113,6 @@ def evaluate(model, val_loader, batch_transforms, val_metric, amp=False, log=Non
 
         pbar.set_description(f"Validation loss: {out['loss'].item():.6}")
         log(val_loss=out["loss"].item())
```
felixdittrich92, line 115:
```
if log:
   log(val_loss=out["loss"].item())
```

references/recognition/train_pytorch.py
```diff
@@ -657,4 +650,6 @@ def parse_args():
 
 if __name__ == "__main__":
     args = parse_args()
+    if not torch.cuda.is_available():
```
felixdittrich92, line 653: We should remove this here it's handled in the main function

felixdittrich92 (changes requested):

references/detection/train_pytorch.py
```diff
@@ -470,58 +523,62 @@ def log_at_step(train_loss=None, val_loss=None, lr=None):
...
+            val_loss, recall, precision, mean_iou = evaluate(
+                model, val_loader, batch_transforms, val_metric, args, amp=args.amp, log=log_at_step
+            )
+            if val_loss < min_loss:
+                pbar.write(f"Validation loss decreased {min_loss:.6} --> {val_loss:.6}: saving state...")
+                params = model.module if hasattr(model, "module") else model
+                torch.save(params.state_dict(), Path(args.output_dir) / f"{exp_name}.pt")
+                min_loss = val_loss
+            if args.save_interval_epoch:
+                pbar.write(f"Saving state at epoch: {epoch + 1}")
```
felixdittrich92, line 541: for interval saving also:

```
params = model.module if hasattr(model, "module") else model
torch.save(params.state_dict(), Path(args.output_dir) / f"{exp_name}_epoch{epoch + 1}.pt")
```

felixdittrich92 (changes requested):

references/detection/train_pytorch.py
```diff
@@ -185,65 +192,105 @@ def evaluate(model, val_loader, batch_transforms, val_metric, amp=False, log=Non
...
+        )
+        pbar.write(
+            f"Validation set loaded in {time.time() - st:.4}s ({len(val_set)} samples in {len(val_loader)} batches)"
+        )
+        with open(os.path.join(args.val_path, "labels.json"), "rb") as f:
+            val_hash = hashlib.sha256(f.read()).hexdigest()
+
+        class_names = val_set.class_names
+    else:
+        class_names = None
```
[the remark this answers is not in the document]
lkosh: It looks like class_names can't be set like this in a distributed training : the variable will exist only on the zero-ranked process, and the rest of the processes will have class_names=None. This is fine when there's only one class, but in case of multiple classes it should have caused problems... If I remove the "else", then the script fails (because on non-zero ranks val_set is undefined). The only fix I can think of is broadcasting the class_names from rank=0 to other ranks, it doesn't look pretty but it works :)

Issue #1936, opened 2025-05-12
Low text recognition accuracy for long words


Razlaw:
### Bug description

First of all, thanks a lot for creating and maintaining docTR :)
It's a really nice open source package and it looks like you are putting a lot of work into it.

I noticed some inconsistencies when using the RecognitionPredictor on long words.
Here is an example crop from a receipt:

![Image](https://github.com/user-attachments/assets/315bd653-c6b2-499b-8b55-a9c3e6d4b81b)

And here the same crop right before the text recognition:

![Image](https://github.com/user-attachments/assets/85128e2d-7d4a-4cca-847b-bb9b1f9070b8)

I saw in an old issue #521 that the proposed workaround was to increase the target aspect ratio for the splitting of the crops.
This didn't help though.
Actually the crop isn't split at all.

### Code snippet to reproduce the bug

--

### Error traceback

--

### Environment

DocTR version: 0.11.1a0
TensorFlow version: N/A
PyTorch version: 2.7.0+cu126 (torchvision 0.22.0+cu126)
OpenCV version: 4.11.0
OS: Ubuntu 22.04.5 LTS
Python version: 3.10.12
Is CUDA available (TensorFlow): N/A
Is CUDA available (PyTorch): No
CUDA runtime version: Could not collect
GPU models and configuration: Could not collect
Nvidia driver version: Could not collect
cuDNN version: Could not collect

### Deep Learning backend

Both not available, running on CPU only.

Razlaw:
Btw here is the result:

![Image](https://github.com/user-attachments/assets/2a3bfccb-52f7-4dbe-a074-c48572f4b883)

I'm currently looking deeper into the code and the example crop should trigger the splitting..

Razlaw:
So, apparently this line should calculate the number of subcrops the crop should be split into
https://github.com/mindee/doctr/blob/db6d0dbe9aef671695b0f6d725196bec3c7d9a7d/doctr/models/recognition/predictor/_utils.py#L41
Putting in the values for the example we have an aspect_ratio of roughly 11.37 = 1126 / 99 and a default target_ratio of 6, I would expect to get 2 subcrops, but the result is 1 = 11.37 // 6.
But even when I adapted this, results didn't improve really.
Should I continue investigating or am I completely on the wrong track? :)

felixdittrich92:
Hi @Razlaw 👋,

[...]

PR #1939, opened 2025-05-16
[Fix] Improve merge for recognition of long words (#1936)

Razlaw:
Hello again @felixdittrich92 :)

I took a deeper look into the splitting and merging of the crops for the recognition and came up with a first version we can have a look at.
I tried to stick to the original approach and keep it simple.
You've had the idea to dynamically compute the hyperparameters for the aspect ratios per crop.
That's not implemented yet, but we I guess this PR is work in progress anyway :)

The main idea so far is:
- Split the crop into equally sized sub-crops with known overlap ratios
- Use the known overlap of neighboring sub-crops to estimate the overlap when merging recognized strings

This worked well for the examples I provided in the issue #1936.

Sub-crops right before being processed by recognizer - with recognized string per sub-crop:
crops_for_recognition_after_preprocessing_batch_0_file_0
2025-02-20T
crops_for_recognition_after_preprocessing_batch_0_file_1
2-20T06:51:
crops_for_recognition_after_preprocessing_batch_0_file_2
06:51:13.000
crops_for_recognition_after_preprocessing_batch_0_file_3
:51:13.000Z
Merged result: 2025-02-20T06:51:13.000Z 
:white_check_mark: 

crops_for_recognition_after_preprocessing_batch_0_file_0
XUrA+b3iHQP
crops_for_recognition_after_preprocessing_batch_0_file_1
3iHQP+F3qxu
crops_for_recognition_after_preprocessing_batch_0_file_2
+F3qxuOZZDR:
crops_for_recognition_after_preprocessing_batch_0_file_3
OZZDRA/628+
crops_for_recognition_after_preprocessing_batch_0_file_4
RA/6Z8+6nbf
Merged result: XUrA+b3iHQP+F3qxuOZZDRA/628+6nbf 
Not perfectly correct but at least correctly merged :)

I used the evaluation script from the repo to check how the changes affect the scores - thanks for providing that script btw.
To summarize, the results are actually not that different to the existing approach..
But at least significantly faster on the CORD dataset.
I've looked into the data and my hypothesis is that there are mostly short words and numbers that do not benefit from the split and merge.

Additionally I ran an evaluation on a dataset we labeled ourselves.
It consists of 160 photographies of german shopping receipts - so it includes at least some lengthy words ;)
The average accuracy improved by 1%.
For dates - consisting of the long datetimes as shown in the example, but also shorter versions of just time or date - the accuracy improved by ~7%.
It's not much, but at least something :)

I'm eager to hear your feedback about the changes. No pressure though ;) 

---

Appendix:
The full results of the evaluation on FUNSD and CORD
Variant 2 and 3 use the same critical_ar and target_ar parameters of 8 and 6 - the default from docTR - to generate sub-crops of comparable sizes. Mostly relevant to compare the run durations. 
Variant 4 uses critical_ar of 4 and target_ar of 4, because this utilizes the input window (32 times 128 pixels) of the recognizer fully.
This variant worked best on our data, but had slightly worse results on FUNSD and CORD.

- Models: db_resnet50 + parseq
- **FUNSD**
	- **1 docTR main**
		- Text Recognition - Accuracy: 86.67% (unicase: 87.47%) OCR - Recall: 73.46% (unicase: 73.99%), Precision: 76.20% (unicase: 76.75%), Mean IoU: 68.00%  
			- 149/149 [30:44<00:00, 12.38s/it]
			- 50/50 [12:03<00:00, 14.48s/it]
	- **2 docTR main - only with fixed computation of number of sub-crops**
		- Text Recognition - Accuracy: 86.46% (unicase: 87.25%) OCR - Recall: 73.34% (unicase: 73.87%), Precision: 76.08% (unicase: 76.63%), Mean IoU: 68.00%  
			- 149/149 [31:42<00:00, 12.77s/it]
			- 50/50 [12:17<00:00, 14.75s/it]
	- **3 Fix with crop parameters similar to version 2**
		- Text Recognition - Accuracy: 86.60% (unicase: 87.40%) OCR - Recall: 73.45% (unicase: 73.98%), Precision: 76.19% (unicase: 76.74%), Mean IoU: 68.00%  
			- 149/149 [32:01<00:00, 12.90s/it]
			- 50/50 [12:14<00:00, 14.69s/it]
	- **4 Fix with full utilization of recognition input window**
		- Text Recognition - Accuracy: 85.71% (unicase: 86.51%) OCR - Recall: 73.14% (unicase: 73.63%), Precision: 75.87% (unicase: 76.38%), Mean IoU: 68.00%

- **CORD - ran on different machine so timings not comparable to FUNSD**
	- **1 docTR main**
		- Text Recognition - Accuracy: 91.71% (unicase: 92.07%) OCR - Recall: 85.03% (unicase: 85.34%), Precision: 79.30% (unicase: 79.58%), Mean IoU: 73.00%
			- 800/800 [38:29<00:00,  2.89s/it]
			- 100/100 [04:40<00:00,  2.80s/it]
	- **2 docTR main - only with fixed computation of number of sub-crops**
		- Text Recognition - Accuracy: 91.68% (unicase: 92.04%) OCR - Recall: 85.03% (unicase: 85.32%), Precision: 79.30% (unicase: 79.57%), Mean IoU: 73.00%
			- 800/800 [38:22<00:00,  2.88s/it]
			- 100/100 [04:38<00:00,  2.78s/it]
	- **3 Fix with crop parameters similar to version 2**
		- Text Recognition - Accuracy: 91.72% (unicase: 92.08%) OCR - Recall: 85.09% (unicase: 85.39%), Precision: 79.35% (unicase: 79.63%), Mean IoU: 73.00%  
			- 800/800 [27:01<00:00,  2.03s/it]
			- 100/100 [03:19<00:00,  1.99s/it]
		- Significantly faster with similar results - maybe because of Hamming distance instead of Levenshtein
	- **4 Fix with full utilization of recognition input window**
		- Text Recognition - Accuracy: 90.16% (unicase: 90.57%) OCR - Recall: 84.25% (unicase: 84.55%), Precision: 78.57% (unicase: 78.85%), Mean IoU: 73.00%
			- 800/800 [38:36<00:00,  2.90s/it]
			- 100/100 [04:30<00:00,  2.71s/it]

### DIFF
--- a/doctr/models/recognition/predictor/_utils.py
+++ b/doctr/models/recognition/predictor/_utils.py
@@ -4,6 +4,8 @@
 # See LICENSE or go to <https://opensource.org/licenses/Apache-2.0> for full license details.
 
 
+from math import ceil, floor
+
 import numpy as np
 
 from ..utils import merge_multi_strings
@@ -15,60 +17,94 @@ def split_crops(
     crops: list[np.ndarray],
     max_ratio: float,
     target_ratio: int,
-    dilation: float,
+    target_overlap_ratio: float,
     channels_last: bool = True,
-) -> tuple[list[np.ndarray], list[int | tuple[int, int]], bool]:
-    """Chunk crops horizontally to match a given aspect ratio
+) -> tuple[list[np.ndarray], list[int | tuple[int, int, float]], bool]:
+    """Split crops horizontally to match a given aspect ratio
 
     Args:
+    ----
         crops: list of numpy array of shape (H, W, 3) if channels_last or (3, H, W) otherwise
-        max_ratio: the maximum aspect ratio that won't trigger the chunk
-        target_ratio: when crops are chunked, they will be chunked to match this aspect ratio
-        dilation: the width dilation of final chunks (to provide some overlaps)
+        max_ratio: the maximum aspect ratio that won't trigger the splitting
+        target_ratio: when a crop are split, it will be split to match this aspect ratio.
+          The default value of 4 corresponds to the recognition models' input size of 32 times 128 pixels
+        target_overlap_ratio: the target ratio how much a split overlaps with a neighboring split
         channels_last: whether the numpy array has dimensions in channels last order
 
     Returns:
+    -------
         a tuple with the new crops, their mapping, and a boolean specifying whether any remap is required
     """
     _remap_required = False
-    crop_map: list[int | tuple[int, int]] = []
+    splits_map: list[int | tuple[int, int, float]] = []
     new_crops: list[np.ndarray] = []
     for crop in crops:
         h, w = crop.shape[:2] if channels_last else crop.shape[-2:]
-        aspect_ratio = w / h
-        if aspect_ratio > max_ratio:
-            # Determine the number of crops, reference aspect ratio = 4 = 128 / 32
-            num_subcrops = int(aspect_ratio // target_ratio)
-            # Find the new widths, additional dilation factor to overlap crops
-            width = dilation * w / num_subcrops
-            centers = [(w / num_subcrops) * (1 / 2 + idx) for idx in range(num_subcrops)]
-            # Get the crops
-            if channels_last:
-                _crops = [
-                    crop[:, max(0, int(round(center - width / 2))) : min(w - 1, int(round(center + width / 2))), :]
-                    for center in centers
-                ]
-            else:
-                _crops = [
-                    crop[:, :, max(0, int(round(center - width / 2))) : min(w - 1, int(round(center + width / 2)))]
-                    for center in centers
-                ]
+        actual_crop_ratio = w / h
+        if actual_crop_ratio > max_ratio:
+            target_split_width = ceil(h * target_ratio)
+            target_split_overlap_width = floor(target_split_width * target_overlap_ratio)
+
+            splits, last_overlap_ratio = _split_horizontally(
+                crop,
+                target_split_width,
+                target_split_overlap_width,
+                channels_last,
+            )
+
             # Avoid sending zero-sized crops
-            _crops = [crop for crop in _crops if all(s > 0 for s in crop.shape)]
+            splits = [split for split in splits if all(s > 0 for s in split.shape)]
             # Record the slice of crops
-            crop_map.append((len(new_crops), len(new_crops) + len(_crops)))
-            new_crops.extend(_crops)
+            splits_map.append((len(new_crops), len(new_crops) + len(splits), last_overlap_ratio))
+            new_crops.extend(splits)
             # At least one crop will require merging
             _remap_required = True
         else:
-            crop_map.append(len(new_crops))
+            splits_map.append(len(new_crops))
             new_crops.append(crop)
 
-    return new_crops, crop_map, _remap_required
+    return new_crops, splits_map, _remap_required
+
+
+def _split_horizontally(
+    image: np.ndarray, split_width: int, split_overlap_width: int, channels_last: bool
+) -> tuple[list[np.ndarray], float]:
+    image_width = image.shape[1] if channels_last else image.shape[-1]
+    if image_width <= split_width:
+        return [image], 0
+
+    splits = []
+    current_split_start_column = 0
+    previous_split_end_column = 0
+    last_overlap_ratio = 0.0
+    has_reached_end_of_image = False
+    while not has_reached_end_of_image:
+        current_split_end_column = current_split_start_column + split_width
+        current_split_end_column = min(current_split_end_column, image_width)
+
+        # Increase overlap of last split to prevent narrow last split
+        has_reached_end_of_image = current_split_end_column == image_width
+        if has_reached_end_of_image:
+            current_split_start_column = max(0, image_width - split_width)
+
+        if channels_last:
+            image_split = image[:, current_split_start_column:current_split_end_column, :]
+        else:
+            image_split = image[:, :, current_split_start_column:current_split_end_column]
+
+        # Save overlap ratio of the last split, because this one might be larger than other overlap ratios
+        if has_reached_end_of_image:
+            current_overlap_width = previous_split_end_column - current_split_start_column
+            last_overlap_ratio = current_overlap_width / split_width
+
+        splits.append(image_split)
+        previous_split_end_column = current_split_end_column
+        current_split_start_column = current_split_end_column - split_overlap_width
+    return splits, last_overlap_ratio
 
 
 def remap_preds(
-    preds: list[tuple[str, float]], crop_map: list[int | tuple[int, int]], dilation: float
+    preds: list[tuple[str, float]], crop_map: list[int | tuple[int, int, float]], split_overlap_ratio: float
 ) -> list[tuple[str, float]]:
     remapped_out = []
     for _idx in crop_map:
@@ -76,8 +112,9 @@ def remap_preds(
         if isinstance(_idx, int):
             remapped_out.append(preds[_idx])
         else:
-            # unzip
             vals, probs = zip(*preds[_idx[0] : _idx[1]])
+            last_split_overlap_ratio = _idx[2]
             # Merge the string values
-            remapped_out.append((merge_multi_strings(vals, dilation), min(probs)))  # type: ignore[arg-type]
+            merged_string = (merge_multi_strings(vals, split_overlap_ratio, last_split_overlap_ratio), min(probs))
+            remapped_out.append(merged_string)
     return remapped_out
--- a/doctr/models/recognition/utils.py
+++ b/doctr/models/recognition/utils.py
@@ -4,81 +4,117 @@
 # See LICENSE or go to <https://opensource.org/licenses/Apache-2.0> for full license details.
 
 
-from rapidfuzz.distance import Levenshtein
+from rapidfuzz.distance import Hamming
 
 __all__ = ["merge_strings", "merge_multi_strings"]
 
 
-def merge_strings(a: str, b: str, dil_factor: float) -> str:
+def merge_strings(a: str, b: str, overlap_ratio: float) -> str:
     """Merges 2 character sequences in the best way to maximize the alignment of their overlapping characters.
 
     Args:
         a: first char seq, suffix should be similar to b's prefix.
         b: second char seq, prefix should be similar to a's suffix.
-        dil_factor: dilation factor of the boxes to overlap, should be > 1. This parameter is
-            only used when the mother sequence is splitted on a character repetition
+        overlap_ratio: estimated ratio of overlapping characters.
 
     Returns:
         A merged character sequence.
 
     Example::
-        >>> from doctr.models.recognition.utils import merge_sequences
-        >>> merge_sequences('abcd', 'cdefgh', 1.4)
+        >>> from doctr.models.recognition.utils import merge_strings
+        >>> merge_strings('abcd', 'cdefgh', 0.33)
         'abcdefgh'
-        >>> merge_sequences('abcdi', 'cdefgh', 1.4)
+        >>> merge_strings('abcdi', 'cdefgh', 0.5)
         'abcdefgh'
     """
-    seq_len = min(len(a), len(b))
-    if seq_len == 0:  # One sequence is empty, return the other
-        return b if len(a) == 0 else a
+    lower_inputs_len = min(len(a), len(b))
+    if lower_inputs_len <= 1:  # One sequence is empty or will be after cropping in next step, return the other
+        return b if len(a) == lower_inputs_len else a
 
-    # Initialize merging index and corresponding score (mean Levenstein)
-    min_score, index = 1.0, 0  # No overlap, just concatenate
+    # Remove last letter of "a" and first of "b", because they might be cut off. The removed characters are still
+    # present in the other string due to overlap
+    a_crop = a[:-1]
+    b_crop = b[1:]
 
-    scores = [Levenshtein.distance(a[-i:], b[:i], processor=None) / i for i in range(1, seq_len + 1)]
+    # Increase overlap step by step and compute number of different characters per overlap
+    max_overlap = min(len(a_crop), len(b_crop))
+    scores = [Hamming.distance(a_crop[-i:], b_crop[:i], processor=None) for i in range(1, max_overlap + 1)]
 
-    # Edge case (split in the middle of char repetitions): if it starts with 2 or more 0
-    if len(scores) > 1 and (scores[0], scores[1]) == (0, 0):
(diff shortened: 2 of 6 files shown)


felixdittrich92:
Hey @Razlaw :wave:,

wow thanks a lot for the PR and the detailed benchmarks I will have a look on monday :+1: Have a nice weekend :)

Razlaw:
Nice job on the refactorings :) 
I've put them in some additional commits, so we can talk about the new state in more detail.
In the end, I'll clean and squash them.

Thanks for the test files :+1: 
Do you have any ground truth labels for those files? Then I wouldn't need to check "by hand" where the merging failed ^^

doctr/models/recognition/utils.py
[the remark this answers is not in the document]
Razlaw: > multi_string: (["aaaa", "aaab", "aabc"], 0.8, 0.3, "aaaabc"), <- IndexError

Interesting, I didn't get an IndexError for the first example, but there was a mistake for the repetitions. That mistake should be fixed now.

> string_merge: ("a", "b", 0.5, "ab"), <-- empty

A added a quick fix for the second example - it's very much up for discussion though. Anyhow, that example is probably an edge case when it comes to splitting long words and merging them back together ;)

tests/common/test_models_recognition_utils.py
```diff
@@ -1,31 +1,206 @@
 import pytest
 
-from doctr.models.recognition.utils import merge_multi_strings, merge_strings
+from doctr.models.recognition.utils import _get_index_closest_to_target_index, merge_multi_strings, merge_strings
 
 
+# Last character of first string and first of last string will be cropped when merging - indicated by X
 @pytest.mark.parametrize(
```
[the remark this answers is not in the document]
Razlaw: I'm not really sure if I understand your comment correctly.
The overlap ratio is what's being used to estimate where to merge the strings.
So we need to adapt it since the strings in the test cases change, right?

tests/common/test_models_recognition_utils.py
```diff
@@ -1,31 +1,206 @@
...
+        (["abcdX", "XdefX", "XefghiX", "Xaijk"], 0.5, 0.8, "abcdefghijk"),
     ],
 )
-def test_merge_multi_strings(seq_list, merged):
-    assert merged == merge_multi_strings(seq_list, 1.4)
+def test_merge_multi_strings(seq_list, overlap_ratio, last_overlap_ratio, merged):
+    assert merged == merge_multi_strings(seq_list, overlap_ratio, last_overlap_ratio)
+
+
+def test_example_from_docs_1():
```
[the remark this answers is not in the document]
Razlaw: Yes, I will refactor the tests once the code is stable.

tests/common/test_models_recognition_predictor.py
```diff
@@ -5,35 +5,160 @@
...
-    preds = remap_preds(preds, crop_map, dilation)
+def test_remap_preds(preds, crop_map, split_overlap_ratio, pred):
+    preds = remap_preds(preds, crop_map, split_overlap_ratio)
     assert len(preds) == len(pred)
     assert preds == pred
     assert all(isinstance(pred, tuple) for pred in preds)
     assert all(isinstance(pred[0], str) and isinstance(pred[1], float) for pred in preds)
+
+
+def test_dont_split_if_not_necessary():
```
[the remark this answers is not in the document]
Razlaw: Same here, I would do that once the code is in a stable state.

> Additional it should work with both channels_last = True | False the same way

Just to make sure I understand correctly. There is one test case that has the channels_last parameter set to False. Would you like to have all other test cases in a version with True and False?

tests/common/test_models_recognition_utils.py
```diff
@@ -136,12 +136,12 @@ def test_multiple_repetitions_shorter():
 
 
 def test_multiple_repetitions_full_overlap():
-    a_ = "ababX"
-    b = "Xabab"
```
Razlaw: These were impossible, because b is supposed to be merged to a from the right, but contains a character more on its left side than a does.

doctr/models/recognition/predictor/_utils.py
```diff
@@ -17,104 +17,126 @@ def split_crops(
...
+        A tuple containing:
+            - The new list of crops (possibly with splits),
+            - A mapping indicating how to reassemble predictions,
+            - A boolean indicating whether remapping is required.
     """
-    _remap_required = False
-    splits_map: list[int | tuple[int, int, float]] = []
+    remap_required = False
     new_crops: list[np.ndarray] = []
+    crop_map: list[int | tuple[int, int]] = []
```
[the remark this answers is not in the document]
felixdittrich92: Correct good catch :+1: 
Yeah I hacked this together in my lunch break .. that you don't have to wait to long for the review 😂

tests/common/test_models_recognition_predictor.py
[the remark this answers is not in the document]
felixdittrich92: Keep up with one test for `channels_last=False` :+1:

I faiced an incositents which goes trough the whole docTR lib .... 

We shouldn't except Tensors for the modular predictor's (incl. ocr_predictor) without handling them correctly (at least for pytorch) ... and maybe it would be also good to go with one fixed shape format (H, W, C) for the inference pipes 🙄
But that's out of scope for your PR

tests/common/test_models_recognition_utils.py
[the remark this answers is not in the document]
felixdittrich92: Forget this comment :sweat_smile: Everything right here :+1:

tests/common/test_models_recognition_utils.py
```diff
@@ -11,7 +11,7 @@
         ("abcdX", "Xdef", 0.75, "abcdef"),
         ("abcdeX", "Xdef", 0.9, "abcdef"),
         ("abcdefX", "Xdef", 0.9, "abcdef"),
-        ("abccccX", "Xcccccc", 0.4, "abcccccccc"),
+        ("abccccX", "Xcccccc", 0.5, "abcccccccc"),
```
felixdittrich92, line 14: Anyway this test case is really hard to understand :sweat_smile: 

So later on each should get and comment about the merge area :)

doctr/models/recognition/utils.py
```diff
@@ -59,7 +60,8 @@ def merge_strings(a: str, b: str, overlap_ratio: float) -> str:
         return a_crop + b_crop
 
     # Find best overlap by minimizing Hamming distance + distance from expected overlap size
-    combined_scores = [score + abs(i - expected_overlap) for i, score in enumerate(scores)]
+    expected_overlap_index = expected_overlap - 1
```
felixdittrich92, line 65: I think it would be better to adjust directly here (L43):

`expected_overlap = round(len(b) * overlap_ratio) - 2` to -3 ?

felixdittrich92:
> Nice job on the refactorings :) I've put them in some additional commits, so we can talk about the new state in more detail. In the end, I'll clean and squash them.
> 
> Thanks for the test files 👍 Do you have any ground truth labels for those files? Then I wouldn't need to check "by hand" where the merging failed ^^

Hey @Razlaw :)

Yep sounds good to me :+1: 
I double checked and it doesn't fails on split ..the model produces garbage in the most cases because of missing padding

Let me try to create annotations for the files today

tests/common/test_models_recognition_utils.py
Razlaw: Yes, true, it was even confusing when I wrote the test ^^ 
I added a comment. Is that okay or should I add more explanations?

Razlaw:
I didn't address the codecov warnings yet. There is one I'm not sure we can even address using a test. I'll leave another comment there.
Btw, you mentioned you wanted to test some possible further improvements. Would you like to keep the PR open?
Did you get the chance to generate the ground truth for the data you provided?

doctr/models/recognition/predictor/_utils.py
```diff
@@ -15,69 +17,126 @@ def split_crops(
...
-            # Avoid sending zero-sized crops
-            _crops = [crop for crop in _crops if all(s > 0 for s in crop.shape)]
-            # Record the slice of crops
-            crop_map.append((len(new_crops), len(new_crops) + len(_crops)))
-            new_crops.extend(_crops)
-            # At least one crop will require merging
-            _remap_required = True
+                # Fallback: treat it as a single crop
+                crop_map.append(len(new_crops))
+                new_crops.append(crop)
```
Razlaw, line 62: Is it even possible to get here?
Addressing codecov warning "Added lines #L61 - L62 were not covered by tests"

felixdittrich92: Mh this is an real extrem edge case but could happen if `overlap_width>=split_width` 

So two options:

1. We add 1 test case which covers this 
2. We ignore covering this - because it'S an real extrem edge case

```
else:  # pragma: no cover
    # Fallback: treat it as a single crop
    crop_map.append(len(new_crops))
    new_crops.append(crop)
 ```

File: doctr/models/recognition/utils.py [hunk 9]
```diff
@@ -4,81 +4,90 @@
...
     """
-
-    def _recursive_merge(a: str, seq_list: list[str], dil_factor: float) -> str:
-        # Recursive version of compute_overlap
-        if len(seq_list) == 1:
-            return merge_strings(a, seq_list[0], dil_factor)
-        return _recursive_merge(merge_strings(a, seq_list[0], dil_factor), seq_list[1:], dil_factor)
-
-    return _recursive_merge("", seq_list, dil_factor)
+    if not seq_list:
```
felixdittrich92, line 86: We can cover this by passing a empty list in the test cases :)

([], 0.5, 0.4, ""),

tests/common/test_models_recognition_predictor.py
felixdittrich92, line 1: This looks really good :+1:

tests/common/test_models_recognition_utils.py
```diff
@@ -4,28 +4,60 @@
...
     [
-        [["abc", "def"], "abcdef"],
-        [["abcd", "def", "efgh", "ijk"], "abcdefghijk"],
-        [["abcdi", "defk", "efghi", "aijk"], "abcdefghijk"],
+        # One character at each conjunction point will be cropped when merging - indicated by X
+        (["abcX", "Xdef"], 0.5, 0.5, "abcdef"),
+        (["abcdX", "XdefX", "XefghX", "Xijk"], 0.5, 0.5, "abcdefghijk"),
+        (["abcdX", "XdefX", "XefghiX", "Xaijk"], 0.5, 0.8, "abcdefghijk"),
+        (["aaaa", "aaab", "aabc"], 0.8, 0.3, "aaaabc"),
+        # Example from docs
```
felixdittrich92, line 58: same here :)

tests/common/test_models_recognition_utils.py
```diff
@@ -4,28 +4,60 @@
...
+        ("abcdX", "Xdef", 0.75, "abcdef"),
+        ("abcdeX", "Xdef", 0.9, "abcdef"),
+        ("abcdefX", "Xdef", 0.9, "abcdef"),
+        # Long repetition - four of seven characters in the second string are in the estimated overlap
+        # X-chars will be cropped during merge, because they might be cut off during splitting of corresponding image
+        ("abccccX", "Xcccccc", 4 / 7, "abcccccccc"),
+        ("abc", "", 0.5, "abc"),
+        ("", "abc", 0.5, "abc"),
+        ("a", "b", 0.5, "ab"),
+        # Example from docs 1
```
felixdittrich92, line 20: What does this mean "Example from docs 1" ? :)

Let's refine the two comments "Example from docs 1" & "Example from docs 2" to something meaningful like the other comments :)

doctr/models/recognition/predictor/_utils.py
```diff
@@ -15,69 +17,126 @@ def split_crops(
...
+        split_width: Width of each split.
+        overlap_width: Width of the overlapping region.
+        channels_last: Whether the image is in channels-last format.
+
+    Returns:
+        - A list of horizontal image slices.
+        - The actual overlap ratio of the last split.
+    """
+    image_width = image.shape[1] if channels_last else image.shape[-1]
+    if image_width <= split_width:
```
felixdittrich92, line 87: Let's add a tiny additional test for this :)

felixdittrich92 (changes requested):
Hi @Razlaw :wave:,

Thanks a lot for the updates only a few minor things left and we are good to merge :+1:

tests/common/test_models_recognition_utils.py
Razlaw: There is not much meaningful behind these test cases :)
They are just the usage examples from the docstrings of the methods
```
    """
    ...
    Example::
        >>> from doctr.models.recognition.utils import merge_strings
        >>> merge_strings('abcd', 'cdefgh', 0.5)
        'abcdefgh'
        >>> merge_strings('abcdi', 'cdefgh', 0.5)
        'abcdefgh'
    """
```
I wanted to make sure that those example cases work as intended.
I can change the comment at the test case to "Happy case" but that's not really meaningful either.
We can also remove those test cases if you want.
Any preference?

doctr/models/recognition/predictor/_utils.py
felixdittrich92: Mh .. I think it would be better to clip the value in this case - to avoid an hard exception

Additional we could add a logger warning ..but I think this would spam a bit to much :sweat_smile:

tests/common/test_models_recognition_utils.py
felixdittrich92: Let's drop :)

doctr/models/recognition/predictor/_utils.py
Razlaw: Yep, logging sounds like spam :) 
On the other hand, I would want to notice if I set a parameter to a bad value -> pro exception or spam. 
On the other other hand, those parameters will not be tuned frequently and anyone playing around with them should roughly know what they do..
You decide ;)

doctr/models/recognition/utils.py
Razlaw: Done :+1:

doctr/models/recognition/predictor/_utils.py
felixdittrich92: Yeah also true - then raise a ValueError because user input

Razlaw:
Regarding the mypy errors: I guess they don't originate from the code that we changed in this PR, right? How should we proceed here?

felixdittrich92:
> Regarding the mypy errors: I guess they don't originate from the code that we changed in this PR, right? How should we proceed here?

Correct - ignore them I will fix it later :)

PR #1958, opened 2025-06-18
[bug] fix missing import

felixT2K:


felixdittrich92:
CC @sebastianMindee It was really the only training script where it was missing :sweat_smile:

PR #1961, opened 2025-06-18
Fix: Update min_loss only on improved validation loss to ensure best …

sneakybatman:
## What does this PR fix?

Previously, model checkpoints were being saved whenever the validation loss decreased compared to the previous epoch, rather than being compared to the *minimum* validation loss so far. This could result in overwriting the best checkpoint with a suboptimal one.

## Solution

- Track the minimum validation loss seen so far.
- Only save the model if the current validation loss is lower than this minimum.

Tested on distributed multi-GPU training; now only the true best checkpoint is preserved.

### DIFF
--- a/references/recognition/train_pytorch.py
+++ b/references/recognition/train_pytorch.py
@@ -589,7 +589,7 @@ def main(args):
                 params = model.module if hasattr(model, "module") else model
 
                 torch.save(params.state_dict(), Path(args.output_dir) / f"{exp_name}.pt")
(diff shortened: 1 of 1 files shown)


felixdittrich92 (changes requested):
Let's remove the comment that should be clear without :)

sneakybatman:
Updated the code after removing the comment. Please review.

felixdittrich92:
> Updated the code after removing the comment. Please review.

Thanks :+1:

felixdittrich92 (approved):

PR #1962, opened 2025-06-19
[models] Drop viptr base config

felixT2K:
This PR:

- Drop viptr base config

Explanation:

We drop the base config because after some experiments the arch with it's base config doesn't learns well / needs tons of more data (10M +++)
After a param (config) grid search the official tiny config is close the best fit


felixdittrich92:
CC @sebastianMindee

felixdittrich92 (approved):

PR #1967, opened 2025-06-20
[BREAKING] Drop TensorFlow backend

felixT2K:
This PR:

- Remove all TensorFlow content
- Add PyTorch as fixed dependency
- Drop documentation versions which rely on TensorFlow backend
- Remove conda which does not support PyTorch builds anymore - since 2.15
- Remove unused code for channels_last / first switch 
- Fix typings & unify predictor & model inputs - predictors (numpy array HWC) - direct model calls (torch tensor CHW)


felixdittrich92:
NOTE: API test will fail until merged

felixdittrich92:
@sebastianMindee That should be everything for a first round to get rid of TensorFlow :sweat: :sweat_smile: 
API test should be green after it's merged

sebastianMindee (dismissed):

felixdittrich92 (approved):

Issue #1970, opened 2025-06-24
Potential bug in estimate_orientation function


Razlaw:
### Bug description

Hey :)
I might have found a small bug in [line 66 of the estimate_orientation](https://github.com/mindee/doctr/blob/d1974d9b5c53d4700e8ecaf0bbf1aa4f49d8728d/doctr/models/_utils.py#L66) function.

```
    page_orientation, orientation_confidence = general_page_orientation or (None, 0.0)
    if page_orientation and orientation_confidence >= min_confidence:
        ...
```

The page_orientation is estimated by the model and can be 0, 90, 180 or 270.
If it's 0, then the if statement interprets it as False and executes the else block.
I guess we rather want to check if it `is not None`, right?
Executing the else block leads to a dilation of the segmentation map, that is used for the fine grained orientation estimation.
Dilation leads to a merge of segmented lines.
For tabular data with small fonts this leads to a merge of the rows.
Those merged blobs are misinterpreted as lines which results in rotation estimates of 90 or -90 degrees.

Two images are probably better than a thousand words ;)

**Without dilation:**
![Image](https://github.com/user-attachments/assets/2ff7921e-c820-4ab3-98d4-1d9154b52853)

**With dilation:**
![Image](https://github.com/user-attachments/assets/348217f3-f662-4901-a8ae-bbd5143830d5)

### Code snippet to reproduce the bug

-

### Error traceback

-

### Environment

-

### Deep Learning backend

-

felixdittrich92:
Hi @Razlaw 👋,

Yeah totally right the else logic should only run if the page orientation isn't given or it's lower than the min_confidence 👍

Razlaw:
Yep, done :)

PR #1971, opened 2025-06-25
[Fix] Correct condition for image dilation in orientation estimation

Razlaw:
#1970


Razlaw:
> @Razlaw I think at line 103 same case :)

Nice catch :+1: I added that.

PR #2059, opened 2026-05-06
[Feat] Add Layout model LW-DETR

felixdittrich92:
This PR:

- Add LW-DETR layout detection model
- Add corresponding tests

PR #2063, opened 2026-05-07
[Feat] Add Vit-Det as LW-DETR encoder

felixdittrich92:
This PR:

- Add Vit-Det as classification model (encoder for LW-DETR)
- Add corresponding tests
- Add docs entry
- Add preps for layout models (onnx export / device and dtype set)
- Add ChannelLayerNorm which will be required by LW-DETR


felixdittrich92:
Related to #2059

PR #2076, opened 2026-06-17
Do not use root logger

lachesis:
Calling logging.info() at module level with no logger configured triggers logging.basicConfig() automatically (Python's "last resort" behavior) — which creates a StreamHandler on root with WARNING level.

This poisons logging config for anything downstream of this.

Instead, create a logger scoped to doctr package.

Example repro script:

```python
#!/usr/bin/env python
from doctr.io import Document

import logging
logging.basicConfig(level=logging.INFO)

logging.info("I do not see this")
logging.warning("I still see this")
```

Output:

```text
WARNING:root:I still see this
```

Expected:

```text
INFO:root:I do not see this
WARNING:root:I still see this
```

### DIFF
--- a/doctr/file_utils.py
+++ b/doctr/file_utils.py
@@ -22,7 +22,7 @@ def requires_package(name: str, extra_message: str | None = None) -> None:  # pr
     """
     try:
         _pkg_version = importlib.metadata.version(name)
-        logging.info(f"{name} version {_pkg_version} available.")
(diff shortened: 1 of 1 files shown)


lachesis:
Oh I see this anti-pattern is pretty common in doctr. It should really be fixed in all of the affected files. Maybe I'll open a more comprehensive PR later.

PR #2108, opened 2026-07-08
[feature] Add preserve_original_coords option for `straighten_pages=True`

### LINKED ISSUE #2107: `straighten_pages=True` returns bounding boxes in straightened-page coordinates, making them unusable for redaction or annotation on the original document
## Bug description

When `straighten_pages=True` is enabled, docTR internally deskews each page before running detection. This improves OCR accuracy, but the bounding boxes it returns are relative to the straightened page, which is an internal image the user never sees.

If you only care about extracting text, this doesn't matter. But if you need the boxes to line up with your original input image (for redaction, annotation, highlighting, or drawing overlays), they are misaligned by the full skew angle and basically unusable.

Right now there is no way to get the boxes back into original-image coordinates. The straightening transform (the rotation angle, the padding, the crop offset) is computed inside `_straighten_pages()` and then thrown away. It only ever exists as local variables.

Image

What you see: the boxes render horizontally while the text runs at -12 degrees. The average word y-span comes out around 16px, when the skewed words actually occupy around 40px vertically. See the attached image, the misalignment is obvious at a glance.

## Root cause

I traced the coordinate chain and there are three separate transforms between original and straightened space. All three lose their parameters:

1. `_straighten_pages()` in `models/predictor/base.py` calls `remove_image_padding(rotate_image(page, angle, expand=...))`. The angle, the expansion padding, and the crop offset are all computed but never stored anywhere.

2. `rotate_image(expand=True)` in `utils/geometry.py` pads the image to the expanded shape, rotates around the padded image center (not the original center), and then adds another aspect-ratio padding layer. So the real transform is a composition of translate, rotate, and translate again, and none of it is exposed to the caller.

3. After straightening, the predictor overwrites the page shapes (`origin_page_shapes = [page.shape[:2] for page in pages]` in `predictor/pytorch.py`), so even the original page dimensions are lost by the time the document builder runs.

There is also a trap for anyone who tries to work around this from outside the library: if you remap boxes back to skewed coordinates before document building, `DocumentBuilder._sort_boxes()` in `models/builder.py` re-estimates the page angle from the boxes and rotates them straight again, silently undoing your correction. (Unrelated to this issue, `_sort_boxes` also hardcodes `orig_shape=(1024, 1024)`, which distorts the rotation for non-square pages. I'll file that separately.)

## Proposed solution

Store the straightening transform as one composite affine matrix per page (pad, then rotate, then crop, built from the actual `cv2.getRotationMatrix2D` matrix plus an analytically computed crop offset). Then, optionally, remap the `Word` geometries back to original-page space after document building. Doing it after `_sort_boxes` means the two never conflict, and `builder.py` stays untouched. The whole thing sits behind a new opt-in flag (`preserve_original_coords=False` by default), so existing behavior does not change at all.

PR incoming. I kept the implementation deliberately minimal, but I'm happy to restructure it if you'd prefer a different shape. 

```python
import numpy as np, cv2

# Create a test document skewed by -12 degrees
img = np.ones((800, 600, 3), dtype=np.uint8) * 255
for i, line in enumerate(["This is a test document", "Sensitive information here",
                          "CONFIDENTIAL - DO NOT DISTRIBUTE", "Redaction required"]):
    cv2.putText(img, line, (50, 200 + i * 100), cv2.FONT_HERSHEY_SIMPLEX, 0.8, (0, 0, 0), 2)
M = cv2.getRotationMatrix2D((300, 400), -12, 1.0)
skewed = cv2.warpAffine(img, M, (600, 800), borderValue=(255, 255, 255))
img_rgb = cv2.cvtColor(skewed, cv2.COLOR_BGR2RGB)

model = ocr_predictor("db_resnet50", "crnn_vgg16_bn", pretrained=True,
                      assume_straight_pages=False, straighten_pages=True,
                      detect_orientation=True)
result = model([img_rgb])

# Draw the returned boxes on the ORIGINAL (skewed) input
vis = img_rgb.copy()
for block in result.pages[0].blocks:
    for line in block.lines:
        for word in line.words:
            pts = np.round(np.array(word.geometry).reshape(-1, 2) * np.array([600, 800])).astype(np.int32)
            cv2.polylines(vis, [pts], True, (255, 0, 0), 2)
cv2.imwrite("misaligned.png", cv2.cvtColor(vis, cv2.COLOR_RGB2BGR))

# Numerical check: the boxes come back flat, but the text is skewed
spans = [np.ptp(np.array(w.geometry).reshape(-1, 2)[:, 1]) * 800
         for b in result.pages[0].blocks for l in b.lines for w in l.words]
print(f"avg word y-span: {np.mean(spans):.1f}px")
print(f"expected for words skewed at -12 degrees: roughly 40px")
```

### Error traceback

No traceback. The code runs without errors, but the returned coordinates are silently wrong.

Output of the reproduction script:

avg word y-span: 16.5px
expected for words skewed at -12 degrees: roughly 40px

The boxes are returned in the internally straightened page's coordinate space instead of the
original input image's space, so drawing them on the original image produces misaligned boxes
(see attached misaligned.png). There is no error or warning indicating this.

### Environment

DocTR version: 1.0.2a0
PyTorch version: 2.12.1+cu130 (torchvision 0.27.1+cu130)
OpenCV version: 4.13.0
OS: Ubuntu 24.04.4 LTS
Python version: 3.12.3
Is CUDA available (PyTorch): No
CUDA runtime version: Could not collect
GPU models and configuration: Could not collect
Nvidia driver version: Could not collect
cuDNN version: Could not collect

saad-rd11:
Closes #2107

side_by_side

*Left: skewed input. Middle: current behavior with `straighten_pages=True`, boxes come back flat in the internally straightened page's coordinate space, misaligned with the input. Right: with `preserve_original_coords=True`, boxes are mapped back to the input image, usable for redaction, annotation, and overlays.*

## Summary

When `straighten_pages=True`, detection runs on internally deskewed pages, and the returned `Word.geometry` is relative to that straightened image, a coordinate space the user never sees. For text extraction this doesn't matter, but for anything that needs the boxes to line up with the input image (redaction, annotation, highlighting), they are unusable, and the straightening transform is discarded inside `_straighten_pages()` so there is no way to recover it.

This PR adds an opt-in flag, `preserve_original_coords` (default `False`, zero behavior change when off), that remaps word geometries back to the coordinate space of the image the user passed in.

## How it works

Two files changed plus one new test file. `builder.py` untouched.

**`models/predictor/base.py`**: When `preserve_original_coords=False` (default), `_straighten_pages()` runs the exact original `remove_image_padding(rotate_image(...))` path, verified byte-identical to main's output across portrait/landscape pages at ±12°. When the flag is on, an analytic pad → rotate → crop path runs instead, recording the straightening transform as one composite affine matrix per page (`inv(C @ R @ P)`, built from the actual `cv2.getRotationMatrix2D` matrix). The flag-on path needs its own analytic crop because a content-dependent pixel scan can't be inverted exactly.

**`models/predictor/pytorch.py`**: original page shapes are captured before straightening. After `DocumentBuilder` returns the `Document`, an optional post-processing pass converts each word polygon to absolute straightened-page pixels, applies the stored inverse matrix, clips to the original page bounds, and renormalizes.

The remap handles both geometry formats: 4-point polygons (`assume_straight_pages=False`) and 2-point boxes (`assume_straight_pages=True`). In the 2-point case the box is expanded to all 4 corners before the transform (rotating only the two stored diagonal points would give a wrong envelope) and returned as the axis-aligned envelope of the rotated corners. The envelope is a conservative superset of the true rotated region, which is the right behavior for redaction, and it preserves the 2-point format contract that downstream consumers expect.

Remapping after document building is deliberate. `DocumentBuilder._sort_boxes()` re-estimates the page angle from box geometry, so boxes remapped any earlier are detected as skewed and silently rotated straight again, undoing the correction. I verified this failure mode directly before settling on the post-builder approach. Doing the remap last means the detection, recognition, and building pipeline runs completely unmodified and the two mechanisms never interact.

## Validation

Three independent checks, ordered from pure math to full pipeline. The test-based checks are included as pytest cases in this PR (`tests/pytorch/test_preserve_original_coords.py`). Full suite: 19 passed in 45s.

**1. Analytic round-trip.** Points pushed through the forward matrix `C @ R @ P` and back through the stored inverse recover to 2.8e-13 px. The inverse is exact by construction; this check confirms no composition-order or convention error.

**2. Real-pixel fiducial test** (`test_straighten_inverse_fiducial`, 14 parametrized cases, no model weights, runs in about 2 seconds). Colored 3x3 dots at known positions go through the same pad, warpAffine, and crop path as `_straighten_pages`, are located in the output by exact color match, and remapped through the stored matrix. This test makes no assumptions about OpenCV's matrix conventions; it measures where real pixels actually land, which is what caught two convention bugs during development. Result: max error 0.49 px across angles of plus and minus 5 and 12 degrees plus 103, 193, and 283 degrees (covering the 90/180/270 base-orientation compositions with fine skew), on both portrait and landscape pages. Sub-degree angles are excluded with a comment in the test: at those rotations interpolation blends every fiducial pixel, so exact-color matching finds nothing to measure.

**3. End-to-end tests** (pretrained `db_resnet50` + `crnn_vgg16_bn`). Text is rendered at known positions, ground-truth word boxes are measured from the ink pixels of the clean render, the page is skewed by plus or minus 12 degrees, and the full predictor runs with `preserve_original_coords=True`. The GT boxes are transformed into the skewed frame (the frame the flag returns coordinates in) and compared against the remapped detections by IoU.

- `test_preserve_original_coords_roundtrip` (4 cases, `assume_straight_pages=False`, module-scoped predictor): mean IoU 0.894 to 0.910 across both page shapes and both skew signs, against a 0.4 threshold.
- `test_preserve_original_coords_2point` (1 case, `assume_straight_pages=True`): asserts the returned geometry stays 2-point and the envelope clears the same IoU threshold, exercising the 2-point expansion path end to end.

The remaining gap to IoU 1.0 is the detection model's own box localization, not the transform: check 2 bounds the transform's contribution at under half a pixel.

## Notes for reviewers

I kept this deliberately minimal and non-invasive, but I'm happy to restructure toward whichever shape fits the codebase better. Two alternatives I considered:

- Extending `rotate_image()` and `remove_image_padding()` in `utils/geometry.py` to optionally return their transform matrices, so the transform logic has a single source of truth there instead of being partially inlined in `_straighten_pages()`. More invasive, but removes the duplicated pad, rotate, and crop logic this PR currently carries.
- Exposing the matrix as `Page`-level metadata instead of rewriting `word.geometry` in place, leaving the remap to users. Cleaner separation, but less useful out of the box for the redaction use case that motivated this.

The validation suite transfers unchanged to either variant.

Separately, while tracing the coordinate chain I found that `_sort_boxes()` hardcodes `orig_shape=(1024, 1024)`, which distorts the reading-order rotation for non-square pages when `assume_straight_pages=False`. It doesn't affect this PR since the remap happens after the builder, so I'll file it as its own issue rather than mixing it in here.


saad-rd11:
Pushed c4f27f4: the analytic crop is now gated behind the flag. With `preserve_original_coords=False`, `_straighten_pages()` takes the exact original `remove_image_padding(rotate_image(...))` path — I verified this empirically by running both branches' `_straighten_pages` on identical inputs (portrait and landscape pages, ±12°, text near edges): outputs are byte-identical (`np.array_equal` per page). The analytic path only activates when the flag is on, since the matrix remap needs a deterministic, exactly invertible crop.

doctr/models/predictor/base.py
```diff
@@ -106,11 +117,55 @@ def _straighten_pages(
...
-            remove_image_padding(rotate_image(page, angle, expand=page.shape[0] != page.shape[1]))
-            for page, angle in zip(pages, origin_pages_orientations)
-        ]
+        if not self.preserve_original_coords:
+            return [
+                remove_image_padding(rotate_image(page, angle, expand=page.shape[0] != page.shape[1]))
+                for page, angle in zip(pages, origin_pages_orientations)
+            ]
+
+        self._straighten_m_inv = []
```
felixdittrich92: This should be outsourced to utils/geometry.py and only run if straighten_pages & preserve_original_coords == True (standalone test in test_utils_geometry.py) otherwise keep the current behaviour

doctr/models/predictor/pytorch.py
```diff
@@ -187,6 +192,34 @@ def forward(
             regions,
             tables,
         )
+
+        if self.straighten_pages and self.preserve_original_coords and self._straighten_m_inv:
```
felixdittrich92: This logic should be moved to _OCRPredictor as method where KIEPredictor also inerhits from - so same logic needs to be applied in kie_predictor/pytorch.py

doctr/models/predictor/pytorch.py
```diff
@@ -38,6 +38,8 @@ class OCRPredictor(nn.Module, _OCRPredictor):
             page. Doing so will slightly deteriorate the overall latency.
         detect_language: if True, the language prediction will be added to the predictions for each
             page. Doing so will slightly deteriorate the overall latency.
+        preserve_original_coords: if True and straighten_pages is True, the returned bounding boxes are mapped back
```
felixdittrich92, line 41: passing as kwarg in ocr_predictor is fine here ftm 👍

doctr/models/predictor/pytorch.py
```diff
@@ -118,9 +122,10 @@ def forward(
             general_pages_orientations = None
             origin_pages_orientations = None
         if self.straighten_pages:
+            _orig_shapes = [p.shape[:2] for p in pages]
```
felixdittrich92:
```python
_orig_shapes = origin_page_shapes
pages = self._straighten_pages(pages, seg_maps, general_pages_orientations, origin_pages_orientations)
 # update page shapes after straightening
 origin_page_shapes = [page.shape[:2] for page in pages]
```

doctr/models/predictor/pytorch.py
felixdittrich92: looks like the `and self._straighten_m_inv` check can be dropped should always exist if self.straighten_pages and self.preserve_original_coords

felixdittrich92 (changes requested):
Hi @saad-rd11 👋 

Thanks for the PR
I left some initial comments 

Additional we should verify that this combination works with `.show()` afterwards correct

felixdittrich92:
> Hi @saad-rd11 👋
> 
> Thanks for the PR I left some initial comments
> 
> Additional we should verify that this combination works with `.show()` afterwards correct

Currently this will still forward the "rotation corrected pages" here: pages = self._straighten_pages(pages, seg_maps, general_pages_orientations, origin_pages_orientations)

with your change you also need to override this to the original input images which then are passed to the builder if both straigthen_pages & preserve_original_coords are True

saad-rd11:
@felixdittrich92 Thanks for the review! 

All makes sense. I'll move the transform to geometry.py with a standalone test, lift the remap into _OCRPredictor so KIE inherits it, apply the simplifications, and verify .show() renders correctly with the flag on. 

Will have it pushed shortly.

saad-rd11:
Thanks for the review @felixdittrich92 all six comments addressed, commits linked on each thread. Rebased onto current main. I found two things during implementation worth mentioning:

KIE had no remap path at all. Moving the logic into _OCRPredictor (per your comment) meant it needed to work for KIE too, but KIEPage stores geometry in a predictions dict rather than the OCR blocks → lines → words tree. Handled it with duck-typing in the shared _remap_to_original_coords method (hasattr(page, "predictions")), and added a smoke test that runs KIE flag-on vs flag-off and asserts the geometries actually differ (L2 ~0.72–1.34), since a structural mismatch there would silently no-op rather than raise.

Verifying .show() caught a real bug as page.dimensions was left at the straightened value after the coordinate swap, so visualize_page scaled boxes by the wrong dimensions (~1.26×, shifting every box down-left). hOCR/XML export would have hit the same. Fixed by restoring dimensions alongside page.page, with an assertion in the roundtrip test. Before/after can been seen down this thread.

Rebase note: upstream dfb8f7c added ignore_regions as a new positional param to _OCRPredictor.__init__, which collided with the positional preserve_original_coords pass fixed with keyword passing.

Full suite green: 20 feature tests + 17 zoo, ruff clean, mypy unchanged from baseline. preserve_original_coords=False (default) is a no-op path, byte-identical to current behavior.

doctr/models/predictor/base.py
saad-rd11: Done in 7fd11bc moved to straighten_page() in geometry.py. Fiducial test moved to tests/common/test_utils_geometry.py.

doctr/models/predictor/pytorch.py
saad-rd11: Done in 16f6276 _remap_to_original_coords() on _OCRPredictor, called from both OCR and KIE forward(). See the top-level note re: KIE storing geometry in a predictions dict rather than blocks/lines/words.

doctr/models/predictor/pytorch.py
saad-rd11: Done in 47558c5 dropped the redundant guard and applied your origin_page_shapes capture pattern.

doctr/models/predictor/pytorch.py
saad-rd11: > looks like the and self._straighten_m_inv check can be dropped should always exist if self.straighten_pages and self.preserve_original_coords

Done in 47558c5

doctr/models/predictor/pytorch.py
saad-rd11: Thanks, left it as is.

saad-rd11:
> > Hi @saad-rd11 👋
> > Thanks for the PR I left some initial comments
> > Additional we should verify that this combination works with `.show()` afterwards correct
> 
> Currently this will still forward the "rotation corrected pages" here: pages = self._straighten_pages(pages, seg_maps, general_pages_orientations, origin_pages_orientations)
> 
> with your change you also need to override this to the original input images which then are passed to the builder if both straigthen_pages & preserve_original_coords are True

Verified renders correctly with the flag on (image below), via docTR's visualize_page path. 

image


This also surfaced a stale page.dimensions bug: after the coordinate swap, page.dimensions still held the straightened size, so visualize_page scaled boxes ~1.26× (down-left shift). Fixed in 12c9918, restored alongside page.page with a roundtrip assertion. hOCR/XML export would have hit the same.

BEFORE:

image

AFTER:

image

saad-rd11:
Hi @felixdittrich92 quick note on the remaining Codacy flag: it's now raising D213 on the same docstring where it previously raised D212. These two rules are mutually exclusive in pydocstyle D212 requires the summary on the first line, D213 on the second so satisfying one always triggers the other.

doctr/models/predictor/base.py
```diff
@@ -212,6 +231,78 @@ def _process_predictions(
...
+        -------
+        the document with remapped word geometries
+
+        """
+        for pidx, (page, m_inv) in enumerate(zip(out.pages, self._straighten_m_inv)):
+            sh, sw = straight_shapes[pidx]
+            oh, ow = orig_shapes[pidx]
+
+            # Collect all geometry-bearing objects (supports Page & KIEPage)
+            if hasattr(page, "predictions"):
```
felixdittrich92: `if isinstance(page, KIEPage) nd hasattr(page, "predictions"):`

doctr/models/predictor/base.py
```diff
@@ -212,6 +231,78 @@ def _process_predictions(
...
+        straight_shapes: list[tuple[int, int]],
+        orig_pages: list[np.ndarray] | None = None,
+    ) -> Document:
+        """Remap word geometries from straightened-page coordinates back to original page coordinates.
+
+        This method is called after the document builder when straighten_pages and
+        preserve_original_coords are both True. It applies the inverse straightening
+        transform to each word's geometry so that bounding boxes align with the input image.
+
+        Supports both Page (blocks -> lines -> words) and KIEPage (predictions dict)
```
felixdittrich92: Replace all `` with single ` for all the added docstrings / comments

doctr/models/predictor/base.py
```diff
@@ -212,6 +231,78 @@ def _process_predictions(
...
+        (pre-straightening) input so that ``.show()`` renders boxes on the correct image.
+
+        Args:
+            out: the document returned by the builder
+            orig_shapes: original (pre-straightening) page shapes (H, W)
+            straight_shapes: straightened page shapes (H, W)
+            orig_pages: optional list of original page images to restore on ``page.page``
+
+        Returns
+        -------
```
felixdittrich92:
```
Returns:
     the document with remapped word geometries
```

doctr/models/kie_predictor/pytorch.py
```diff
@@ -192,6 +199,9 @@ def forward(
             languages_dict,
             regions,
         )
+
```
felixdittrich92, line 204: A comment would be good 
```
# manipulate the already built Document to restore the original pages / shapes and geometries
```

Additional self._remap_to_original_coords must also handle the geometry from the Layout Elements and table elements not only the word geometry

doctr/utils/geometry.py
```diff
@@ -398,6 +399,71 @@ def remove_image_padding(image: np.ndarray) -> np.ndarray:
...
+        page: the original page image (H, W, C)
+        angle: rotation angle in degrees (between -90 and +90)
+
+    Returns:
+        a tuple (straightened, M_inv) where straightened is the page rotated and cropped to the
+        non-padded region, and M_inv is a (3, 3) array mapping points in the straightened image's
+        pixel space (x, y, 1) back to the original page's pixel space.
+    """
+    h, w = page.shape[:2]
+    expand = h != w
```
felixdittrich92: Square pages (h == w) will break the analytic crop.

doctr/models/predictor/base.py
```diff
@@ -134,11 +145,19 @@ def _straighten_pages(
...
-            remove_image_padding(rotate_image(page, angle, expand=page.shape[0] != page.shape[1]))
-            for page, angle in zip(pages, origin_pages_orientations)
-        ]
+        if not self.preserve_original_coords:
+            return [
+                remove_image_padding(rotate_image(page, angle, expand=page.shape[0] != page.shape[1]))
+                for page, angle in zip(pages, origin_pages_orientations)
+            ]
+
+        self._straighten_m_inv = []
```
felixdittrich92: This makes the predictor stateful we should avoid saving it in self

tests/common/test_utils_geometry.py
```diff
@@ -326,3 +326,38 @@ def test_extract_rcrops(mock_pdf, assume_horizontal):
 
     # No box
     assert geometry.extract_rcrops(doc_img, np.zeros((0, 4, 2)), assume_horizontal=assume_horizontal) == []
+
+
+@pytest.mark.parametrize("angle", [5, 12, -5, -12, 90 + 13, 180 + 13, 270 + 13])
+@pytest.mark.parametrize("shape", [(800, 600), (600, 800)])
+def test_straighten_page_inverse(angle, shape):
+    """M_inv returned by straighten_page is the exact inverse of the
```
felixdittrich92: Remove docstrings after the test function names - it's enough to add small comments at places where it's not clear / required

tests/pytorch/test_preserve_original_coords.py
felixdittrich92, line 1: This file should be included in tests/pytorch/test_models_zoo_pt.py extend the existing tests parameterization should be enough / maybe a geometry check afterwards but not a standalone test file :)

felixdittrich92 (changes requested):

saad-rd11:
@felixdittrich92 Sure! I will apply the style items, remove the predictor state, fold the e2e tests into the zoo parametrization, handle square pages (will add a square case to the fiducial test first), and extend the remap to Layout/Table element geometry. Should have it pushed in a day or two.

doctr/models/predictor/base.py
saad-rd11: Fixed in 2ad882a changed to isinstance(page, KIEPage) with the import added to base.py. Verified no circular import: elements.py only imports from doctr.file_utils, doctr.utils.*, and top-level doctr, none of which import from the predictor modules.

doctr/models/predictor/base.py
saad-rd11: Fixed in fcc3273: all double backticks replaced with single across the added docstrings and comments.

doctr/models/predictor/base.py
saad-rd11: Fixed in fcc3273: changed to Returns: with indented body, matching the file's existing convention.

doctr/models/kie_predictor/pytorch.py
saad-rd11: Comment added verbatim in 46e6d56.

The layout/table/artefact geometry handling was addressed in 19869a8: extracted a _remap_geometry static helper and extended _remap_to_original_coords to walk every geometry-bearing object — Word, Line, Block (envelope), Artefact, LayoutElement, Table, TableCell, and Prediction. Verified against elements.py: KIEPage takes layout but not tables (line 618), so the walk uses guarded access and skips tables gracefully on KIE pages; Page handles both.

Coverage in be3db15: a synthetic-Document test constructs one of every geometry type (no pretrained layout models needed), runs the real _remap_to_original_coords with a realistic 12° rotation m_inv, and asserts every geometry changed plus a containment assertion that remapped Block envelopes still bound their remapped Words.

doctr/utils/geometry.py
saad-rd11: Confirmed and fixed in 2ad882a. Reproduced first: for (700, 700) at 12°, the expand = h != w guard skipped padding entirely, content corners rotated outside the canvas, and the fiducial remap error was ~90px.

Fix: removed the h != w guard, compute_expanded_shape is now called unconditionally. At angle ≈ 0 it returns the original shape (zero pad), so near-zero rotations are unchanged. The aspect-ratio padding branch guard was updated to h_pad > 0 or w_pad > 0 accordingly. The flag-off path was not touched: the "False is byte-identical" guarantee holds.

Square shape (700, 700) added to the fiducial parametrization: all 21 cases (3 shapes × 7 angles, including the 90/180/270 base-orientation compositions) pass with max error < 0.6px. One note: the (700,700) × 193° case initially found zero fiducials, the 180°+13° composition's resampling blended the 3×3 dots below exact-color-match threshold. Resolved by using 5×5 dots throughout; all cases now pass with no skips.

doctr/models/predictor/base.py
saad-rd11: Fixed in 8f0c936: _straighten_pages now returns (pages, m_invs) instead of writing to self, both OCR and KIE forward() unpack the tuple and pass m_invs as a parameter to _remap_to_original_coords. The flag-off path returns an empty list to keep the return type uniform. No instance state remains: grep for _straighten_m_inv across doctr/ returns zero results.

tests/common/test_utils_geometry.py
saad-rd11: Fixed in 4cad1c9: test docstrings removed; the one non-obvious rationale (sub-degree angle exclusion) kept as an inline comment.

tests/pytorch/test_preserve_original_coords.py
saad-rd11: Done in 2d96c4d and 6bcbc59: standalone file deleted, coverage folded into test_models_zoo_pt.py following its existing patterns: module-scoped fixtures for the pretrained predictors, test_preserve_original_coords_roundtrip (4 cases, mean IoU > 0.4, page.page and page.dimensions assertions), test_preserve_original_coords_2point (assume_straight_pages=True variant), test_preserve_original_coords_kie_smoke (flag-on vs flag-off geometry difference), plus the _remap_geometry unit test and the synthetic-Document walk test. All 25 zoo tests pass locally (321s with cached weights). The fiducial inverse test stays in tests/common/test_utils_geometry.py.

saad-rd11:
Thanks for the second round @felixdittrich92 all eight comments addressed, replies with commits on each thread. The short version:

Regarding your square-page catch: I reproduced it first: (700, 700) at 12° skipped padding entirely because of the inherited h != w expand guard, content corners rotated off the canvas, and the remap error was around 90px. Fixed by removing the guard; compute_expanded_shape now decides unconditionally, which costs nothing at near-zero angles. Square shapes are now part of the fiducial test parametrization, all 21 cases pass under 0.6px.
The predictor is now stateless. _straighten_pages returns the inverse matrices instead of writing them to self, and they flow through forward() as locals into _remap_to_original_coords as a parameter. No instance state left grep for the old attribute comes back empty.

The remap now covers every geometry-bearing object, not just words: Line, Block, Artefact, LayoutElement, Table, TableCell, and Prediction, for both Page and KIEPage (checked against elements.py. KIEPage carries layout but not tables, so access is guarded). A synthetic-Document test builds one of every type and asserts each geometry actually changes after the remap, plus a containment check that remapped Block envelopes still bound their remapped Words. No pretrained layout models needed for that coverage.
Tests are folded into test_models_zoo_pt.py per your preference: module-scoped fixtures, the roundtrip/2-point/KIE cases, the walk test, and the _remap_geometry unit test. The standalone file is gone. The fiducial inverse test stays in test_utils_geometry.py. All 25 zoo tests pass locally, full suite is 61 green, ruff clean, mypy unchanged from baseline.


Style items (isinstance, backticks, Returns: format, the comment in kie_predictor, test docstrings) all applied as requested.
As before: preserve_original_coords=False is untouched: the default path stays byte-identical to main.

One note on Codacy: it now flags D406/D407 on the Returns: docstring, those rules enforce numpy-style formatting, i.e. they flag the exact format requested in review. Along with the earlier D213, all three notices come from Codacy's Prospector profile enforcing conventions that aren't in the repo's ruff config. Nothing actionable on the PR side as far as I can tell.

doctr/models/predictor/base.py
```diff
@@ -134,11 +145,19 @@ def _straighten_pages(
                 for seq_map, general_orientation in zip(seg_maps, general_pages_orientations)
             ]
         )
-        return [
-            # expand if height and width are not equal, then remove the padding
-            remove_image_padding(rotate_image(page, angle, expand=page.shape[0] != page.shape[1]))
-            for page, angle in zip(pages, origin_pages_orientations)
-        ]
+        if not self.preserve_original_coords:
+            return [
```
felixdittrich92, line 150: comment like before should be keept:

```
# expand if height and width are not equal, then remove the padding
remove_image_padding(rotate_image(page, angle, expand=page.shape[0] != page.shape[1]))
```

tests/pytorch/test_models_zoo_pt.py
felixdittrich92, line 1: Move all imports to the top of the file where the other imports are placed

felixdittrich92, line 1: two tests should be enough here:

- The two "smoke" tests (one for ocr_predictor / one for kie_predictor)  - `test_ocr_predictor_straighten_with_preserve_original_coords` and `test_kie_predictor_straighten_with_preserve_original_coords`
- As a test file we could use the mock_payslip fixture (https://github.com/mindee/doctr/blob/1a1018c0c3992f2113d4fdb3c752b99621997a44/tests/conftest.py#L46)

doctr/models/predictor/base.py
```diff
@@ -134,11 +145,19 @@ def _straighten_pages(
                 for seq_map, general_orientation in zip(seg_maps, general_pages_orientations)
             ]
         )
-        return [
-            # expand if height and width are not equal, then remove the padding
-            remove_image_padding(rotate_image(page, angle, expand=page.shape[0] != page.shape[1]))
-            for page, angle in zip(pages, origin_pages_orientations)
-        ]
+        if not self.preserve_original_coords:
```
felixdittrich92, line 148: I was still thinking about a possible way to avoid the duplicated logic now in straighten_page and remove_image_padding(rotate_image(..)) but let's keep it as is ftm 👍

doctr/models/predictor/base.py
```diff
@@ -212,6 +231,81 @@ def _process_predictions(
...
+            straight_shapes: straightened page shapes (H, W)
+            m_invs: inverse affine matrices from the straightening pipeline, one per page
+            orig_pages: optional list of original page images to restore on `page.page`
+
+        Returns:
+            the document with remapped word geometries
+
+        """
+        for pidx, (page, m_inv) in enumerate(zip(out.pages, m_invs)):
+            sh, sw = straight_shapes[pidx]
```
felixdittrich92: comment 

```
# straightened_height, straightened_width
sh, sw =
# original_height, original_width
oh, ow = 
```

or rename variables original_height, original_width, straightened_height, straightened_width

felixdittrich92 (changes requested):
Thanks again @saad-rd11 👍 

One last round and we should be good to merge 👍

saad-rd11:
Thanks @felixdittrich92! I am on it, all four are quick, will push today. 👍

doctr/models/predictor/base.py
saad-rd11: Restored verbatim in e582b32.

tests/pytorch/test_models_zoo_pt.py
saad-rd11: Resolved in 24d121e; the functions carrying local imports were removed in the test consolidation, so no function-level imports remain in the file.

saad-rd11: Done in 24d121e: standalone tests and both module-scoped fixtures removed, replaced with test_ocr_predictor_straighten_with_preserve_original_coords (mock_payslip: values, reference geometry per the test_trained_ocr_predictor pattern, page/dimensions restoration) and test_kie_predictor_straighten_with_preserve_original_coords. For the KIE case I used mock_tilted_payslip so the remap is exercised on genuine skew end to end — and that fixture immediately paid off: its real-world dimensions (2291×2025 at 30°) surfaced a floating-point bug in the analytic crop that the synthetic fiducial shapes never triggered. A content corner projected to x = −0.033, floor gave −1, and Python's negative slice rotated[:, -1:] silently collapsed the crop to a 1-pixel strip. Fixed in 6f3f737 with a max(0, ...) clamp on cx/cy, and the failing shape/angle is pinned in the fiducial parametrization (now 22 cases, max error < 0.6px). Both smoke tests pass locally (~25s KIE, comparable to test_trained_kie_predictor).

doctr/models/predictor/base.py
saad-rd11: 👍

doctr/models/predictor/base.py
saad-rd11: Added as inline comments in e582b32 (# straightened_height, straightened_width / # original_height, original_width).

saad-rd11:
Thanks @felixdittrich92: all items addressed, latest round pushed (three commits):

e582b32: restore flag-off path comment + variable name clarification
6f3f737: fix analytic crop: clamp cx/cy non-negative (floating-point corner projection; details on the thread) + regression case in the fiducial parametrization
24d121e: tests consolidated to the two smoke tests (−310/+54)

Test suite: 59 green locally (22 fiducial inverse cases < 0.6px, both new smoke tests passing), ruff clean, mypy unchanged from baseline. As throughout: preserve_original_coords=False remains byte-identical to main.

On Codacy: same situation as before, the 3 notices are D213/D406/D407 from its Prospector profile, which enforce docstring conventions contradicting the repo's requested format. Nothing actionable that I can see.

tests/pytorch/test_models_zoo_pt.py
```diff
@@ -659,3 +659,76 @@ def test_end_to_end_torch_compile(det_arch, reco_arch, mock_payslip):
         word.value == compiled_out.pages[0].blocks[0].lines[0].words[i].value
         for i, word in enumerate(out.pages[0].blocks[0].lines[0].words)
     )
+
+
+def test_ocr_predictor_straighten_with_preserve_original_coords(mock_payslip):
```
felixdittrich92: One last thing:

This test should also compare between predictor_on & predictor_off like it's done for the kie_predictor test case.

felixdittrich92 (changes requested):
Last change and we can merge 👍

tests/common/test_utils_geometry.py
```diff
@@ -326,3 +326,64 @@ def test_extract_rcrops(mock_pdf, assume_horizontal):
...
+        recovered = (np.array([float(fx), float(fy), 1.0]) @ m_inv.T)[:2]
+        errors.append(np.linalg.norm(recovered - np.array([float(dx), float(dy)])))
+
+    if len(errors) == 0:
+        return
+    assert max(errors) < 0.6, f"Max remap error {max(errors):.4f}px exceeds 0.6px threshold"
+
+
+def test_straighten_page_projected_corner_clamp():
+    """Regression: large images at 30° can project a content corner to cx=-1 via
```
felixdittrich92: Remove docstring here

tests/pytorch/test_models_zoo_pt.py
saad-rd11: Done in d87dc657: the OCR test now mirrors the KIE case: predictor_on vs predictor_off on mock_tilted_payslip, asserting the geometries differ, "Mr." is recognized on the deskewed content, and page/dimensions restore to the original input shape. Both smoke tests now exercise a genuine non-identity remap: OCR through the blocks tree, KIE through the predictions dict.

tests/common/test_utils_geometry.py
saad-rd11: Removed in d87dc657: the non-obvious rationale kept as a one-line inline comment.

felixdittrich92 (approved):
Thanks @saad-rd11 for all the iterations looks good now 👍 
After CI is green I will merge it

saad-rd11:
Thanks @felixdittrich92. I learned a lot from this.

PR #2133, opened 2026-08-27
feat(io): estimate x_size and x_fsize in hOCR export

WSHAPER:
The hOCR export was emitting `x_size 0` on every `ocr_line` element (with an in-code note that the value was left uninitialized), so text height and font size were unspecified for consumers of the export.

This PR estimates both properties from the element box height:

- `x_size`: absolute text height in pixels, derived from the bbox height of the line, table row or KIE prediction
- `x_fsize`: font size in points, converted from `x_size` through a new `dpi` export argument (default 72, matching the hOCR point convention and docTR's PDF scale convention)

`baseline`, `x_descenders` and `x_ascenders` remain 0 since the pipeline has no signal for them.

## How was it tested?

New test covering lines, table rows, KIE predictions and the `dpi` scaling, plus the existing export suites: `pytest tests/common/test_io_elements.py tests/common/test_io_exporters.py` (55 passed). `ruff check`, `ruff format --check` and `mypy` are clean.

### DIFF
--- a/doctr/io/exporters.py
+++ b/doctr/io/exporters.py
@@ -640,6 +640,22 @@ def _hocr_bbox(geometry: BoundingBox, width: int, height: int) -> str:
     )
 
 
+def _hocr_text_size(geometry: BoundingBox, height: int, dpi: int = 72) -> tuple[int, int]:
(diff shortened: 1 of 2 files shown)


felixdittrich92 (approved):
Thanks @WSHAPER LGTM 👍

PR #2069, opened 2026-06-02
fix: handle empty inputs in OrientationPredictor

badranX:
`OrientationPredictor` does not handle empty input batches and can raise an exception when orientation detection is enabled and no valid crops/images are passed to the predictor. This PR handles empty inputs similar to the `RecognitionPredictor`.

### Reproduction

```python
import requests

from doctr.io import DocumentFile
from doctr.models import ocr_predictor

receipt = requests.get(
    "https://github.com/mindee/doctr/releases/download/v0.3.0/mock_receipt.jpeg"
).content

predictor = ocr_predictor(
    pretrained=True,
    det_arch="fast_base",
    reco_arch="parseq",
    assume_straight_pages=False,
    detect_orientation=True,
    disable_crop_orientation=False,
    disable_page_orientation=False,
    straighten_pages=True,
)

docs = DocumentFile.from_images([receipt])
docs[0] = docs[0][:10, :10]

result = predictor(docs)
```

Result:

```text
File "/home/ubuntu/dev/pr/doctr/doctr/models/preprocessor/pytorch.py", line 73, in batch_inputs
    if isinstance(samples[0], tuple):
                  ~~~~~~~^^^
IndexError: list index out of range
```

PR #2138, opened 2026-09-05
fix: handle empty batches in detection and end-to-end predictors

linhongyu510:
`DetectionPredictor`, `OCRPredictor` and `KIEPredictor` raise on an empty page list instead of returning an empty result. `RecognitionPredictor` has always short-circuited on empty input (`recognition/predictor/pytorch.py:50`), and `OrientationPredictor` gained the same behaviour in #2069 — these three were the remaining gap, so this follows that precedent rather than introducing a new convention.

Filtering a batch down to nothing is ordinary caller code (skip already-processed pages, drop files that failed a check upstream), and today it fails from internals that never mention the empty input.

### Reproduction

```python
from doctr.models import ocr_predictor, kie_predictor, detection

ocr_predictor(pretrained=False)([])       # IndexError: list index out of range
kie_predictor(pretrained=False)([])       # IndexError: list index out of range
detection.detection_predictor(pretrained=False)([])  # IndexError
```

There are three separate failure points on the way down, which is why the fix is not a one-liner in a single helper:

| Location | Failure |
|---|---|
| `preprocessor/pytorch.py:73` | `num_batches` is correctly `0`, then `samples[0]` is read to pick the tuple/tensor branch → `IndexError` |
| `utils/geometry.py:124` | `zip(*(_detach(box) for box in boxes))` over no boxes → `ValueError: not enough values to unpack (expected 2, got 0)` |
| `models/_utils.py:276` | KIE path only: `{k: ... for k in x[0]}` → `IndexError` |

I confirmed the ordering by fixing them one at a time: guarding `batch_inputs` moved the crash to `detach_scores`, and guarding that one let the whole call return `Document(pages=[])`.

### Fix

Guard at the three public entry points instead of patching each internal, so `batch_inputs`, `detach_scores` and `invert_data_structure` keep their non-empty precondition and stay simple.

Two details worth flagging:

- `DetectionPredictor` returns the shape its own `return_maps` contract promises: `[]` normally, `([], [])` when `return_maps=True`.
- `KIEPredictor` returns `KIEDocument(pages=[])`, not `Document(pages=[])`. `KIEDocument` subclasses `Document`, so returning the base class would type-check and pass an `isinstance` assertion while silently dropping the per-class page shape. The test asserts the exact type for this reason — I verified it catches the degradation by deliberately returning `Document` there and watching the test go red.

### Tests

Added `test_predictors_on_empty_batch`, which covers detection (both `return_maps` values), recognition (already-correct, asserted so the three stay consistent), and both end-to-end predictors.

The test is load-bearing: reverting the guards fails it at `preprocessor/pytorch.py:73`.

```
tests/pytorch/test_models_zoo_pt.py::test_predictors_on_empty_batch  1 passed
tests/pytorch/test_models_zoo_pt.py                                 14 passed, 6 errors
tests/common/                                                        523 passed, 1 failed, 7 errors
ruff check . / ruff format --check                                   clean
mypy doctr/                                                          clean (171 source files)
```

The errors and the one failure are pre-existing on this machine and unrelated: the errors are `requests.exceptions.ConnectionError` from tests that download weights or fixtures, and `tests/common/test_io.py::test_read_html` fails with `OSError: cannot load library` (missing WeasyPrint system libs). I verified the `test_read_html` failure reproduces on a clean checkout via `git stash`.

AI assistance was used for this change; I reviewed every changed line and ran the commands above locally.

### DIFF
--- a/tests/pytorch/test_models_zoo_pt.py
+++ b/tests/pytorch/test_models_zoo_pt.py
@@ -124,6 +124,47 @@ def test_ocrpredictor(
     assert out.pages[0].orientation["value"] == orientation
 
 
+def test_predictors_on_empty_batch(mock_vocab):
+    """An empty page list must yield an empty Document instead of raising.
+
+    Filtering a batch down to nothing is ordinary caller code, and
+    `RecognitionPredictor` (and `OrientationPredictor` since #2069) already
+    return empty results for it. The detection and end-to-end predictors did
+    not, so they crashed deep in the stack -- `IndexError` from `samples[0]` in
+    `PreProcessor.batch_inputs`, then `ValueError` from `zip(*...)` in
+    `detach_scores`, then `IndexError` from `x[0]` in `invert_data_structure`
+    on the KIE path -- none of which names the empty input.
+    """
+    det_predictor = DetectionPredictor(
+        PreProcessor(output_size=(512, 512), batch_size=2),
+        detection.db_mobilenet_v3_large(pretrained=False, pretrained_backbone=False, assume_straight_pages=True),
+    )
+    reco_predictor = RecognitionPredictor(
+        PreProcessor(output_size=(32, 128), batch_size=32, preserve_aspect_ratio=True),
+        recognition.crnn_vgg16_bn(pretrained=False, pretrained_backbone=False, vocab=mock_vocab),
+    )
+
+    # Detection keeps the shape its `return_maps` contract promises.
+    assert det_predictor([]) == []
+    assert det_predictor([], return_maps=True) == ([], [])
+
+    # The recognition predictor already behaved; asserted here so the three
+    # stay consistent if one of them is touched again.
+    assert reco_predictor([]) == []
+
+    for predictor, expected_type in (
+        (OCRPredictor(det_predictor, reco_predictor, assume_straight_pages=True), Document),
+        (KIEPredictor(det_predictor, reco_predictor, assume_straight_pages=True), KIEDocument),
+    ):
+        out = predictor([])
+        # Exact type, not isinstance: KIEDocument subclasses Document, so an
+        # isinstance check would not notice the KIE path degrading to the base
+        # class and dropping the per-class prediction shape.
+        assert type(out) is expected_type
+        assert out.pages == []
+        assert out.export() == {"pages": []}
+
+
 def test_ocrpredictor_layout(mock_pdf, mock_vocab, mock_payslip):
     det_predictor = DetectionPredictor(
         PreProcessor(output_size=(512, 512), batch_size=2),
--- a/doctr/models/detection/predictor/pytorch.py
+++ b/doctr/models/detection/predictor/pytorch.py
@@ -40,6 +40,9 @@ class DetectionPredictor(nn.Module):
         return_maps: bool = False,
         **kwargs: Any,
     ) -> list[dict[str, np.ndarray]] | tuple[list[dict[str, np.ndarray]], list[np.ndarray]]:
+        if len(pages) == 0:
+            return ([], []) if return_maps else []
+
         # Extract parameters from the preprocessor
         preserve_aspect_ratio = self.pre_processor.resize.preserve_aspect_ratio
         symmetric_pad = self.pre_processor.resize.symmetric_pad
--- a/doctr/models/kie_predictor/pytorch.py
+++ b/doctr/models/kie_predictor/pytorch.py
@@ -9,7 +9,7 @@ import numpy as np
 import torch
 from torch import nn
 
-from doctr.io.elements import Document
+from doctr.io.elements import Document, KIEDocument
 from doctr.models._utils import get_language, invert_data_structure
 from doctr.models.detection.predictor import DetectionPredictor
 from doctr.models.layout.predictor import LayoutPredictor
@@ -79,6 +79,9 @@ class KIEPredictor(nn.Module, _KIEPredictor):
         pages: list[np.ndarray],
         **kwargs: Any,
     ) -> Document:
+        if len(pages) == 0:
+            return KIEDocument(pages=[])
+
         # Dimension check
         if any(page.ndim != 3 for page in pages):
             raise ValueError("incorrect input shape: all pages are expected to be multi-channel 2D images.")
--- a/doctr/models/predictor/pytorch.py
+++ b/doctr/models/predictor/pytorch.py
@@ -93,6 +93,9 @@ class OCRPredictor(nn.Module, _OCRPredictor):
         pages: list[np.ndarray],
         **kwargs: Any,
     ) -> Document:
+        if len(pages) == 0:
+            return Document(pages=[])
+
         # Dimension check
         if any(page.ndim != 3 for page in pages):
             raise ValueError("incorrect input shape: all pages are expected to be multi-channel 2D images.")


tests/pytorch/test_models_zoo_pt.py
```diff
@@ -124,6 +124,47 @@ def test_ocrpredictor(
     assert out.pages[0].orientation["value"] == orientation
 
 
+def test_predictors_on_empty_batch(mock_vocab):
+    """An empty page list must yield an empty Document instead of raising.
+
+    Filtering a batch down to nothing is ordinary caller code, and
```
felixdittrich92: Please remove the docstring afterwards fine to merge 👍

felixdittrich92 (changes requested):
Hi @linhongyu510 👋,

Thanks for the PR. Left one comment.

linhongyu510:
Docstring removed in `05d0615` — thanks for catching it. I checked the convention before deleting rather than just complying: 3 of 371 test functions in `tests/` carry a docstring, and the other two are single-line, so the 10-line block here was the outlier. The load-bearing context was already in the inline comments (the `return_maps` shape contract and the reason for `type(out) is` over `isinstance`), so nothing was lost by dropping it.

Also rebased onto current `main` (`c7f7218`), which only touches `vocabs.py` / docs / pre-commit — no overlap with the four files here, and 0 commits behind now.

Re-verified after the rebase rather than assuming the earlier run still held:

```
pytest tests/pytorch/test_models_zoo_pt.py::test_predictors_on_empty_batch   1 passed
pytest tests/pytorch/test_models_zoo_pt.py                                  14 passed, 6 errors
ruff check / ruff format --check (4 changed files)                           clean
mypy doctr/                                                                 clean, 171 files
```

The 6 errors are fixture downloads timing out on this machine (`3.bp.blogspot.com` unreachable); I confirmed the attribution by `git stash`-ing the change and re-running — identical 14 passed / 6 errors with the same error set, so they are not from this diff.

I also re-confirmed the test still earns its place after the edit, by reverting each guard one at a time:

| reverted guard | failure |
|---|---|
| `DetectionPredictor` | `IndexError` at `preprocessor/pytorch.py:73` (`samples[0]`) |
| `OCRPredictor` | `ValueError: not enough values to unpack` at `utils/geometry.py:124` |
| `KIEPredictor` | `IndexError` at `models/_utils.py:276` (`x[0]`) |

Three distinct failure points, matching the three in the description. All guards restored afterwards.

linhongyu510:
One thing that needs your click: the nine workflows on `05d0615` are all sitting at `action_required` — the fork-PR approval gate, not a failure on my side. Compare with the previous head `a11910c`, where the same nine came back green once approved:

| workflow | `a11910c` (approved) | `05d0615` (current) |
|---|---|---|
| tests / style / builds / demo / references / scripts / pull_requests / Docker image on ghcr.io | success | `action_required` |
| docker | cancelled | `action_required` |

So CI here needs an "Approve and run" whenever I push. Since the only change between the two heads is the 10-line docstring deletion plus the rebase onto `c7f7218`, I would expect the same result, but I would rather point at the gate than claim a green run I cannot produce myself.

linhongyu510:
The docstring is gone and CI is green now that you approved the run — **56 of the 57 checks pass**, including `tests`, `mypy`, `ruff`, all 15 `build` matrix jobs, and `codecov/patch` at **100.00% of diff hit**.

The one red check is `codecov/project` (`97.21%, -0.02%`), and I dug into it rather than asking you to wave it through. **It is not caused by this PR** — it is a pre-existing nondeterminism in the coverage of a file this PR never touches.

**Codecov's own per-file data says so.** Of the 169 files, exactly one has a changed miss count, and it is not one of mine:

| file | in this PR's diff? | misses base → head |
|---|---|---|
| `models/detection/predictor/pytorch.py` | yes | 1 → 1 |
| `models/kie_predictor/pytorch.py` | yes | 0 → 0 |
| `models/predictor/pytorch.py` | yes | 42 → 42 |
| **`transforms/functional/base.py`** | **no** | **4 → 6** |

For that last file Codecov reports `has_diff: false`, `patch: null`, and identical line counts (68 → 68) — same code, 2 more misses. The arithmetic matches the project delta exactly: 278 → 280 misses, and 9761/10039 − 9765/10045 = 0.018%.

**Root cause.** `create_shadow_mask` in that file draws from an unseeded RNG and branches on it:

```python
_params = np.random.rand(1)
quad_idx = int(_params[0] / 0.25)   # 0,1,2,3 -> four different branches
if quad_idx % 2 == 0:              # line 173
    ...
    if quad_idx == 0: ...          # line 175
else:                              # line 178
    ...
    if quad_idx == 1: ...          # line 180
```

Each run reaches only some of those lines, so the miss count for the file moves on its own. Measured with `coverage.py`, changing **nothing but the seed**:

```
seed=0 : 31 missing lines   missing in 165-185: [173, 174, 175]
seed=1 : 31 missing lines   missing in 165-185: [178, 179, 180]
seed=7 : 36 missing lines   missing in 165-185: [175, 178, 179, 180]
seed=3 : 31 missing lines   missing in 165-185: [178, 179, 180]
```

Over 400 seeds the four quadrants come up 104 / 99 / 95 / 102 times, so which lines get counted is close to a coin flip per run.

**This also shows up on `main` without any PR involved.** From Codecov's commit history:

| commit on main | what it changed | misses | coverage |
|---|---|---|---|
| `60cb879` | Correct some vocabs | 276 | 97.25% |
| `1a6f4c1` | **Update docs — no code at all** | 278 | 97.23% |
| `c7f7218` | Correct vocabs (#2139) | 278 | 97.23% |

A docs-only commit moved it by the same 2 misses / 0.02%. My own earlier head `a11910c` happened to land on the lucky side and reported `+<0.01%`; after the rebase onto `c7f7218` it landed on the unlucky side. The diff between those two heads is the 10-line docstring deletion.

So this PR adds 7 executable lines, all 7 covered, and touches no other file. Happy to seed the RNG in the shadow-mask tests as a separate PR if you'd like that flake fixed — it is unrelated to this change, so I did not fold it in here.

One request: the review is still marked `changes_requested`, which is what `mergeable_state: blocked` is reporting now that the checks are done. The thread is already showing as outdated against `05d0615`.

linhongyu510:
Opened #2144 for the coverage flake described above — it is independent of this PR and touches only `tests/pytorch/test_transforms_pt.py` (`+37/-0`, no source change), so the two can be reviewed and merged in either order.

If #2144 goes in first, the `codecov/project` noise on this PR should settle on its own, since that file stops contributing a random delta.

linhongyu510:
@felixdittrich92 Two small maintainer actions would unblock this pair — no code change needed on either.

**Here (#2138):** the docstring you asked to remove is gone (`05d0615`), and that thread is now marked outdated. CI on this head is **56 of 57 green**, including `tests`, `mypy`, `ruff`, all 15 `build` matrix jobs, and `codecov/patch` at 100% of diff hit. The single red is `codecov/project` (`-0.02%`), which I traced to a pre-existing coverage flake in `transforms/functional/base.py` — a file this PR never touches — rather than asking you to wave it through. So the only thing holding `mergeable_state` at `BLOCKED` is the `changes_requested` from 09-15; a dismissal or re-review would clear it.

**And #2144**, the fix for that flake, currently shows only **1 check run** — the fork-PR approval gate has not been triggered on it at all, so there is no result to look at. An "Approve and run" would let it report. It is test-only, `+37/-0`, no source change.

Worth noting the two are order-independent: if #2144 lands first, the `codecov/project` noise here should settle on its own, since that file stops contributing a random delta.

felixdittrich92 (approved):
Thanks 👍

PR #2140, opened 2026-09-19
[datasets/detection] Stable class channels for multi-class targets and optional background supervision

fercho1999leon:
## What

Fixes two issues in multi-class text detection training and adds an option to supervise absent classes as background.

### Stable class → channel mapping
`pre_transform_multiclass` only included the classes present in each image, while `build_target` (DBNet and FAST) assigns a channel per *position* of the target dict. When a class had no box in an image, the channels of the following classes shifted and the model learnt crossed labels.

- `pre_transform_multiclass` takes an optional `class_names` argument; when given, every sample yields every class in that order, with an empty `(0, 4)` / `(0, 4, 2)` array for the absent ones.
- `DetectionDataset` and `LayoutDataset` pass their (sorted) class list, so the mapping is identical across samples.

### Images without any box
`format_polygons` raised on `np.concatenate([])` (and `LayoutDataset` on a `(0,)` array). Both datasets now accept such images as all-background samples.

### Absent classes and empty images (DBNet, FAST, LinkNet)
An image without any box, or a class absent from an image, is supervised as **background**: what the labels say is the truth (a blank page is a true negative). For partially annotated data the new `mask_empty_classes=True` option (`--mask-empty-classes` in `references/detection/train.py`) masks those channels out of the loss instead. `build_target` maps the target dict to channels by class name (any dict order) and rejects unknown or missing keys; the losses no longer crash on a batch made only of empty images. The training script also checks that train and val expose the same classes.

## Tests
- `test_detection_dataset_multiclass_missing_classes`: samples with a missing class / no box expose every class with the right empty shapes (boxes and polygons), through a `DataLoader`.
- `test_layout_dataset_empty_image`.
- `test_detection_multiclass_empty_class_target` (DB, FAST, LinkNet): target masks for both modes, finite loss and backward, including a batch made only of an empty image.
- `test_detection_multiclass_target_channels_by_name`: channel order independent of the dict order, unknown / missing keys rejected.

`ruff check`, `mypy doctr/` (same two pre-existing errors as `main`) and the detection/layout dataset and model tests pass locally.

## Context
Found while fine-tuning `db_resnet50` on a 6-class receipt dataset where a rare class was absent from ~80 % of the images; with the fix the model stopped predicting that class on every page.


felixdittrich92 (changes requested):
Thanks for the PR @fercho1999leon 👍 

1. **LinkNet training breaks:** `train.py` always passes `mask_empty_classes`, but `LinkNet.__init__` doesn't accept it, so `--arch linknet_*` fails with a `TypeError`. Please add the option to LinkNet, like DBNet/FAST (`linknet/base.py` has the same masking block).
2. **DBNet crashes on a batch with no boxes:** `l1_loss`/`focal_loss`/`dice_loss` are only set inside the `if torch.any(...)` branches, so a batch with no boxes raises `UnboundLocalError`. This could already happen before, but empty images are now loadable, so any empty image with `batch_size=1` or an all-empty batch hits it during training or validation. Please initialize the losses to zero, e.g. `out_map.sum() * 0`.

**Should fix**
3. With the default `mask_empty_classes=True`, empty images are fully masked and add nothing to the loss. That doesn't match "accepted as all-background samples". For single-class text detection a blank page is a true negative, so please either reconsider the default or adjust the wording/docs.
4. `build_target` still maps classes to channels by position (`enumerate(tgt.values())`). Mapping with `self.class_names.index(k)`, and raising on unknown or missing keys, would make it independent of dict order and of the order of the model's `class_names`.

**Nits**
- `pre_transform_multiclass`: all missing classes share the same `empty` array; please create one per class.
- Test: `bool(x.any()) == expected` is clearer than `x.any() is np.bool_(...)`.
- README: "every class must appear in every labels.json" can be read as "every image". Each class only has to appear somewhere in each split.
- The new model test downloads the backbone. Pass `pretrained_backbone=False` so it runs offline and faster.

fercho1999leon:
Thanks for the thorough review @felixdittrich92, all points addressed in 42b150c (rebased on `main`):

1. **LinkNet**: `mask_empty_classes` added to `LinkNet.__init__` and `_LinkNet.build_target` (same masking block as DBNet/FAST); `linknet_*` trains again with the script.
2. **Empty batches**: DBNet initialises `focal_loss` / `dice_loss` / `l1_loss` (and `focal_scale`, which was also inside the `if`) to `out_map.sum() * 0`; LinkNet returns the same zero when the whole batch is masked. Covered by a new assertion that runs forward + backward on a batch made only of an empty image, for the three architectures and both masking modes.
3. **Default**: reconsidered, you are right that a blank page is a true negative. The default is now `mask_empty_classes=False`: what the labels say is the truth, and an image without any box, or a class absent from an image, is supervised as background. Masking is opt-in for partially annotated data (`--mask-empty-classes` in `train.py`, replacing `--exhaustive-labels`). Docstrings and README reworded. Note that before this PR an image without boxes could not even be loaded, and a class absent from an image hit the channel-shift bug, so no existing behaviour is silently changed.
4. **`build_target` by name**: DBNet, FAST and LinkNet now iterate over `self.class_names` and read `tgt[class_name]`, raising a `ValueError` naming the unknown / missing keys. New test `test_detection_multiclass_target_channels_by_name` checks that a dict given in a different order lands on the right channels and that extra / missing keys are rejected.

Nits: one empty array per class in `pre_transform_multiclass`, `bool(x.any()) == expected` in the test, README now says every class must appear somewhere in each split (empty list inside an image), and the model test uses `pretrained_backbone=False` (runs offline, 2.7 s for the 9 cases).

fercho1999leon:
Full runs of `tests/pytorch/test_models_detection_pt.py`, `tests/pytorch/test_datasets_pt.py` and `tests/common/test_datasets_utils.py` on a Colab L4 with this branch: **176 passed** (all detection architectures with pretrained weights, straight boxes and polygons). One run hit `test_synthtext[recognition]` (9 crops instead of 10), which passes on rerun and on `main` too: its fixture draws the word boxes with an unseeded `np.random.randint`, so a degenerate crop is occasionally dropped. Unrelated to this PR; happy to send a one-line fix seeding that fixture if you want.

felixdittrich92 (changes requested):
Thanks for the updates @fercho1999leon 👍 

Looks mostly good to me.

One thing left:

DBNet + LinkNet: fine

FAST: "background" isn't fully supervised now because of the dice loss

suggestion:

```python
        pooled_logits = self.pooling(out_map)
        kernels = torch.sigmoid(out_map)
        prob_map = torch.sigmoid(pooled_logits)

        def masked_class_mean(per_class: torch.Tensor, valid: torch.Tensor) -> torch.Tensor:
            # Average over the classes that have something to supervise; a graph-preserving zero if none has
            return (per_class * valid).sum() / valid.sum().clamp(min=1)

        # As described in the paper, we use the Dice loss for the text segmentation map and the Dice loss scaled by 0.5.
        selected_masks = torch.cat(
            [ohem_sample(score, gt, mask) for score, gt, mask in zip(prob_map, seg_target, seg_mask)], 0
        ).float()
        inter = (selected_masks * prob_map * seg_target).sum((0, 2, 3))
        cardinality = (selected_masks * (prob_map + seg_target)).sum((0, 2, 3))
        text_dice = 1 - 2 * inter / (cardinality + eps)
        # With no positive pixel in the batch, Dice is a constant with zero gradient: false positives of that class
        # would never be penalised. Supervise such classes as background with a BCE on the selected pixels instead.
        num_selected = selected_masks.sum((0, 2, 3))
        has_pos = (selected_masks * seg_target).sum((0, 2, 3)) > 0
        neg_bce = (selected_masks * F.softplus(pooled_logits)).sum((0, 2, 3)) / num_selected.clamp(min=1)
        text_loss = masked_class_mean(torch.where(has_pos, text_dice, neg_bce), num_selected > 0) * 0.5

        # As described in the paper, we use the Dice loss for the text kernel map.
        selected_masks = seg_target * seg_mask
        inter = (selected_masks * kernels * shrunken_kernel).sum((0, 2, 3))  # noqa
        cardinality = (selected_masks * (kernels + shrunken_kernel)).sum((0, 2, 3))  # noqa
        # The kernel map is only defined inside text regions: skip the classes without any in the batch
        kernel_dice = 1 - 2 * inter / (cardinality + eps)
        kernel_loss = masked_class_mean(kernel_dice, selected_masks.sum((0, 2, 3)) > 0)

        return text_loss + kernel_loss
```

Instead of:


```python
        if len(self.class_names) > 1:
            kernels = torch.softmax(out_map, dim=1)
            prob_map = torch.softmax(self.pooling(out_map), dim=1)
        else:
            kernels = torch.sigmoid(out_map)
            prob_map = torch.sigmoid(self.pooling(out_map))

        # As described in the paper, we use the Dice loss for the text segmentation map and the Dice loss scaled by 0.5.
        selected_masks = torch.cat(
            [ohem_sample(score, gt, mask) for score, gt, mask in zip(prob_map, seg_target, seg_mask)], 0
        ).float()
        inter = (selected_masks * prob_map * seg_target).sum((0, 2, 3))
        cardinality = (selected_masks * (prob_map + seg_target)).sum((0, 2, 3))
        text_loss = (1 - 2 * inter / (cardinality + eps)).mean() * 0.5

        # As described in the paper, we use the Dice loss for the text kernel map.
        selected_masks = seg_target * seg_mask
        inter = (selected_masks * kernels * shrunken_kernel).sum((0, 2, 3))  # noqa
        cardinality = (selected_masks * (kernels + shrunken_kernel)).sum((0, 2, 3))  # noqa
        kernel_loss = (1 - 2 * inter / (cardinality + eps)).mean()

        return text_loss + kernel_loss
```

Please do a seeded check between main and the loss change  for single classes loss.

fercho1999leon:
Thanks @felixdittrich92, your FAST suggestion is applied in 60bfa65 (with `self.pooling(out_map)` computed once as `pooled_logits`), plus a regression test.

**Seeded single-class check** (`fast_tiny`, `CLASS_NAME`, batch of 2 with boxes, 3 SGD steps, same seed, same script run against `main` and this branch):

| step | `main` loss / grad norm | this PR |
|---|---|---|
| 1 | 0.892184 / 24.173953 | 0.892184 / 24.173953 |
| 2 | 0.810673 / 2.993625 | 0.810673 / 2.993625 |
| 3 | 0.799742 / 3.326867 | 0.799742 / 3.326867 |

Bit-identical, as expected: with one class and positives present, `has_pos` is true, `masked_class_mean` over a single valid class equals `.mean()`, and sigmoid was already the single-class path.

**Where it does change**, which is the point of the fix:

| case | `main` | this PR |
|---|---|---|
| single class, batch with **no** boxes: loss / grad norm | 1.5 / **0.0** | 8.325836 / 168.55 |
| same batch, 15 Adam steps | 1.5 at every step (Dice is a constant, nothing to learn) | 7.71 → 4.51 → 3.32 → 1.75 → ... → 0.014 |
| false positive of a class absent from the batch (logits −5 everywhere vs class "b" at +5) | 1.275 → 1.500 (only indirect, through the softmax competition) | 1.235 → 2.485 (directly penalised) |
| multi-class batch with one absent class | 1.2566 | 2.0425 (the absent class is now supervised) |

Test added: `test_fast_supervises_classes_without_positives` (false positives cost more than silence, and a background-only batch yields a non-zero gradient). `tests/pytorch/test_models_detection_pt.py -k "fast or multiclass"`: 26 passed. `ruff check` and `mypy doctr/` clean (same pre-existing `doctr.version` error as `main`).

fercho1999leon:
Also covered the line codecov flagged in `doctr/datasets/utils.py` (0f69946): `test_pre_transform_multiclass` in `tests/common/test_datasets_utils.py` checks the `class_names` ordering, the per-class empty arrays and the `ValueError` on an unknown class. `tests/common/test_datasets_utils.py`: 30 passed.

tests/pytorch/test_models_detection_pt.py
```diff
@@ -274,3 +274,22 @@ def test_detection_multiclass_target_channels_by_name(arch_name):
         model.build_target([{"a": box_a, "b": box_b, "c": box_a}], (2, 64, 64))
     with pytest.raises(ValueError, match="missing"):
         model.build_target([{"a": box_a}], (2, 64, 64))
+
+
+def test_fast_supervises_classes_without_positives():
```
felixdittrich92: This test can be removed

felixdittrich92 (changes requested):

tests/pytorch/test_models_detection_pt.py
fercho1999leon: Removed in ee39f3db2; the rest of the detection tests still pass (26 passed).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

felixdittrich92 (approved):
Thanks LGTM 👍

PR #2141, opened 2026-09-19
[layout] Batch the LW-DETR Hungarian matching and loss across all prediction sets

fercho1999leon:
## What

`LWDETR.forward` called `compute_loss` once per (group, prediction set): 13 groups × (final decoder layer + auxiliary decoder layers + encoder proposals) = 52 calls per training step. Each call copied every sample's cost matrix to the host for `linear_sum_assignment` and re-uploaded the targets, i.e. ~200 device synchronizations per step during which the GPU sits idle.

This PR stacks all prediction sets into `(S, B, Q, ...)` tensors, computes the matching costs of every (set, sample) pair on the device in one pass, copies them to the host **once**, solves the (tiny) assignment problems on the CPU and sends the matched indices back in a single transfer. The matched pairs are then scored with vectorized operations (IA-BCE, L1, ProbIoU) exactly as before.

- `compute_loss` keeps its public signature and now delegates to `_compute_sets_loss` (S = 1).
- `_pad_targets` builds the padded target tensors with one host-to-device transfer.
- Behaviour is identical: all sets were already weighted `1 / group_detr` and normalized by the same `num_boxes`, so the batched sum equals the previous sum of per-set losses up to float summation order.

## Numbers

| | before | after |
|---|---|---|
| loss + backward of the loss block, batch 4, 52 sets (Apple M-series, MPS) | 434 ms | 13 ms |
| full epoch, 589 images at 768 px, batch 8, bf16, Colab L4 | 3.1 min (3 img/s, inference ran at 17 img/s) | 1.1 min |
| loss / gradient relative difference on a fixed batch | – | 6e-8 / 7e-10 |

## Tests
- `test_lw_detr_batched_loss_matches_per_set`: batched loss and gradients equal the sum of the per-set `compute_loss` calls on a batch with a missing class and an empty image.
- Existing `lw_detr_s` model/zoo tests pass; `ruff check` and `mypy doctr/` unchanged.


doctr/models/layout/lw_detr/pytorch.py
```diff
@@ -650,90 +642,146 @@ def compute_loss(
...
+            for s in range(n_sets):
+                q_ind, t_ind = linear_sum_assignment(cost[s, b, :, :n])
+                set_idx.append(np.full_like(q_ind, s))
+                batch_idx.append(np.full_like(q_ind, b))
+                query_idx.append(q_ind)
+                tgt_idx.append(t_ind)
+        if set_idx:
+            index = torch.from_numpy(np.stack([np.concatenate(i) for i in (set_idx, batch_idx, query_idx, tgt_idx)]))
+            index = index.to(device, dtype=torch.long)  # one transfer for every matched pair
+            s_i, b_i, q_i, t_i = index
```
felixdittrich92: Please use more meaningful names set_idx, batch_idx, and so on

felixdittrich92 (changes requested):
Thanks for the PR @fercho1999leon 👍 

This change will increase the peak memory - would be good to have a comparision between main and your modifications same for a small seeded iteration that the loss does not shift.

fercho1999leon:
Thanks for the review @felixdittrich92! Both points addressed:

**Naming**: the matched index arrays are now `matched_set_ids` / `matched_sample_ids` / `matched_query_ids` / `matched_target_ids` (host side) and `set_ids` / `sample_ids` / `query_ids_t` / `target_ids_t` (device side), with a comment explaining the gather. Rebased on `main`.

**Memory and seeded comparison** (`lw_detr_s`, 6 classes, batch 4, 512 px, one empty image in the batch, 5 AdamW steps with the same seed, same script run against `main` and this branch):

| | `main` | this PR |
|---|---|---|
| losses, steps 1-5 | 14.579844, 14.527025, 14.457556, 14.398611, 14.409478 | 14.579844, 14.527027, 14.457555, 14.398612, 14.409481 |
| peak allocated device memory (MPS, `torch.mps.current_allocated_memory()` sampled after forward and after backward) | 2967 MB | 2966 MB |
| 5 steps, CPU (Apple M3) | 11.7 s | 10.8 s |

The losses differ only in the 6th decimal (float summation order; gradients match to ~1e-9 in the added unit test). On the memory side the extra tensors are the stacked `(S, B, Q, C)` logits / `(S, B, Q, 6)` boxes and the `(S, B, Q, Nmax)` cost: with S = 52, B = 4, Q = 195, C = 6, Nmax = 6 that is under 5 MB in float32, which is why the peak is unchanged relative to the backbone/decoder activations. The per-set intermediates the old code kept alive for backward (one `prob` / `pos_weights` / `neg_weights` per set) are the same ones the batched version keeps, just contiguous.

Peak RSS on CPU varies by ±400 MB between identical runs on this machine, so I did not include it; happy to add a CUDA `max_memory_allocated` number if you prefer that one.

fercho1999leon:
CUDA numbers as promised (Colab **NVIDIA L4**, `lw_detr_s`, 6 classes, batch 4, 768 px, one empty image in the batch, 10 AdamW steps, same seed, same script against `main` and this branch):

| | `main` | this PR |
|---|---|---|
| `torch.cuda.max_memory_allocated()` | 8407 MB | **8407 MB** |
| `torch.cuda.max_memory_reserved()` | 10828 MB | 10830 MB |
| 10 training steps | 13.05 s (12.93 s, 12.42 s on reruns) | **5.58 s** (5.69 s) |
| loss, step 1 | 14.551973 | 14.551972 |

So the peak memory is unchanged and the step is 2.3x faster end to end on this GPU.

On the seeded trajectory: step 1 matches to 1e-6, then the runs drift apart in the 2nd-3rd decimal (step 2: 14.630 vs 14.643). That drift is GPU non-determinism, not the loss change: three runs of **`main`** itself give 14.630 / 14.620 / 14.622 at step 2 and 14.518 / 14.541 / 14.533 at step 3, i.e. the same spread. On CPU, where the kernels are deterministic, the 5-step trajectories of `main` and this branch match to 1e-6 (previous comment), and the unit test checks loss and gradients at 1e-5 / 1e-4 tolerance.

felixdittrich92 (approved):
LGTM Thanks 👍

PR #2142, opened 2026-09-19
[references] Device selection, bfloat16 AMP, checkpoint sidecar and small fixes for the detection and layout training scripts

fercho1999leon:
## What

Quality-of-life changes to `references/detection/train.py` and `references/layout/train.py`, found while training on Apple Silicon, Windows and Colab:

- **`--device`** accepts a CUDA index, `cuda:N`, `mps` or `cpu` (default: CUDA, then MPS, then CPU). Tensors follow the model device instead of hard-coded `.cuda()`; `pin_memory` and cudnn benchmark only on CUDA; `--amp` is refused off CUDA with a clear message. (`resolve_device` lives in each script's `utils.py`.)
- **`--amp-dtype {float16,bfloat16}`**: bfloat16 has the float32 range, so no `GradScaler` is needed and the LW-DETR Hungarian matching no longer overflows (with float16 the cost matrix contained `inf` and `linear_sum_assignment` raised `matrix contains invalid numeric entries`).
- **`--no-hflip`** to disable the horizontal flip augmentation (mirrored text is misleading when the classes are semantic regions whose position matters).
- **Checkpoint sidecar**: every `<name>.pt` is written with a `<name>.json` holding the architecture, ordered class names, input size, dataset hashes, git revision, versions and the full argument list, so a checkpoint can be reloaded for inference (e.g. with the right `class_names`) without remembering how it was trained. `output_dir` is created if missing (the scripts used to crash at the first save).
- The identity augmentation is a module-level function: `lambda x: x` could not be pickled by DataLoader workers under the `spawn` start method (macOS / Windows), so `-j > 0` failed there.
- Layout: **`--labels-name`** to pick another label file inside the split folders.

Both scripts were smoke-trained for one epoch on MPS (`db_mobilenet_v3_large`, `lw_detr_s`); the CUDA/DDP paths are unchanged apart from the device plumbing. READMEs updated.


felixdittrich92:
Hi @fercho1999leon :wave:,

Thanks for the PR :+1: 

We should split this into several PRs:

1. a PR with the device / amp assignment which applies to every train / val script in the `references` subfolders (not only detection and layout)

2. a PR (also in every training script) which saves useful metadata besides the checkpoints - also a short documentation part should be added to each README in the references subfolders

3. an PR to fix the following bug (The identity augmentation is a module-level function: lambda x: x could not be pickled by DataLoader workers under the spawn start method (macOS / Windows), so -j > 0 failed there.)

4. first an issue about the training configuration (something like `--no-hflip` is not really useful this way and requires additional documentation) - something we could think about is per task a .yaml configuration file which contains the whole config to avoid adding more and more CLI args - but I would like to discuss this first via issue :smiley:

fercho1999leon:
Thanks @felixdittrich92, agreed on all four points. Splitting it up:

1. device / AMP for every `references` script: next PR (this one will be closed once it is up).
2. checkpoint metadata for every script + README sections: separate PR after 1.
3. the picklable identity augmentation: #2148 (also fixes `references/table/train.py`, which has the same lambda).
4. training configuration: opened #2149 with a proposal for a per-task YAML config and `--set key=value` overrides; `--no-hflip` and the augmentation switches would live there instead of as flags.

I'll close this PR when 1 and 2 are open so the discussion stays in one place per topic.

fercho1999leon:
Closing in favour of the split: #2150 (device / AMP for every training and evaluation script), #2148 (picklable identity augmentation, merged path), #2149 (training configuration discussion). The checkpoint metadata part will come as its own PR.

PR #2143, opened 2026-09-19
[transforms] RandomPerspective, RandomGlare and RandomLighting for photographed pages

fercho1999leon:
## What

Three augmentations for training on pages that end up **photographed** (phone photos of screens or sheets) rather than scanned or screenshotted. The existing pipeline covers rotation, crops, shadows, noise and blur, but nothing simulates a page seen at an angle or the lighting artefacts of a photo.

- **`RandomPerspective(distortion)`** (geometric): moves the four image corners inwards by up to `distortion × size` and warps the image so that the original corners land on the moved ones, as if photographed at an angle. The same homography is applied to the boxes / polygons (format preserved, clipped to `[0, 1]`) and to the padding mask, so the layout models' validity masks stay correct. Multi-class dict targets share one warp; empty classes are kept.
- **`RandomGlare(opacity_range, num_spots)`** (photometric): soft specular reflections, like a light source mirrored on a screen or a glossy sheet.
- **`RandomLighting(strength_range)`** (photometric): multiplies the image by a smooth low-frequency brightness field (uneven lighting, crumpled or curved sheet shading, vignetting).

Both photometric transforms keep the input dtype (float in `[0, 1]` or uint8), like `RandomShadow`. The functional counterparts (`random_glare`, `random_lighting`, `perspective_sample`) live in `doctr.transforms.functional`.

Visually (a receipt with the three transforms plus `RandomShadow`): the page is warped like a photo taken at an angle, with a diagonal shadow, a reflection and uneven brightness, and the drawn boxes stay exactly on the fields; the test suite checks that geometry programmatically.

## Tests
- `test_random_perspective` (boxes and polygons): zero distortion is the identity, distorted geometries stay in `[0, 1]` with the same format, padding is masked out, every class of a dict target gets the same warp.
- `test_random_glare_and_lighting`: shapes, value range and dtype preservation for float and uint8.

`ruff check`, `mypy doctr/` (same two pre-existing errors as `main`) and `tests/pytorch/test_transforms_pt.py` pass locally. Docs entries added to `modules/transforms.rst`.

## Context
Fine-tuning the LW-DETR layout model on receipts with `RandomPerspective(0.2)` + these photometric transforms (40 epochs from a converged checkpoint) improved the clean validation metrics as well (mAP@[.5:.95] 68.8 → 70.8, AP@.5 88.1 → 89.0, val loss 2.98 → 2.90) and made the model usable on phone photos. Wiring them into the reference training scripts as `--perspective` / `--photo-aug` flags can follow once #2142 lands, to avoid conflicting changes in `references/`.


felixdittrich92 (approved):
Thanks @fercho1999leon LGTM 👍

PR #2147, opened 2026-09-21
docs: drop non-existent angle_tol/ratio_tol from bitmap_to_boxes docstrings

linhongyu510:
## What

`bitmap_to_boxes` in the three detection post-processors documents two parameters that do not exist:

```python
def bitmap_to_boxes(self, pred: np.ndarray, bitmap: np.ndarray) -> np.ndarray:
    """...
    Args:
        pred: Pred map from ...
        bitmap: Bitmap map computed from pred (binarized)
        angle_tol: Comparison tolerance of the angle with the median angle across the page   # not a parameter
        ratio_tol: Under this limit aspect ratio, we cannot resolve the direction of the crop # not a parameter
    """
```

The signature is `(pred, bitmap)` in all three, and `angle_tol` / `ratio_tol` appear **nowhere else** in `doctr/` — the only six occurrences in the package are these docstring lines. They look like copy-paste leftovers that propagated across:

- `doctr/models/detection/differentiable_binarization/base.py` — `DBPostProcessor`
- `doctr/models/detection/linknet/base.py` — `LinkNetPostProcessor`
- `doctr/models/detection/fast/base.py` — `FASTPostProcessor`

## Change

Remove the two stale lines from each of the three docstrings so the `Args` block matches the actual signature. `+0/-6`, docs only, no code touched.

## How I found and verified it

Found by parsing every docstring's `Args:` block in `doctr/` with `ast` and diffing the documented names against the real signature — these three were the only functions documenting a non-existent parameter.

Verified after the edit:

```
grep -rn 'angle_tol\|ratio_tol' doctr/            -> no matches
ruff check / ruff format --check (3 files)         clean
python -c import + inspect on all three classes:
    DBPostProcessor.bitmap_to_boxes    params=[self, pred, bitmap]  stale tokens in doc: False
    LinkNetPostProcessor.bitmap_to_boxes params=[self, pred, bitmap]  stale tokens in doc: False
    FASTPostProcessor.bitmap_to_boxes  params=[self, pred, bitmap]  stale tokens in doc: False
```

This is a small follow-up in the same spirit as #2138 / #2144. I left one adjacent nit untouched to keep the diff focused: the `bitmap:` line in `fast/base.py` and `linknet/base.py` both say "differentiable linknet output" (fast's should say fast) — happy to fold that in here or in a separate PR if you'd like.


felixdittrich92 (approved):
Thanks 👍

PR #2148, opened 2026-09-21
[references] Picklable identity augmentation so DataLoader workers start under spawn

fercho1999leon:
## What

Split out of #2142 as requested (bug fix only).

The detection, layout and table training scripts use `lambda x: x` as the "no augmentation" branch of the `OneOf` image transforms. DataLoader workers pickle the dataset and its transforms under the `spawn` start method, the default on macOS and Windows, and a lambda defined inside `main()` cannot be pickled, so any `-j > 0` fails there:

```
AttributeError: Can't pickle local object 'main.<locals>.<lambda>'
```

The lambda is replaced by a module-level `identity` function in `references/detection/train.py`, `references/layout/train.py` and `references/table/train.py` (the three scripts that had it). No behaviour change on Linux (`fork`).

## Verification

On macOS (Python 3.11, torch 2.14), `references/detection/train.py db_mobilenet_v3_large ... -j 2`:
- `main`: crashes at the first batch with the error above
- this branch: trains normally; same for `references/layout/train.py lw_detr_s ... -j 2`

`ruff check` passes.


references/table/train.py
```diff
@@ -43,6 +43,15 @@
 from utils import EarlyStopper, build_param_groups, plot_recorder, plot_samples
 
 
+def identity(x):
+    """No-op augmentation.
```
felixdittrich92: Let's shorten it (also in the other scripts):

"""No-op augmentation.""" is enough as docstring

felixdittrich92 (changes requested):
Small comment afterwards can be merged 👍

fercho1999leon:
Done, shortened to `"""No-op augmentation."""` in the three scripts. Thanks!

felixdittrich92 (approved):
Thanks 👍

Issue #2149, opened 2026-09-21
Per-task YAML configuration for the reference training scripts


fercho1999leon:
### 🚀 The feature

Let each `references/*/train.py` read its full training configuration from a YAML file (one per task, with a shipped default), keeping the CLI only for overriding individual keys:

```shell
python references/detection/train.py --config references/detection/configs/default.yaml \
    --set data.train_path=/data/train data.val_path=/data/val training.epochs=20
```

A first sketch of the schema, mirroring what the scripts already accept (the six training scripts currently declare 27 to 36 `add_argument` calls each, most of them shared):

```yaml
model:
  arch: db_resnet50
  pretrained: true
  input_size: 1024
  assume_straight_pages: true          # `--rotation` today
data:
  train_path: null                     # or built-in datasets: train_datasets: [FUNSD, SVHN]
  val_path: null
  workers: null
  labels_name: labels.json
augmentations:                         # per task defaults, all switchable
  horizontal_flip: 0.15
  random_crop: {p: 0.25, scale: [0.75, 1.0], ratio: [0.85, 1.15]}
  random_resize: {p: 0.25, scale_range: [0.4, 0.9]}
  rotation: {enabled: false, max_angle: 90, expand: true}
  perspective: {p: 0.0, distortion: 0.2}        # RandomPerspective (#2143)
  photometric: {glare: 0.0, lighting: 0.0, shadow: 0.3, blur: 0.3, noise: 0.1, grayscale: 0.15, color_inversion: 0.3}
training:
  epochs: 10
  batch_size: 2
  optimizer: {name: adam, lr: 1e-3, weight_decay: 0, backbone_lr: null}
  scheduler: poly
  amp: {enabled: false, dtype: float16}
  freeze_backbone: false
  early_stopping: {enabled: false, patience: 5, min_delta: 0.01}
  device: null                         # cuda index, cuda:N, mps, cpu; null = auto
  backend: nccl
runtime:
  output_dir: .
  name: null
  save_interval_epoch: false
  resume: null
  test_only: false
  logging: {wandb: false, clearml: false, push_to_hub: false}
```

Implementation idea, to keep the diff reviewable:
1. `references/common/config.py`: load YAML → nested dict → `SimpleNamespace`/dataclass, apply `--set key.path=value` overrides (typed from the default), and validate unknown keys.
[...]

PR #2150, opened 2026-09-21
[references] Device selection and AMP dtype for every training and evaluation script

fercho1999leon:
## What

Point 1 of the split of #2142: the same device handling for **all** `references` scripts, i.e. the six training scripts (`classification/train_character.py`, `classification/train_orientation.py`, `detection/train.py`, `layout/train.py`, `recognition/train.py`, `table/train.py`) and the four evaluation scripts (`detection`, `layout`, `recognition`, `table`).

- **`--device`** accepts a CUDA index (`0`), `cuda:N`, `mps` or `cpu`; default: CUDA if available, then MPS, then CPU. A `resolve_device` helper lives in each folder's `utils.py` (same per-folder convention as `ddp_utils.py`). Tensors are moved to the model's device instead of hard-coded `.cuda()` / `torch.cuda.is_available()` checks; `pin_memory` and `cudnn.benchmark` only on CUDA; `--amp` is refused off CUDA with a clear message. Distributed mode (`torchrun`) is unchanged: each rank keeps its own GPU.
- **`--amp-dtype {float16,bfloat16}`** (training and evaluation): bfloat16 has the float32 range, so `GradScaler` is disabled for it and losses that overflow in float16 (the LW-DETR matching cost produced `inf` → `linear_sum_assignment` failure) train normally on Ampere or newer GPUs.
- A **"Device and mixed precision"** section in each of the five READMEs.

No behaviour change on a CUDA machine with the default arguments.

## Verification (Apple Silicon M3, MPS and CPU)

| script | run |
|---|---|
| `classification/train_character.py mobilenet_v3_small` | 1 epoch, `--device mps` |
| `classification/train_orientation.py --type page` | 1 epoch, `--device mps` |
| `recognition/train.py crnn_mobilenet_v3_small` (synthetic words) | 1 epoch, `--device cpu` |
| `detection/train.py db_mobilenet_v3_large` | 1 epoch `--device mps`, then `--resume --test-only --device cpu` (same val loss) |
| `layout/train.py lw_detr_s` | 1 epoch, `--device cpu` |
| `detection/evaluate.py db_mobilenet_v3_large --dataset FUNSD` | `--device cpu` |
| `recognition/evaluate.py crnn_mobilenet_v3_small --dataset FUNSD` | `--device cpu` (87.3 % exact) |
| `detection/train.py ... --device cpu --amp` | rejected with `ValueError` |
| `table/train.py`, `table/evaluate.py`, `layout/evaluate.py` | argument parsing only (no dataset at hand) |

`ruff check` passes. Follow-ups from #2142 still to come as separate PRs: checkpoint metadata for every script (point 2). Point 3 is #2148 and point 4 is issue #2149.


references/detection/evaluate.py
```diff
@@ -23,6 +23,18 @@
...
+from utils import amp_dtype, model_device, resolve_device
+
+AMP_DTYPE = torch.float16  # set from --amp-dtype in main()
+
+
+def _autocast():
+    return torch.amp.autocast("cuda", dtype=AMP_DTYPE)
+
+
+def _scaler():
```
felixdittrich92: `_scaler()` is unused in every evaluate script please remove

references/classification/utils.py
```diff
@@ -89,3 +90,45 @@ def early_stop(self, validation_loss: float) -> bool:
...
+            return torch.device("mps")
+        return torch.device("cpu")
+    if isinstance(device, int) or (isinstance(device, str) and device.isdigit()):
+        index = int(device)
+        if not torch.cuda.is_available():
+            raise AssertionError("PyTorch cannot access your GPU. Please investigate!")
+        if index >= torch.cuda.device_count():
+            raise ValueError("Invalid device index")
+        return torch.device("cuda", index)
+    dev = torch.device(device)
```
felixdittrich92, line 119:
```python
if dev.type == "cuda":
    if not torch.cuda.is_available():
        raise AssertionError("PyTorch cannot access your GPU. Please investigate!")
    if dev.index is None:
        dev = torch.device("cuda", torch.cuda.current_device())
    elif dev.index >= torch.cuda.device_count():
        raise ValueError("Invalid device index")
```

Otherwise by passing only `--device "cuda"` it will crash at torch.cuda.set_device() afterwards.

Needs to be fixed in each utils.py

felixdittrich92 (changes requested):
two small things left afterwards good to merge 👍

fercho1999leon:
Thanks @felixdittrich92, both fixed in 50dffe3:

1. `_scaler()` removed from the four evaluate scripts (they never train, so it was dead code; it slipped through because my check for the call also matched the definition).
2. `resolve_device` now pins an index-less `"cuda"` to `torch.cuda.current_device()` and validates the index otherwise, as you suggested, in the five `utils.py`.

Checked every spec on this Mac (`None`, `""`, `cpu`, `mps`) and, with `torch.cuda` monkeypatched to simulate a 2-GPU box, the CUDA branches:

| `--device` | resolved | |
|---|---|---|
| `cuda` (current device = 1) | `cuda:1` | safe for `torch.cuda.set_device` |
| `cuda:0` / `0` | `cuda:0` | |
| `cuda:1` / `1` | `cuda:1` | |
| `cuda:5` / `5` | `ValueError: Invalid device index` | |

Re-ran `detection/train.py --device mps` (1 epoch) and `detection/evaluate.py --dataset FUNSD --device cpu` after the change; `ruff check` clean.

felixdittrich92 (approved):
Thanks 👍

PR #2152, opened 2026-09-22
fix(io): stop Block sharing one mutable default list across instances

linhongyu510:
## What

`Block.__init__` used mutable default arguments:

```python
def __init__(
    self,
    lines: list[Line] = [],
    artefacts: list[Artefact] = [],
    ...
```

A mutable default is created once and shared across all calls, and `Element.__init__` stores it directly (`setattr(self, k, v)`), so every `Block` constructed without passing `lines`/`artefacts` aliased the same two list objects. Mutating one `Block` leaked into every other default-constructed one:

```python
b1 = Block(geometry=g)
b2 = Block(geometry=g)
b1.lines is b2.lines          # True
b1.lines.append(x)
b2.lines                      # [x]   <-- polluted
```

## Fix

Default to `None` and create fresh lists inside `__init__` — the standard remedy — leaving all other behaviour identical (the empty-list case is unchanged; nothing else in the method depends on the default identity):

```python
lines: list[Line] | None = None,
artefacts: list[Artefact] | None = None,
...
    if lines is None:
        lines = []
    if artefacts is None:
        artefacts = []
```

Scope kept to `Block` — it's the only `__init__` in `doctr/io/elements.py` with this problem. `Page`/`KIEPage` already use `| None = None` for their optional lists, and the other constructors take their collections as required positional arguments.

## Verification

Added `test_block_default_lists_are_not_shared`:

```
tests/common/test_io_elements.py   19 passed
ruff check                          clean
```

Load-bearing — reverting the `__init__` change fails the new test with `assert [] is not []` (the two Blocks share one list). The existing `test_block` and the rest of the file are unaffected.


doctr/io/elements.py
```diff
@@ -386,11 +386,16 @@ class Block(Element):
...
     def __init__(
         self,
-        lines: list[Line] = [],
-        artefacts: list[Artefact] = [],
+        lines: list[Line] | None = None,
+        artefacts: list[Artefact] | None = None,
         geometry: BoundingBox | np.ndarray | None = None,
         objectness_score: float | None = None,
     ) -> None:
+        # Avoid shared mutable default arguments: each Block must own its lists.
```
felixdittrich92: comment can be removed

tests/common/test_io_elements.py
```diff
@@ -432,6 +432,23 @@ def test_block():
     }
 
 
+def test_block_default_lists_are_not_shared():
```
felixdittrich92: test can be removed

felixdittrich92 (changes requested):

linhongyu510:
Thanks for the review! Dropped both the explanatory comment and the regression test — the diff is now just the `None` default + list init in `Block.__init__`.

felixdittrich92 (approved):
Thanks 👍🏼

